PR #1, opened 2021-12-22
feat: Create pre-commit checks

cdnninja:


cdnninja:
@fuatakgun  could you merge this once you create a token for this repo? Will need to be named PAT as of the correct commit.

fuatakgun (approved):

fuatakgun:
we should create an organization for both of us, rather than me keeping these under my account.

PR #5, opened 2021-12-23
feat: Added temp range support

cdnninja:


cdnninja:
Will merge to confirm this works for Canada.  If so will move the same logic into other regions.

cdnninja:
@fuatakgun this is ready for your review with temp range support.  I have left USA as is.  It reports Low and High as something our old override but I suspect we shouldn't force that to a number since in many cars low and high are valid values.

hyundai_kia_connect_api/KiaUvoApiCA.py
```diff
@@ -118,6 +119,15 @@ def get_cached_vehicle_status(self, token: Token):
 
         vehicle_status = {}
         vehicle_status["vehicleStatus"] = response
+        
+        #Coverts temp to usable number.  Currently only support celsius. Future to do is check unit in case the care itself is set to F. 
+        
+        value = vehicle_status["vehicleStatus"]["airTemp"]["value"]
```
fuatakgun, line 125: this looks like a perfect function into util.py eg;
```
get_hex_temp_into_index(value):
    value = value.replace("H", "")
    value = value.replace("C", "")
    value = "0x" + value
    value = int(value,16)
```

fuatakgun (changes requested):
can we move shared stuff into a function?

rest looks good to me, thanks

hyundai_kia_connect_api/KiaUvoApiCA.py
cdnninja: Done.

cdnninja:
As a note reason I was naming c vs F is that these cars can return but units.  Depends what the car itself is set to.  We can update that later though if required.  Right now I don't know the range a Canadian car supports in F.  Not worth my time to play with.

cdnninja:
Need your thoughts on how to fix this.  I removed imports on util since this causes circular imports.  This would cause the one function to fail though.   How do we best fix this?

fuatakgun (approved):

PR #10, opened 2022-01-03
feat: Added additional fields

cdnninja:
Let me know thoughts on these new fields - in particular the unit approach.  Thinking distance will all be the same unit same with temp.   Not sure the correct type though so left it as string.

### DIFF
--- a/hyundai_kia_connect_api/Vehicle.py
+++ b/hyundai_kia_connect_api/Vehicle.py
@@ -30,6 +30,7 @@ class Vehicle:
     _engine_is_running: bool = None
     _distance_unit: str = None
     _temperature_unit: str = None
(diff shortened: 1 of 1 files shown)


fuatakgun:
What will we store in distance unit field? Unit type coming from api response or unit type visible to user?

For some UK users, somw distance units were coming in KM some were Miles from api response, so it might be different per sensor.

i was thinking about using setter functions and having both value and unit as input of setter function and store both of them. As i understand, HA core supports unit conversion on UI side.

cdnninja:
I am good with that.  Yes HA can convert if we provide the unit correctly.

fuatakgun:
Still trying to validate if my assumption is correct :-)

PR #17, opened 2022-01-15
feat: Added Actions

cdnninja:


cdnninja:
Thoughts on starting to return a string on each of these items as the action status tracking for the APIs that support it?  Right now the return nothing.

fuatakgun (commented):
Overall;
- rename VehicleID to vehicle_id
- instead of reading vehicles[vehicle_id], let's use get_vehicle(id). I feel we might need to do something in that function in the future.
- cancel_climate sounds a bit different


Thanks for great work.

fuatakgun (commented):
Function names should be imperative;
I would go with start_climate or start_climate or just start or stop if you don't want to use climate

fuatakgun:
> Thoughts on starting to return a string on each of these items as the action status tracking for the APIs that support it?  Right now the return nothing. 

Let's have s sync on this, you are right

cdnninja:
Tweaked with the above feedback and also did some prep for the action tracking action.  I haven't updated the API files yet to provide the responses though.  Thought this would be a good to upload before I do that.

hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -88,6 +88,30 @@ def check_and_refresh_token(self) -> bool:
...
         return False
 
+    def start_climate(self, vehicle_id: str, options: ClimateRequestOptions) -> str:
+        return self.api.start_climate(self.token, self.get_vehicle(vehicle_id), options)
+            
+    def stop_climate(self, vehicle_id: str) -> str:
+        return self.api.stop_climate(self.token, self.get_vehicle(vehicle_id))
+
+    def lock(self, vehicle_id: str) -> str:
+        return self.api.lock_action(self.token, self.get_vehicle(vehicle_id), "close")
```
fuatakgun, line 98: Magical strings are evil :-) let's use ENUM from const.py

hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -88,6 +88,30 @@ def check_and_refresh_token(self) -> bool:
...
+    
+    def unlock(self, vehicle_id: str) -> str:
+        return self.api.lock_action(self.token, self.get_vehicle(vehicle_id), "open")
```
fuatakgun, line 101: Same magical string

hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -88,6 +88,30 @@ def check_and_refresh_token(self) -> bool:
...
+
+    def start_charge(self, vehicle_id: str) -> str:
+        return self.api.start_charge(self.token, self.get_vehicle(vehicle_id))
+
+    def stop_charge(self, vehicle_id: str) -> str:
+        return self.api.stop_charge(self.token, self.get_vehicle(vehicle_id))
+
+    def set_charge_limits(self, vehicle_id: str, ac_limit: int, dc_limit: int) -> str:
+        return self.api.start_climate(self.token, self.get_vehicle(vehicle_id), ac_limit, dc_limit)
```
fuatakgun, line 110: Correction: api.set_charge_limits

hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -88,6 +88,30 @@ def check_and_refresh_token(self) -> bool:
...
+
+    def check_action_status(self, vehicle_id: str, action_id: str):
+        return self.api.check_last_action_status(self.token, self.get_vehicle(vehicle_id), action_id)
```
fuatakgun, line 113: Correction: api.check_action_status

fuatakgun (commented):
I would say, for trackable actions, we should return a callable function to consumer as our internal functions working based on direct calls, no async loop here.

Caller can take callable and check action status with it, if interested.

hyundai_kia_connect_api/VehicleManager.py
cdnninja: Are you able to show me what that would look like?  Sounds like another concept for me to google.  :)

fuatakgun: As a starting point, we can return a id as you are doing now and create another function to check its status, not an urgent but it will make our life a lot of easier for HA, and anyone can use it inside async scope.

fuatakgun: Sorry, my comment was about rather than passing `close` or `open`, can we pass LockAction.Open and LockAction.Close which are in fact defined in const.py

fuatakgun (changes requested):

hyundai_kia_connect_api/VehicleManager.py
cdnninja: Should be resolved now.

cdnninja:
I need to do some live testing of this change on my car still.  Opening for review as I would like another pass at it and for you to test on yours.

PR #20, opened 2022-01-16
feat: Added detailed seat status

cdnninja:
Added detailed seat status.  This is much easier to do here vs in the custom component.  I haven't run a test of this yet, I also don't have seats that report state to test with.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -18,6 +18,7 @@ from .const import (
     DOMAIN,
     DISTANCE_UNITS,
     TEMPERATURE_UNITS,
+    SEAT_STATUS,
 )
 from .Token import Token
 from .utils import get_child_value, get_hex_temp_into_index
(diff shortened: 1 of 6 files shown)


fuatakgun:
is it working with home assistant without any errors?

cdnninja:
No, I edited the first comment before you posted but noticed in my mobile phone it doesn't show the edit.  Odd. 

I am going to do a round of testing with this plus #21 at the same time.  I also can't test the values as my car doesn't show seat status.

cdnninja:
I have now tested this with home assistant core.  Reader for merge if you are good with it.  As mentioned I can't see the sensors or data though.  It doesn't error for those without.

cdnninja:
As a note #21 will fix the failing python builds.

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -216,18 +217,18 @@ def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> No
         vehicle.side_mirror_heater_is_on = get_child_value(
             state, "vehicleStatus.sideMirrorHeat"
         )
-        vehicle.front_left_seat_heater_is_on = get_child_value(
+        vehicle.front_left_seat_status = SEAT_STATUS[get_child_value(
```
fuatakgun, line 220: overall, should we stick with `front_left_seat_heater_status` and `SEAT_HEAT_STATUS`? or do we store any state other than heat here?

fuatakgun (changes requested):

hyundai_kia_connect_api/KiaUvoApiEU.py
cdnninja: We store states including cool.   The constant file has the text states for your reference.

cdnninja: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/783b6abace0fc9154f9ea4c93f4aabb43dbaf440/hyundai_kia_connect_api/const.py#L22

cdnninja:
Ready for review and merge.

PR #21, opened 2022-01-17
fix:  Fix flake8 and pytest errors

cdnninja:
More exist.  I will leave this PR open for us to fix them all.


cdnninja:
@fuatakgun could you look at this one? 

./hyundai_kia_connect_api/VehicleManager.py:63:25: F821 undefined name 'datetime'
        started_at_utc: datetime = dt.datetime.now(pytz.utc)

We also have 

./hyundai_kia_connect_api/VehicleManager.py:63:25: F821 undefined name 'datetime'
        started_at_utc: datetime = dt.datetime.now(pytz.utc)

cdnninja:
Resolved all flake8 issues and put a blank test in.  If we want to use pytest it is now running.  Next up I need to figure out how to pass a username password into it without leaving it plain text.

cdnninja:
I am going to pause on this for tonight.  If you know the issue please update this branch.   After the import statement works I think the test will run using my account.

cdnninja:
Sorry for the endless commits.  It will squash to one in main.  This is ready for your review.  This is a single test at this point confirming login and that atleast 1 car is returned.  I have tested with home assistant core.  I am seeing a location issue still but I don't think it is related to this commit and not required for core.

tests/empty_test.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+from hyundai_kia_connect_api.VehicleManager import VehicleManager
+
+
+def test_login():
+    username = os.environ["KIA_CA_CDNNINJA_USERNAME"]
```
fuatakgun, line 7: curious: how are we going to store these securely?

fuatakgun (approved):

tests/empty_test.py
cdnninja: This is a GitHub token stored in the repon.  The GitHub actions sets it as a environment variable.  You can't view the password in GitHub and it is encrypted.   Only access is via actions so risk would be someone logs it or commits a pr that exposes it.

PR #30, opened 2022-01-21
feat(CA): Populate Engine Type

cdnninja:
Guessing on the letters used for Canada beyond ICE. Can't find a record in bluelinky or elsewhere what letter is correct. This is the framework of how other brands will use this field.


cdnninja:
Ready for review and merge.

fuatakgun:
👏👏👏👏

PR #39, opened 2022-02-08
feat: Update Vehicles

cdnninja:
Update vehicles instead of one time list.

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -24,6 +24,8 @@ class ClimateRequestOptions:
 class ApiImpl:
     data_timezone = dt.timezone.utc
     temperature_range = None
+    refresh_vehicles_on_token_refresh = False
+
 
     def __init__(self) -> None:
         """Initialize."""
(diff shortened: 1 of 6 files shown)


cdnninja:
This doesn't solve my second Kia USA vehicle.key issue but cleans up the code some.

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -34,8 +36,8 @@ def login(self, username: str, password: str) -> Token:
         """Login into cloud endpoints and return Token"""
         pass
 
-    def get_vehicles(self, token: Token) -> list[Vehicle]:
-        """Return all Vehicle instances for a given Token"""
+    def get_vehicles(self, token: Token, vehicles: list[Vehicle]) -> None:
```
fuatakgun, line 39: I still think this is a bad idea, `get_vehicles` should get the vehicles, let's keep API functions as simple and lean as much as possible.

cdnninja:
@fuatakgun  I am good to cancel this PR.  Current code base has get_vehicles as it was before and a refresh_vechiles as you mention.   Anything from this PR I should re-create though?

PR #45, opened 2022-02-11
fix(US Kia): Correct temperature setting when starting climate

nprez83:
In the US Kia vehicle and UVO app, the next step below 62 is "LOW" and the next step above 82 is "HIGH". So if you try to set the climate to an actual number lower than 62 or higher than 82, it causes an error that leads to the engine and the rear defroster coming on, but the AC does not. I don't know of an easy way to change the service call itself to accept numbers and also the words "LOW" and "HIGH", so adding the logic check to the API accounts for that and prevents the error.


cdnninja:
Thank you for the contribution!

PR #49, opened 2022-02-13
test: add new env variables

fuatakgun:


fuatakgun:
failures are expected, this PR is supposed to fix failures

PR #51, opened 2022-02-23
feat(EU): Force Location update

cdnninja:
Force EU to update location when odometer has changed since last call.   Interested to hear about thoughts on approach on this.  Currently this is using the cached data call to know when to trigger location calls.

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -52,6 +52,10 @@ class ApiImpl:
         """Get cached vehicle data and update Vehicle instance with it"""
         pass
 
+    def force_refresh_vehicle_location(self, token: Token, vehicle: Vehicle) -> None:
+        """force update vehicle location"""
+        pass
+
     def get_fresh_vehicle_state(self, token: Token, vehicle: Vehicle) -> None:
         pass
 
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -295,7 +295,7 @@ class KiaUvoApiCA(ApiImpl):
 
         return status
 
-    def _get_next_service(self, token: Token, vehicle: Vehicle) -> None:
+    def _get_next_service(self, token: Token, vehicle: Vehicle) -> dict:
         headers = self.API_HEADERS
         headers["accessToken"] = token.access_token
         headers["vehicleId"] = vehicle.id
@@ -306,7 +306,7 @@ class KiaUvoApiCA(ApiImpl):
         response = response["result"]["maintenanceInfo"]
         return response
 
-    def get_location(self, token: Token, vehicle: Vehicle) -> None:
+    def get_location(self, token: Token, vehicle: Vehicle) -> dict:
         url = self.API_URL + "fndmcr"
         headers = self.API_HEADERS
         headers["accessToken"] = token.access_token
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -373,7 +373,7 @@ class KiaUvoApiEU(ApiImpl):
             response = requests.get(url, headers=headers)
             response = response.json()
             _LOGGER.debug(f"{DOMAIN} - Get Vehicle Location {response}")
-            if response["resCode"] != 0000:
+            if response["resCode"] != "0000":
                 raise Exception("No Location Located")
 
         except:
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -322,8 +322,15 @@ class KiaUvoApiEU(ApiImpl):
 
         response = requests.get(url, headers=headers)
         response = response.json()
+        response = response["resMsg"]["vehicleStatusInfo"]
         _LOGGER.debug(f"{DOMAIN} - get_cached_vehicle_status response {response}")
-        return response["resMsg"]["vehicleStatusInfo"]
+        if vehicle.odometer:
+            #To Do:  Confirm date format is being stored correctly and add logic here to confirm we have old data before calling.
+            if vehicle.odometer < get_child_value(response, "odometer.value"):
+                response["vehicleLocation"] = self.get_location(token, vehicle)
+        else:
+            response["vehicleLocation"] = self.get_location(token, vehicle)
+        return response
 
     def force_refresh_vehicle_state(self, token: Token, vehicle: Vehicle) -> None:
         url = self.SPA_API_URL + "vehicles/" + vehicle.id + "/status"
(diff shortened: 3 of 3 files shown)


PierreLevres:
Refreshing location in the EU Kia and Hyundai apps needs a valid token, after 10 minutes it expires and then you need PIN. In the API call I do not see a check on valid token. Is the application handling the api calls supposed to do that?

fuatakgun (commented):
I didn't get why we ended up on this. Earlier we had discussed that location and data have different timestamps to be compared and make a decision to make this call.

In best cases, every time we get a response from force update response, we can get an up to date location data.

While this solution might work for worst case, we should aim to minimize number of calls to kia/hyundai servers, right?

I don't recall where we have discussed this exactly, under discussions?

PierreLevres (changes requested):
Location get in lines 330 and 332 fails when no valid token is present. Since this is used in initialise, a valid token is not set yet.
Plus: WHY refresh location when the call is for a CACHED vehicle status, which already contains a location? Handle refreshing the location seperately.

fuatakgun:
> Location get in lines 330 and 332 fails when no valid token is present. Since this is used in initialise, a valid token is not set yet.
> Plus: WHY refresh location when the call is for a CACHED vehicle status, which already contains a location? Handle refreshing the location seperately.

Sometimes, independent of data source (cached or forced), vehicle location data might be out of sync from vehicle data. I face this issue even using the app, so not specific to our implementation.

My garage has very low cell phone coverage and almost zero gps reception, so it happens time to time to me and some users had reported similar issues.

PierreLevres:
> > Location get in lines 330 and 332 fails when no valid token is present. Since this is used in initialise, a valid token is not set yet.
> > Plus: WHY refresh location when the call is for a CACHED vehicle status, which already contains a location? Handle refreshing the location seperately.
> 
> Sometimes, independent of data source (cached or forced), vehicle location data might be out of sync from vehicle data. I face this issue even using the app, so not specific to our implementation.
> 
> My garage has very low cell phone coverage and almost zero gps reception, so it happens time to time to me and some users had reported similar issues.

Just trying to help, but if you do not want to hear any suggestions, fine by my. I won't make anymore suggestion. 
Before the PR this procedure worked fine, now it isn't. The lack of token checking is the reason.

fuatakgun:
> > > Location get in lines 330 and 332 fails when no valid token is present. Since this is used in initialise, a valid token is not set yet.
> > > Plus: WHY refresh location when the call is for a CACHED vehicle status, which already contains a location? Handle refreshing the location seperately.
> > 
> > Sometimes, independent of data source (cached or forced), vehicle location data might be out of sync from vehicle data. I face this issue even using the app, so not specific to our implementation.
> > 
> > My garage has very low cell phone coverage and almost zero gps reception, so it happens time to time to me and some users had reported similar issues.
> 
> Just trying to help, but if you do not want to hear any suggestions, fine by my. I won't make anymore suggestion. 
> Before the PR this procedure worked fine, now it isn't. The lack of token checking is the reason. 
> Bluelinky and app and my own python implementation are refreshing location just fine. So lack of GPS coverage is not the issue. So yes it is specific to your implementation.

You get me wrong, you are more than welcome and i am happy to have the discussion. I didn't say anything about if the pr should work or not, i am talking about why a solution is needed. I am personally against this one, rather i am proposing to have time based check rather than odometer one. Why did you think in that way?

fuatakgun:
@cdnninja , to keep the api simple and similar as much as possible, i suggest to expose this location api and let the consumers decide to use it.

On home assistant end, we can give this out of the box, if vehicle data time is out of sync with location time, we can initiate a location call to get the latest.

PierreLevres:
> > > > Location get in lines 330 and 332 fails when no valid token is present. Since this is used in initialise, a valid token is not set yet.
> > > > Plus: WHY refresh location when the call is for a CACHED vehicle status, which already contains a location? Handle refreshing the location seperately.
> > > 
> > > 
> > > Sometimes, independent of data source (cached or forced), vehicle location data might be out of sync from vehicle data. I face this issue even using the app, so not specific to our implementation.
> > > My garage has very low cell phone coverage and almost zero gps reception, so it happens time to time to me and some users had reported similar issues.
> > 
> > 
> > Just trying to help, but if you do not want to hear any suggestions, fine by my. I won't make anymore suggestion.
> > Before the PR this procedure worked fine, now it isn't. The lack of token checking is the reason.
> > Bluelinky and app and my own python implementation are refreshing location just fine. So lack of GPS coverage is not the issue. So yes it is specific to your implementation.
> 
> You get me wrong, you are more than welcome and i am happy to have the discussion. I didn't say anything about if the pr should work or not, i am talking about why a solution is needed. I am personally against this one, rather i am proposing to have time based check rather than odometer one. Why did you think in that way?

There should be a location call. But it needs a token gotten with a pin!
There should not be a forced location update  in requesting a information request from cache.
So part of this PR is OK to implement (but is wrongly implemented), part is not.
Please look at the python implementation of JJKostelijk, where a working pin/token and working location is implemented.https://github.com/JanJaapKo/BlUVO/blob/master/bluvo_lib.py

fuatakgun:
@PierreLevres we checked this with @cdnninja earlier and we were able to get location without pin and pin  based token here: https://github.com/fuatakgun/kia_uvo/pull/277

PierreLevres:
> @PierreLevres we checked this with @cdnninja earlier and we were able to get location without pin and pin based token here: [fuatakgun/kia_uvo#277](https://github.com/fuatakgun/kia_uvo/pull/277)

I see your point. Maybe the Pin isn't necessary. But then there is an other error in the code. I think i found it.
I dug deeper into the code, and it errs in get_location:
` if response["responseHeader"]["responseCode"] != 0:
`This gives an exception since response is already in the form of  
`{'retCode': 'S', 'resCode': '0000', 'resMsg': {'gpsDetail': {'coord': {'lat':`
and the responseHeader has been cut off by the response = response.json() command.

IMHO you should test on the get statement before jsonifying it:
```
        try:
            response = requests.get(url, headers=headers)
            _LOGGER.debug(f"{DOMAIN} - Get Vehicle Location {response}")
            if response.status_code != 200:
                raise Exception("No Location Located")
        except:
            _LOGGER.warning(f"{DOMAIN} - Get vehicle location failed")
            response = None
            return response
        else:
            response = response.json()
            return response["resMsg"]["gpsDetail"]
```

cdnninja:
Looks like we have two things going on here.  One is changing the logic as per @fuatakgun second is the call itself doesn't work. 

 Thank you both for the feedback - some of this is a challenge since I am not in the EU.  
 
 I have since updated the code just to fix @PierreLevres's feedback.  I was checking the wrong place for the response code.  As a note two response codes exist - one being the header one being Kia/Hyundai's.  Unfortunately they aren't the same.  @PierreLevres  could you let me know if the call seems to work after my tweak or if you are actually getting a HTTP error? 

I will work on the changes on the logic front next.

fuatakgun:
Before moving forward, i am curious why you changed your mind into odometer based logic rather than timestamp?

cdnninja:
> Before moving forward, i am curious why you changed your mind into odometer based logic rather than timestamp? Maybe, i might have missed something

Mentions here: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/6d69d40eae8d9fdd942b59154e2567ba7edf0baa/hyundai_kia_connect_api/KiaUvoApiEU.py#L328

Ideally I think both odometer and timestamp.  Odometer is a data point we already know - if it didn't change I don't see any value regardless of timestamp in calling the location.  Unless it was moved by a tow truck it hasn't moved.   Timestamp I would think should be secondary - if we know it has moved since last location call check if the "cached" location is updated recently, if so not worth the call.  Part of this also comes from my CA experience - we don't get any location data back in a cached called and full rely on calling location to get any data. 

As I look at the code though I can't use odometer for that type of check if we move away from having this logic linked with get cached update as that method is the only one that know both the old odometer reader and the new before it updates it.

PierreLevres:
Odometer is rounded. So the car could  have moved 900 meters with same odometer. 

Also speed is part of the location object. Speed is reduced to zero when parking the car in the last few meters.  

So yes it is kind of complicated

PierreLevres (changes requested):
Errors with me, unless I quote the rescode test '0000'
line 368: if response["resCode"] != '0000':

PR #83, opened 2022-10-29
feat: Custom Exceptions

cdnninja:


hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -111,7 +113,7 @@ def login(self, username: str, password: str) -> Token:
             authorization_code = self._get_authorization_code_with_form(
                 username, password, cookies
             )
-
+            raise AuthenticationError("Login Failed") 
         if authorization_code is None:
             return None
```
zulufoxtrot, line 118: Raise your `AuthenticationError` here instead of line 116.

Raising it here breaks login. We must have exhausted all options (authorization_code == None) to raise it.

Result should look like this:

```
        except Exception as ex1:
            _LOGGER.debug(f"{DOMAIN} - get_authorization_code_with_redirect_url failed")
            authorization_code = self._get_authorization_code_with_form(
                username, password, cookies
            )

        if authorization_code is None:
            raise AuthenticationError("Login Failed")

```

Unit tests will pass once this is done.

zulufoxtrot (changes requested):
Looks good, just fix the exception. Once this is merged, I'll use your exceptions :)

PR #118, opened 2022-11-10
fix(EU): Convert temp to C

cdnninja:


kalaws:
Actually, that value is not attached to any visible sensor. Seems to correspond to 2 C, and no sensor is showing that or 02H.

cdnninja:
A new issue would be helpful.  This is an old merged code.  Those temp values map to c.

PR #131, opened 2022-11-16
fix(EU):  Use cached location for cached update

cdnninja:


zulufoxtrot:
Works fine for cached data ✅

Wonder if we could also get rid of `_get_location` for forced refreshes.

Right now I can't check if a call to /status returns the location data. I get timeouts on this endpoint. I'll try again later.

cdnninja:
The EU Api is known to be very unreliable.   It has been a challenge with the HA integration.   Been working to try handle the random API failures better on that side so it just tries again 30 minutes later instead of breaking things. I will merge this fix for now.   Here is some history around get location for EU: 

https://github.com/fuatakgun/kia_uvo/issues/47

zulufoxtrot:
Thanks for the history. So it looks like force updates might still need a separate API call to refresh location.
Unless Kia changed something server-side or in the newer car firmwares.
I'll try force refreshes with and without the location call at 2 different locations and see what happens.

PR #134, opened 2022-11-17
fix(EU): Force Update is missing update time

cdnninja:
https://github.com/fuatakgun/kia_uvo/issues/424

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -187,9 +187,12 @@ class KiaUvoApiEU(ApiImpl):
         self._update_vehicle_properties(vehicle, state)
 
     def _update_vehicle_properties(self, vehicle: Vehicle, state: dict) -> None:
-        vehicle.last_updated_at = self.get_last_updated_at(
-            get_child_value(state, "vehicleStatus.time")
-        )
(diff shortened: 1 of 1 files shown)


hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -187,9 +187,12 @@ def force_refresh_vehicle_state(self, token: Token, vehicle: Vehicle) -> None:
...
     def _update_vehicle_properties(self, vehicle: Vehicle, state: dict) -> None:
-        vehicle.last_updated_at = self.get_last_updated_at(
-            get_child_value(state, "vehicleStatus.time")
-        )
+        if get_child_value(state, "vehicleStatus.time"):
+            vehicle.last_updated_at = self.get_last_updated_at(
+                get_child_value(state, "vehicleStatus.time")
+            )
+        else: 
+            vehicle.last_update_at = dt.datetime.now(self.data_timezone)
```
zulufoxtrot, line 195: `_update_vehicle_properties` is also called by `update_vehicle_with_cached_state`.

If we request cached data, replacing `last_updated_at` with the current time will result in wrong data.

If this data is really missing from the response, then maybe it's better to just not fill it.

Do we know when this data is missing? I did some tests today (Kia EU) and `vehicleStatus.time` is present in the response for forced updates.

PR #143, opened 2022-11-22
tests:(Kia EU) Unit tests for _check_response_for_errors

zulufoxtrot:
- Add unit tests for _check_response_for_errors
- remove "stage" argument from exception constructor (unless there's a good reason to leave it?)


zulufoxtrot:
Added a response check for the /vehicles endpoint, forgot it in the previous PR. More endoints might be compatible.

cdnninja:
Maybe we rename the tests to reflect it's only EU? If the same method title is used for other regions it wouldn't conflict?

zulufoxtrot:
Renamed appropriately :)

PR #146, opened 2022-11-24
fix: Logging and reduce calls

cdnninja:
To assist with https://github.com/fuatakgun/kia_uvo/issues/450#issuecomment-1326782443.

I suspect this is a unit or time zone issue.

### DIFF
--- a/hyundai_kia_connect_api/VehicleManager.py
+++ b/hyundai_kia_connect_api/VehicleManager.py
@@ -69,13 +69,12 @@ class VehicleManager:
         for vehicle_id in self.vehicles.keys():
             vehicle: Vehicle = self.get_vehicle(vehicle_id)
             _LOGGER.debug(
-                f"time diff - {(started_at_utc - vehicle.last_updated_at).total_seconds()}"
+                f"{DOMAIN} - time diff - {(started_at_utc - vehicle.last_updated_at).total_seconds()}"
             )
             if (
                 started_at_utc - vehicle.last_updated_at
             ).total_seconds() > force_refresh_interval:
                 self.force_refresh_vehicle_state(vehicle)
-                self.update_vehicle_with_cached_state(vehicle)
             else: 
                 self.update_vehicle_with_cached_state(vehicle)
 


cdnninja:
@zulufoxtrot wouldn't mind a quick review.   As well the link above is the source of why I am tweaking logs.

hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -69,13 +69,12 @@ def check_and_force_update_vehicles(self, force_refresh_interval: int) -> None:
...
             vehicle: Vehicle = self.get_vehicle(vehicle_id)
             _LOGGER.debug(
-                f"time diff - {(started_at_utc - vehicle.last_updated_at).total_seconds()}"
+                f"{DOMAIN} - time diff - {(started_at_utc - vehicle.last_updated_at).total_seconds()}"
             )
             if (
                 started_at_utc - vehicle.last_updated_at
             ).total_seconds() > force_refresh_interval:
                 self.force_refresh_vehicle_state(vehicle)
-                self.update_vehicle_with_cached_state(vehicle)
```
zulufoxtrot, line 78: By removing this line, we acknowledge the fact that all the necessary steps were already handled in force_refresh_vehicle_state, and this call was redundant, right?

zulufoxtrot: the necessary steps being:
- pulling fresh vehicle data from the API (not just triggering a refresh from the server to the vehicle)
- filling the Vehicle object with the API response

sorry if my comment in confusing. I'm just wondering why this line was there to begin with.

cdnninja: Yes*.  Both canada and EU were adjusted to actually take the force update response and process.  The old approach was ignore the response and call another cached state call as above.   

Now USA has not been adjusted for this as they aren't well supported since we don't have testers for it.  I just dropped a PR for Kia USA that would account for this by shifting the above logic into the region.  I will do the same for Hyundai USA which doesn't even have force update implemented - no one has sniffed this traffic before including bluelinky.

zulufoxtrot:
@cdnninja review done.

Just a quick question, somewhat related:
Is there a reason why the kia_uvo project does not use this project (hyundai_kia_connect_api) as a library/dependency? Maybe they have diverged too much or it has neven been considered before?

zulufoxtrot (approved):

cdnninja:
> @cdnninja review done.
> 
> Just a quick question, somewhat related: Is there a reason why the kia_uvo project does not use this project (hyundai_kia_connect_api) as a library/dependency? Maybe they have diverged too much or it has never been considered before?

Version 2.X of kia_uvo does use this.   This python library was founded to support a re-write of kia_uvo.  kia_uvo when I joined it only supported a single car and a single account.  The APIs all has to be re-written to be car aware.    Now this become a much larger undertaking.  @fuatakgun  did the initial break out of this package and showed me the ropes on the new model. 

https://github.com/fuatakgun/kia_uvo/tree/v2.0 is using this API.  I have pre-releases in github for home assistant for it.   This is how I have been gathering feedback on regions and pushing this API forward to where it is today.   As per the home assistant telemetry https://analytics.home-assistant.io/custom_integrations.json  we have over 30 users on this 2.X stream now.  That is for those who opt into telemetry.

zulufoxtrot:
Great! Thanks for explaining 👍

PR #150, opened 2022-11-28
fix: Align EV Charge limits to data model

cdnninja:
BREAKING CHANGE: EvChargeLimits has been removed and now uses ac and dc variables.


cdnninja:
@zulufoxtrot  this isn't done but before I spend too much time on it want to get your take as I see the other PR uses a similar approach of a separate data definition.   I was going to shrink ev charge limits into a parameter.  Reason being is the sensors for kia_uvo use a line that I can't an easy way to get the value from ev limits.  

Specifically this: https://github.com/fuatakgun/kia_uvo/blob/d205462d5bab58ad5cb415b3a3e409dddd315536/custom_components/kia_uvo/number.py#L54

I want it to be generic to work on all car values so the number entity can support other areas.   This numbers file is mostly aligned but that line won't work, partially because the key isn't set right either.  

Interested in thoughts on if ev charge limits should be a parameter or not.

zulufoxtrot:
I'm OK with the proposed format, it's probably what I'd do if written from scratch. I'll submit a code review in a moment.

Just a note: if the limitation comes from kia_uvo, then maybe it's kia_uvo that will need to adapt eventually.
(professor hat on) When writing a library, the data model / architecture should not be too influenced by the constraints of a single project that depends on it. It should be the other way around whenever possible.
Are there other examples of attributes that kia_uvo can't use easily?

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -262,12 +260,14 @@ def ev_estimated_station_charge_duration(self, value):
         self._ev_estimated_station_charge_duration = value[0]
 
     @property
-    def ev_charge_limits(self) -> EvChargeLimits:
-        return self._ev_charge_limits
+    def ev_charge_limits(self) -> int:
```
zulufoxtrot, line 263: It works, but if I were you, I'd just make the attributes `ev_charge_limits_ac` and `ev_charge_limits_ac` public (by removing the leading underscore) and remove the `ev_charge_limits` attribute. I don't think we need this extra level of complexity.

I'm using the term "public" loosely here, but it basically means the attributes can be read/written directly, without the need for getters and setters. It's the right way to do it in simple cases like this.

More on this here: https://towardsdatascience.com/whats-the-meaning-of-single-and-double-underscores-in-python-3d27d57d6bd1

I'm still getting familiar with the data model. It looks like using a dataclass for the Vehicle class was a good idea at the beginning, but the class has gotten bloated with getters and setters and I think some functions should be moved to the Vehicle class. But it's a breaking change, we can discuss later if it's worth doing.

zulufoxtrot (changes requested):

hyundai_kia_connect_api/Vehicle.py
cdnninja: Makes sense. I will adjust to _ac and _dc attributes. I need to go through each API and align this too.

cdnninja:
> I'm OK with the proposed format, it's probably what I'd do if writing from scratch. I'll submit a code review in a moment.
> 
> Just a note: if the limitation comes from kia_uvo, then maybe it's kia_uvo that will need to adapt eventually. (professor hat on) When writing a library, the data model / architecture should not be too influenced by the constraints of a single project that depends on it. It should be the other way around whenever possible. Are there other examples of attributes that kia_uvo can't use easily?

On the kia_uvo topic - yes I do agree. I think this is more my lack of experience trying to figure out the "right" way to do this.  

Most sensors work like this file: https://github.com/fuatakgun/kia_uvo/blob/v2.0/custom_components/kia_uvo/sensor.py which is what I am working to align the new number sensors too. 

In that it provides them all up front and are parsed. If the parsing logic can be adjusted we can change this.  Right now it is clean approach build by @fuatakgun  I just don't want to break it to support more data classes.

hyundai_kia_connect_api/KiaUvoApiCA.py
```diff
@@ -567,13 +567,12 @@ def stop_charge(self, token: Token, vehicle: Vehicle) -> str:
         return response_headers["transactionId"]
     
     def _update_vehicle_properties_charge(self, vehicle: Vehicle, state: dict) -> None:   
-        vehicle.ev_charge_limits = EvChargeLimits(
-            dc=[x['level'] for x in state if x['plugType'] == 0][-1],
+        vehicle.ev_charge_limits = (dc=[x['level'] for x in state if x['plugType'] == 0][-1],
```
zulufoxtrot, line 570: Correct syntax:
If we keep your model:
```
        vehicle.ev_charge_limits = ([x['level'] for x in state if x['plugType'] == 0][-1],
            [x['level'] for x in state if x['plugType'] == 1][-1]
        )
```
If we use my proposed model:
```
vehicle.ev_charge_limit_ac = [x['level'] for x in state if x['plugType'] == 0][-1]
vehicle.ev_charge_limit_dc = [x['level'] for x in state if x['plugType'] == 1][-1]
```

fuatakgun:
Hi all, you are becoming actual owners here, so feel free to change the approach. I am also fine jump into a call or a chat to discuss faster, if you would be interested too.
Big applause to @cdnninja @zulufoxtrot :clap:

cdnninja:
Ready for a review @zulufoxtrot

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -109,6 +105,10 @@ class Vehicle:
 
     # EV fields (EV/PHEV)
 
+
+    ev_charge_limits_dc: int = None
```
zulufoxtrot, line 109: _Context: You may already know this but Python has "weak" typing, meaning any object can be assigned with any type. Type hinting does not prevent it but at least makes the code more readable, allows tools/IDEs to identify typing errors, and helps prevent human errors._

I'm going to nitpick here, but if an attribute/variable is typed as `int` (or any other type other than `NoneType`), then it should not be `None`.

On the other hand, if the API does not always return the value, I agree we should have a default that indicates that we don't have a value, rather than a default with a wrong value (i.e zero).

To be honest I'm not sure how to handle this (probably very common) case. My advice is to use `typing.Union` to allow both `None` and `int`.

More on Union: https://stackoverflow.com/questions/33945261/how-to-specify-multiple-return-types-using-type-hints

so the syntax would be: 
```
ev_charge_limits_dc: typing.Union[int, None] = None
```

zulufoxtrot (changes requested):

hyundai_kia_connect_api/Vehicle.py
cdnninja: We could have None type. We only set vehicle values based on the data we get so if no data received no data set and None remains.  

This also allows us to clearly tell what the car offers in kia_uvo or other use cases as if no value we aren't left guessing if it actually is 0 or if that just the default.  Typing Union looks like the correct way to go.

zulufoxtrot (approved):

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -109,6 +106,10 @@ class Vehicle:
 
     # EV fields (EV/PHEV)
 
+
+    ev_charge_limits_dc: typing.Union[int, None] = None
```
zulufoxtrot, line 110: the `typing` module must be imported

`import typing`

unit tests should pass after that

zulufoxtrot (changes requested):

PR #153, opened 2022-11-30
feat(EU): Charge Port Action

cdnninja:
Unable to test this as I am not in the EU.  I also don't have an EV.

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -113,3 +113,6 @@ class ApiImpl:
         """Sets charge limits. Returns the tracking ID"""
         pass
 
+    def charge_port_action(self, token: Token, vehicle: Vehicle, action: str) -> str:
+        """Opens or closes the charging port of the car. Returns the tracking ID"""        
+
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -113,6 +113,7 @@ class ApiImpl:
         """Sets charge limits. Returns the tracking ID"""
         pass
 
-    def charge_port_action(self, token: Token, vehicle: Vehicle, action: str) -> str:
-        """Opens or closes the charging port of the car. Returns the tracking ID"""        
+    def charge_port_action(self, token: Token, vehicle: Vehicle, action: CHARGE_PORT_ACTION) -> str:
+        """Opens or closes the charging port of the car. Returns the tracking ID"""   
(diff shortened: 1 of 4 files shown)


zulufoxtrot (changes requested):

zulufoxtrot:
Looking good.

So I see this is a V2 endpoint, do we know which cars are compatible?
I don't think my 2020 niro has the ability to remotely operate the charge port.
How do we handle differences in support for cars?
I'm fine with not performing pre checks and letting the API reject the request.

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -113,3 +113,6 @@ def set_charge_limits(self, token: Token, vehicle: Vehicle, ac: int, dc: int)->
         """Sets charge limits. Returns the tracking ID"""
         pass
 
+    def charge_port_action(self, token: Token, vehicle: Vehicle, action: str) -> str:
```
zulufoxtrot, line 116: action: CHARGE_PORT_ACTION

zulufoxtrot (changes requested):

cdnninja:
> The original request is for Hyundai EU, but this PR implements the feature in Kia EU. Is it intended? [Hyundai-Kia-Connect/kia_uvo#243](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/243)

KIA/Hyundai are the same API in the EU just different endpoints.   Only region that doesn't combine the two is USA.

cdnninja:
> Looking good.
> 
> So I see this is a V2 endpoint, do we know which cars are compatible? I don't think my 2020 niro has the ability to remotely operate the charge port. How do we handle differences in support for cars? I'm fine with not performing pre checks and letting the API reject the request.

No unfortunately I am not sure which cars support it and which don't.

zulufoxtrot (approved):

zulufoxtrot:
> No unfortunately I am not sure which cars support it and which don't.

OK.

Maybe at some point we could implement something to let the user know if the car is compatible with the action. I'm not sure it's a good idea to do it at the library level, though. And it requires maintenance (adding car models as they get released).

Feel free to merge, https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/155 is stalled at the moment.

cdnninja:
It wouldn't be bad for us to collect it as we know.  I have commented in the kia_uvo side to validate using your native app.  If it exists in that it will probably work.    The native apps must have a way of knowing though.

PR #155, opened 2022-11-30
Refactor Vehicle class

zulufoxtrot:
Intro
====

This is a pretty extensive refactor but it's simpler than it looks.

The main purpose is to move "business" methods (functions) to the Vehicle class. IMO, a class/object should represent a real-world object whenever possible, and class methods should represent the actions on the real-world object.

Example
======

 if I want to start charging, here's how we do it currently:

```
vehicle = vehicle_manager.get_vehicle("xxx")
vehicle_manager.start_charge(vehicle.id)
```

and here's how we would do it with this PR:

```
vehicle = vehicle_manager.get_vehicle("xxx")
vehicle.start_charge()
```

lower-layer functions and properties updates are still handled by `ApiImpl` and subclasses.

Changes
======

- move vehicle-specific "business" methods (start_climate, lock, unlock, ...) to the Vehicle class whenever possible
- move Token to ApiImpl (it's the only class that needs it, makes more sense to have it there) and remove Token argument from methods
- VehicleManager: use a list of Vehicle objects rather than a list of dicts
- factorize get_last_updated_at (it's pretty much the same between all API implems)
- update unit tests

Attention points
============

- still having some recursive import issues (between ApiImpl, VehicleManager and Vehicle I believe) -- it works as is, but it's not the cleanest
- this PR changes the library's interface, so any apps/libs using it (including kia_uvo) will need updating

TODO
====
- update documentation

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -1,34 +1,20 @@
 import datetime as dt
 import logging
+import re
 from dataclasses import dataclass
 
 import requests
 
 from .const import *
 from .Token import Token
-from .Vehicle import Vehicle, EvChargeLimits
+from .Vehicle import Vehicle, EvChargeLimits, ClimateRequestOptions
 
 from .utils import (
     get_child_value,
-    get_hex_temp_into_index,
-    get_index_into_hex_temp,
 )
 
 _LOGGER = logging.getLogger(__name__)
 
-
-
-@dataclass
-class ClimateRequestOptions:
-    set_temp: float = None
-    duration: int = None
-    defrost: bool = None
-    climate: bool = None
-    heating: int = None
-    front_left_seat: int = None
-    front_right_seat: int = None
-    rear_left_seat: int = None
-    rear_right_seat: int = None
 
 class ApiImpl:
     data_timezone = dt.timezone.utc
@@ -36,40 +22,54 @@ class ApiImpl:
 
     def __init__(self) -> None:
         """Initialize."""
+        self.token: Token = None
 
-    def login(self, username: str, password: str) -> Token:
+    def login(self, username: str, password: str, pin: str):
         """Login into cloud endpoints and return Token"""
         pass
 
-    def get_vehicles(self, token: Token) -> list[Vehicle]:
+    def get_vehicles(self) -> list[Vehicle]:
         """Return all Vehicle instances for a given Token"""
         pass
 
-    def refresh_vehicles(self, token: Token, vehicles: list[Vehicle]) -> None:
+    def refresh_vehicles(self, vehicles: list[Vehicle]) -> None:
         """Refresh the vehicle data provided in get_vehicles. Required for Kia USA as key is session specific"""
         pass
 
     def get_last_updated_at(self, value) -> dt.datetime:
         """Convert last updated value of vehicle into into datetime"""
-        pass
+        if value is not None:
+            m = re.match(r"(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})(\d{2})", value)
+            _LOGGER.debug(f"{DOMAIN} - last_updated_at - before {value}")
+            converted_value = dt.datetime(
+                year=int(m.group(1)),
+                month=int(m.group(2)),
+                day=int(m.group(3)),
+                hour=int(m.group(4)),
+                minute=int(m.group(5)),
+                second=int(m.group(6)),
+                tzinfo=self.data_timezone,
+            )
+            _LOGGER.debug(f"{DOMAIN} - last_updated_at - after {converted_value}")
+            return converted_value
 
-    def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> None:
+    def update_vehicle_with_cached_state(self, vehicle: Vehicle) -> None:
         """Get cached vehicle data and update Vehicle instance with it"""
         pass
 
-    def check_last_action_status(self, token: Token, vehicle: Vehicle, action_id: str) -> bool:
+    def check_last_action_status(self, vehicle: Vehicle, action_id: str) -> bool:
         """Check if a previous placed call was successful. Returns true if complete. False if not.  Does not confirm if successful only confirms if complete"""
         pass
 
-    def force_refresh_vehicle_state(self, token: Token, vehicle: Vehicle) -> None:
+    def force_refresh_vehicle_state(self, vehicle: Vehicle) -> None:
         """Triggers the system to contact the car and get fresh data"""
         pass
 
-    def update_geocoded_location(self, token: Token, vehicle: Vehicle, use_email: bool) -> None:
+    def update_geocoded_location(self, vehicle: Vehicle, use_email: bool) -> None:
 
         email_parameter = ""
         if use_email == True:
-            email_parameter = "&email=" + token.username
+            email_parameter = "&email=" + self.token.username
 
         url = (
             "https://nominatim.openstreetmap.org/reverse?lat="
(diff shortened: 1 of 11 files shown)


zulufoxtrot:
@cdnninja what do you think?

It's not strictly necessary, it doesn't add functionality, it probably breaks kia_uvo... but I think it's cleaner.

If you think it's worth doing, then I'll ask you (and other contributors) to defer other merges until this one is approved and merged. It will prevent conflicts and make my job easier.

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -1,75 +1,75 @@
...
         """Initialize."""
+        self.token: Token = None
 
-    def login(self, username: str, password: str) -> Token:
+    def login(self, username: str, password: str, pin: str):
         """Login into cloud endpoints and return Token"""
         pass
 
-    def get_vehicles(self, token: Token) -> list[Vehicle]:
+    def get_vehicles(self) -> list[Vehicle]:
```
zulufoxtrot, line 31: the Token attribute is now part of the ApiImpl class so no longer needed as a method argument (can be accessed directly by the object methods)

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -1,75 +1,75 @@
...
+    def refresh_vehicles(self, vehicles: list[Vehicle]) -> None:
         """Refresh the vehicle data provided in get_vehicles. Required for Kia USA as key is session specific"""
         pass
 
     def get_last_updated_at(self, value) -> dt.datetime:
         """Convert last updated value of vehicle into into datetime"""
-        pass
-
-    def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> None:
+        if value is not None:
```
zulufoxtrot, line 41: this method is the same for all regional APIs, so I chose to factorize and put it in thep arent class (ApiImpl). The only thing I'm not sure about is this condition (if value is not None). it as present in KiaEU but not in the other classes.

hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -33,93 +33,61 @@ def __init__(self, region: int, brand: int, username: str, password: str, pin: s
...
+                return v
+        raise VehicleNotFoundError("No vehicle found with this ID")
 
     def update_all_vehicles_with_cached_state(self) -> None:
-        for vehicle_id in self.vehicles.keys():
-            self.update_vehicle_with_cached_state(self.get_vehicle(vehicle_id))
+        for vehicle in self.vehicles:
+            vehicle.update_with_cached_state()
 
-    def update_vehicle_with_cached_state(self, vehicle: Vehicle) -> None:
```
zulufoxtrot, line 62: moved to `Vehicle`

hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -33,93 +33,61 @@ def __init__(self, region: int, brand: int, username: str, password: str, pin: s
...
+        if self.api.token.valid_until <= dt.datetime.now(pytz.utc):
             _LOGGER.debug(f"{DOMAIN} - Refresh token expired")
-            self.token: Token = self.api.login(self.username, self.password)
-            self.token.pin = self.pin
-            self.api.refresh_vehicles(self.token, self.vehicles)
+            self.api.refresh_vehicles(self.vehicles)
             return True
         return False
 
-    def start_climate(self, vehicle_id: str, options: ClimateRequestOptions) -> str:
```
zulufoxtrot, line 100: moved to `Vehicle`

cdnninja:
Wow awesome work.  It will take me some time to look at this as align a matching PR in kia_uvo.  Since version is defined in requirements it won't hold this up beyond me ensuring it doesn't break anything critical.   I believe this won't impact kia_uvo's ability to run multiple accounts as today it can do that. 

Also going to send you a repo invite to give you access to merge etc.  You have become a key contributor to it. 

It looks like a few recent changes are reverted in it.  I will hold off on charge port for now but thinking evlimits should be ported over to this.

zulufoxtrot:
Thanks!

> It will take me some time to look at this as align a matching PR in kia_uvo

OK. So I take that as a "go" to keep working on my branch. I'll notify you when it is ready for review. In the meantime, feel free to add comments / ask questions.

> I believe this won't impact kia_uvo's ability to run multiple accounts as today it can do that.

I assume kia_uvo instanciates (=creates) one VehicleManager per account? If so, then it should still work the same. The rest should be pretty easy to adapt.

> It looks like a few recent changes are reverted in it.

Yep. Probably not reverted; just diverged. I started working from an older revision of the `master` branch (probably a few days old) and some changes have been merged into `master` in the meantime, resulting in differences. I will include these new changes into my branch before we merge. But it can sometimes be a little tedious, which is why I ask to hold off other merges in the meantime.

> I will hold off on charge port for now but thinking evlimits should be ported over to this.

Sounds good. I'll make sure to include evlimits (and any other already merged change)

fuatakgun:
Wow, great work.

I should summarize the initial idea behind the design. There are regional API implementation and there is an interface used for API.

Vehicle is the state object that we keep all the information about vehicle but vehicle doesn't know anything about API nor it has access to API.

Vehicle manager should be talking with API interface and updating the vehicle state in the memory.

Now, we are breaking this relationship between api - vehicle manager - vehicle.

We need to look closely at circular dependency issues and more importantly, would there be any benefit of this change from simplicity or architecture perspective?

zulufoxtrot:
@fuatakgun 

you are right to ask these questions, because to be honest I don't have a very clear picture of the implications of my changes.

So if I understand correcly, in the current model:

- Vehicle is just a data class
- VehicleManager is a sort of functional layer
- ApiImpl and subclasses do the heavy lifting: 1) perform requests 2) update Vehicle object with request results

In my model:
- Vehicle becomes an hybrid data/functional class, taking some (but not all) of the functions of VehicleManager
- Vehicle now has access to ApiImpl (the ApiImpl object is passed during instantiation)
- VehicleManager still handles stuff that does not belong in Vehicle (like: perform login, get vehicles list, update all vehicles at once)
- ApiImpl does not really change (It just has the token as an attribute now) / region subclasses still work

So:
- both Vehicle and VM now access ApiImpl
- Apimpl accesses Vehicle

I did get circular dependencies issues, and I'm not sure how to deal with them. They appear to be caused by type hints (the class must be imported in module A to be used in type hints of module B, and vice versa)

What do you think? Is it a dead end?

fuatakgun:
Just to recap, i really appreciate this, I don't want to sound against it.

You are right about Vehicle object, it is a data class without any business logic. Vehicle Manager is the orchestrator between external consumers and Vehicle data class.

From dependency perspective, Vehicle depends on API and API depends on Vehicle wouldn't work. We need to come up with another middle layer if we want this to work out. So, we can make it happen but let's find out possible gains with this.

From OOP perspective, i agree that some of the functions inside Vehicle Manager belong to Vehicle, so it looks technically correct. On the other hand, from lean design principles, each module needs to be less dependent to other modules as much as possible.

A can depend on B, B can depend on C, their dependency degree are both 1. If we change anything on A, only B would be impacted and so.

If we increase the dependency count, any changes would impact greater number of places.

So, we can start from scratch and find a better design with different approach, but it should overall get simplified or got better in terms of design principles.

fuatakgun:
Btw, there was an intention to supply the token object as a parameter but it was postponed with lower priority.

The reasoning is behind external home assistant integration, we shouldn't need to login every time, we can just serialize it and store it in db and use it next time if it is still valid (1 day in EU)

zulufoxtrot:
Thank you for your feedback, I agree.

I'm struggling to find a simple solution. I'm not very well versed in software design.

One possibility: remove any reference to Vehicle in ApiImpl and subclasses. Instead, just make ApiImpl return the raw data / API response and process/assign the data in Vehicle (basically move all _update_vehicle_*** methods to Vehicle). This removes the circular dependency.
The problem with this approach is that not all APIs return identical data. So we would need to implement something in Vehicle (like conditions, or regional subclasses) to make it work. Which would become redundant with ApiImpl subclasses.
And it's a big refactor.

Maybe the original design is the rational choice? Do you see another option?

fuatakgun:
https://docs.google.com/document/d/1PjpdkJKSN4N7fZ7meVW-Sj5tK-UsqJBR4_aay0Y9Rh0/edit?usp=drivesdk

We have set the initial design in this document.

As you had pointed out, because of different responses from regional API implementations, we just decided that APIs should generate vehicle instances but manager would be the entry point for everything.

There might be many other options, but let's agree on what is the limiting factor in the current design and what do we want to solve with a new one?

zulufoxtrot:
OK, I agree. I'll close this PR and submit a different one with the lighter modifications (token, etc).

PR #156, opened 2022-12-01
fix(EU): Steering Wheel Heat

cdnninja:
EU Steering wheel heat comes as int yet I believe is boolean. Not sure why they did this. 

https://github.com/fuatakgun/kia_uvo/issues/291#issuecomment-1036271960

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -289,7 +289,7 @@ class KiaUvoApiEU(ApiImpl):
         )
         if steer_wheel_heat in [0, 2]:
             vehicle.steering_wheel_heater_is_on = False
-        else if steer_wheel_heat == 1:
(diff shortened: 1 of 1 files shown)


zulufoxtrot:
Possible guess...
0 = off
1 = on
2 = not available (not in this trim)

My  eniro does not have steering wheel heating and always returns 2.

zulufoxtrot (approved):

zulufoxtrot:
This PR can be merged, no conflicts with https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/155.

cdnninja:
> Possible guess... 0 = off 1 = on 2 = not available (not in this trim)
> 
> My eniro does not have steering wheel heating and always returns 2.

From the link I posted above one user reports 0 is off when charging 2 is off when not charging.    Now that is just a guess based on other data.  Very odd thing for them to do.

PR #161, opened 2022-12-03
feat(EU): Charge Port Door Status

cdnninja:


hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -351,9 +351,13 @@ def _update_vehicle_properties(self, vehicle: Vehicle, state: dict) -> None:
         vehicle.ev_battery_is_charging = get_child_value(
             state, "vehicleStatus.evStatus.batteryCharge"
         )
+
         vehicle.ev_battery_is_plugged_in = get_child_value(
             state, "vehicleStatus.evStatus.batteryPlugin"
         )
+        vehicle.ev_charge_port_door_is_open = get_child_value(
```
zulufoxtrot, line 358: It looks like the type returned by the api is an `int` (once the JSON is parsed). The declared type for this attribute is `bool`.

I took a look and this is not the only attribute with a type mismatch. It's the same for `battery_is_plugged_in` for instance:

`'batteryPlugin': 2`

In Python, 0 evaluate to False and any number >= 0 evaluates to True when converted to boolean. So, it works so far in conditions etc, but probably by accident.

My recommendation is to:
- first make a list of the possible values for this attribute. It looks like the value can be 2 in some cases (found here: https://forum.iobroker.net/topic/43592/adapter-hyundai-bluelink-oder-kia-uvo/191?lang=en-GB&page=12)
- depending on the possible values:
  - change attribute type to `int`, or even better, use a dedicated `Enum`
  - convert the result of `get_child_value` to `bool` before storing it in the Vehicle attribute

zulufoxtrot (changes requested):

hyundai_kia_connect_api/KiaUvoApiEU.py
cdnninja: Yes I suspect we have a few of those that need adjusting.   The hard part is knowing what the values mean.  We got caught by this on heated seats as an example.   All the car data we saw only returned 0 and 1 so we assumed on and off.  

Thats true for cars that only offer on an off in the app.  Turns out 2 is also off. 3 is heated low etc etc. 

For this one for now how about I adjust it to convert to a bool, open or closed since today we don't know what the other values may mean if these exist at all.

In the future I think we need to define a standard approach to this.  Could have both state and status.  One being a bool one being text enum.  

An example seats could have state, on or off(bool).  Yet status reports Cooling level 2(enum).  

This lets us report the bool for now if we don't know what other values are but add to it in the future.  It also lets downstream consumers of the API represent it in either detail or simple form depending on the goal.  

Thoughts?

zulufoxtrot: > For this one for now how about I adjust it to convert to a bool, open or closed since today we don't know what the other values may mean if these exist at all.

Sounds good.

> In the future I think we need to define a standard approach to this. Could have both state and status. One being a bool one being text enum.

That's a good way to do it, yes.

cdnninja: Done. Please review again.  Currently _is_open is the state.   Status would be added if we find out that it truly has more values beyond 0 and 1.

zulufoxtrot (approved):

PR #162, opened 2022-12-03
feat(EU): Target SOC range

cdnninja:
Add Target range SOC to vehicle

### DIFF
--- a/hyundai_kia_connect_api/Vehicle.py
+++ b/hyundai_kia_connect_api/Vehicle.py
@@ -143,6 +143,14 @@ class Vehicle:
     _ev_estimated_station_charge_duration_value: int = None
     _ev_estimated_station_charge_duration_unit: str = None
 
+    _ev_target_range_charge_AC: typing.Union[float, None] = None
+    _ev_target_range_charge_AC_value: typing.Union[float, None] = None
+    _ev_target_range_charge_AC_unit: typing.Union[str, None] = None
+
+    _ev_target_range_charge_DC: typing.Union[float, None] = None
+    _ev_target_range_charge_DC_value: typing.Union[float, None] = None
+    _ev_target_range_charge_DC_unit: typing.Union[str, None] = None
+
     # IC fields (PHEV/HEV/IC)
     _fuel_driving_range: float = None
     _fuel_driving_range_value: float = None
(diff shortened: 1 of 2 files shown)


cdnninja:
@zulufoxtrot  Interested to see how this one should be represented. It actually is died to the target percentage that already exists.  Just a child attribute.  Thoughts?

zulufoxtrot:
Before I review, what is "target SOC range"?

How does it differ from the target percentage? How is it tied to it?

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -284,6 +292,26 @@ def ev_estimated_station_charge_duration(self, value):
...
 
+    @property
+    def ev_target_range_charge_AC(self):
+        return self._ev_target_range_charge_AC
+
+    @ev_target_range_charge_AC.setter
+    def ev_target_range_charge_AC(self, value):
+        self._ev_target_range_charge_AC_value = value[0]
+        self._ev_target_range_charge_AC_unit = value[1]
+        self._ev_target_range_charge_AC = value[0]
```
zulufoxtrot, line 303: Why use 2 attributes to store the same value?

The 2 attributes in question being _ev_target_range_charge_AC_value and _ev_target_range_charge_AC

Same remark for DC

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -284,6 +292,26 @@ def ev_estimated_station_charge_duration(self, value):
         self._ev_estimated_station_charge_duration_unit = value[1]
         self._ev_estimated_station_charge_duration = value[0]
 
+    @property
+    def ev_target_range_charge_AC(self):
+        return self._ev_target_range_charge_AC
```
zulufoxtrot, line 297: The setter sets 3 properties: _ev_target_range_charge_AC_value, _ev_target_range_charge_AC_unit and _ev_target_range_charge_AC

This getter only returns one of these properties. Is it intentional or are the 2 other getters missing?

---

Overall, I'm not a fan of this getter/setter use, I find it a little too hard to read/use.

My recommendation would be to just have 2 public attributes and do away with the getters/setters:
- ev_target_range_charge_AC_value
- ev_target_range_charge_AC_unit

hyundai_kia_connect_api/Vehicle.py
cdnninja: Following the rest of the file.  Beyond that I don't know.

hyundai_kia_connect_api/Vehicle.py
cdnninja: I am thinking that may be the way to go.    That being said I am going to merge this as is since all range items follow this.   A refactor should clean up the whole thing but I want to keep consistency for now.

PR #172, opened 2022-12-09
fix(EU): Force Update Blanks Values

cdnninja:
Force update deletes all values since it actually doesn't hold its items in "vehicleStatus". Should be the final PR to resolve: https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/468


cdnninja:
@zulufoxtrot  could you test this for me?

PR #187, opened 2022-12-16
fix(Hyundai CA): calculate and set timezone based on Vehicle setting

torbjornaxelsson:
The purpose of this PR is to fix an issue with certain vehicles (Hyundai CA) returning last_updated_at values in their own timezone rather than UTC.

When KiaUvoApiCA.force_refresh_vehicle_state is called, we compare the last_updated_at value from the vehicle with current system time and set a custom timezone to account for the offset, if any. Once the custom timezone is set, the offset will be zero for subsequent calls to force_refresh_vehicle_state.

Tested and working for Hyundai CA.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -2,6 +2,7 @@ import json
 import logging
 import datetime as dt
 import re
+from dateutil.tz import *
 
 import requests
 import pytz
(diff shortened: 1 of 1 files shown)


cdnninja:
Looks to work on my side. It does: `hyundai_kia_connect_api - Offset between vehicle and UTC: 0 hours` as I would expect.

cdnninja:
One item I just realized that with this code it assumes a single time zone per car. As the API class is shared by all cars.  So if two cars were existed with different timezones this would flip back and forth.    Now I doubt that case exists very often.   To solve that we would need to start storing timezone in the vehicle object for this case.   API could set the default in the vehicle. Previously this was a none issue since it was API dependent not car.

torbjornaxelsson:
Good point. I actually found more information on the cause if the timing issue. Apparently there is a bug in that also affects some time based functionality within the vehicle, scheduling, statistics etc. I found a couple of posts from owners complaining about it.

The bug is in the AV head unit of 2019-2021 Kona EVs sold in Canada, quite possibly other Hyundai vehicles. Some dealers have been replacing these head units for owners that complained, but there is no official recall and most vehicles will likely never have the issue fixed. Seems insane to me that they can't just fix it with a firmware update! Oh well.

In conclusion, I agree that timezone should be stored per vehicle. It's quite likely that Hyundai owners have more than one vehicle, in which case this would become an issue. I'll have a look at it.

cdnninja:
Wow good find! Yes if specific to the head unit then this logic would indeed cause problems.   Was thinking if all Hyundais work like this would be a non issue.

PR #189, opened 2022-12-16
Removed not from is_locked

GD-L:
Removed the not status from is_locked to correct inversion of status.

Tested by changing kia_uvo lock.py with:
```python
    @property
    def is_locked(self):
        not_locked = not getattr(self.vehicle, "is_locked")
        return not_locked
        # return getattr(self.vehicle, "is_locked")
```
Found the upstream status and changed the toggle. This was also present in v1.X, which I corrected there as well.


cdnninja:
Thank you for this!

GD-L:
Cheers, and thanks for all your work on this!

PR #193, opened 2022-12-17
fix(CA): timezone stored per vehicle

torbjornaxelsson:
In CA api, store timezone per vehicle to avoid issues for a user who has two vehicles on different timezones.

Tested and working as expected for me.


cdnninja:
Thanks for the help!

PR #195, opened 2022-12-17
feat: Last Action Tracking

torbjornaxelsson:
For information see https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/511


hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -247,6 +247,10 @@ def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> No
             else:
                 self._update_vehicle_drive_info(vehicle, state)
 
+    def check_last_action_status(self, token: Token, vehicle: Vehicle, action_id: str) -> bool:
```
cdnninja, line 250: I would remove this and update this so any API that doesn't have this feature will still work: 

https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/464656e213cc7470b0609f4ed9850ab1229a96c2/hyundai_kia_connect_api/ApiImpl.py#L60-L62

hyundai_kia_connect_api/KiaUvoApiCA.py
```diff
@@ -550,16 +551,19 @@ def check_last_action_status(self, token: Token, vehicle: Vehicle, action_id: st
...
+
+        api_result = response["result"]["transaction"]["apiResult"]
+        last_action_completed = ( api_result != "P" )
         if last_action_completed:
-            action_status = response["result"]["transaction"]["apiStatusCode"]
-            _LOGGER.debug(f"{DOMAIN} - Last action_status: {action_status}")
+            if api_result == "C":
+                state = { "status": response['result']['vehicle'] }
+                _LOGGER.debug(f"{DOMAIN} - Last action completed, updating vehicle state with payload: {state}")
+                self._update_vehicle_properties_base(vehicle, state)
```
cdnninja, line 566: Does the state vehicle response match a full cached call?  If not it will delete values from vehicle.

hyundai_kia_connect_api/KiaUvoApiCA.py
```diff
@@ -424,6 +424,7 @@ def _get_pin_token(self, token: Token, vehicle: Vehicle) -> None:
         )
         _LOGGER.debug(f"{DOMAIN} - Received Pin validation response {response.json()}")
         result = response.json()["result"]
+        token.last_action_pin_token = result["pAuth"]
```
cdnninja, line 427: I would rename this to last_action_pauth.  It isn't actually a full token but rather just a pin authorization.

cdnninja (changes requested):
Overall looks good!  Just a few of my thoughts.

hyundai_kia_connect_api/KiaUvoApiEU.py
torbjornaxelsson: In EU I made check_last_action_status always return True, this should make Coordinator.async_wait_for_action_completed work even if the feature isn't implemented in the API. Maybe you're thinking of a better solution?

hyundai_kia_connect_api/KiaUvoApiCA.py
torbjornaxelsson: Yes, the data (in my case) is identical to a full cached call. This is how the web app updates vehicle data after an action, presumably the mobile app works the same way

hyundai_kia_connect_api/KiaUvoApiCA.py
torbjornaxelsson: I agree. I followed the naming convention of the existing method _get_pin_token, maybe we rename _get_pin_token to _get_pauth as well?

hyundai_kia_connect_api/KiaUvoApiEU.py
cdnninja: More referencing we also have hyundai usa which doesn't have a stub out so by doing it in the parent location it will cover the usa.

hyundai_kia_connect_api/KiaUvoApiCA.py
cdnninja: Awesome!

hyundai_kia_connect_api/KiaUvoApiCA.py
cdnninja: Good call!

zulufoxtrot:
Out of curiosity, how do we know the action ID? I assume it's the header returned by the server:
- `Xid` for USA
- `transactionId` for Canada
?

cdnninja:
Depends on the region.  I don't think EU supports this.  However they way the native Canada app works is after placing an action call such as starting climate it loops calling the status API with the transactionId.  It will do this on a regular interval, say every 15 seconds.  Once it returns true it takes the data to update the app. 

In our api these calls return a string for the regions that support it.

zulufoxtrot:
>  I don't think EU supports this.

I looked at the EU API response: each response has a `msgID` in the JSON body:

`'{"retCode":"S","resCode":"0000","msgId":"d78f1909-686f-4645-952b-1dd0b929687c"}'`

I don't know if it has the same purpose as other regions. But the EU app does have confirmation for actions. They arrive 10-30 seconds after the order is sent, so I assume it works in a similar way to the CA app.
I'd be happy to implement it once this PR is merged.

zulufoxtrot:
Analysis for Kia EU here: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/discussions/272
It looks like it's doable. I'll do another PR for it.

Are we still awaiting changes in this PR?

hyundai_kia_connect_api/KiaUvoApiCA.py
torbjornaxelsson: finally got this done, last_pauth and last_action_id stored per vehicle

torbjornaxelsson:
Hey guys! Sorry for my absence. @zulufoxtro I just made a couple of commits to this PR as well as one to the corresponding PR in kia_uvo at https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/511

hyundai_kia_connect_api/KiaUvoApiEU.py
torbjornaxelsson: Thanks for taking care of that!

hyundai_kia_connect_api/KiaUvoApiCA.py
```diff
@@ -586,34 +589,38 @@ def stop_climate(self, token: Token, vehicle: Vehicle) -> str:
...
         response = response.json()
 
-        last_action_completed = (
-            response["result"]["transaction"]["apiStatusCode"] != "null"
-        )
+        _LOGGER.debug(f"{DOMAIN} - Last action_status API response: {response}")
+
+        api_result = response["result"]["transaction"]["apiResult"]
+        last_action_completed = api_result != "P"
         if last_action_completed:
```
zulufoxtrot, line 609:
```
        last_action_completed = api_result != "P"
        if last_action_completed:
```
Unless I'm reading it wrong, these lines are not necessary.
Testing the value of `api_result` is enough, and you're already doing it in the line that follows.
If we want to test for "P" then I would consider using a simple  `else if` block after the one for "C".

By the way, what does "P" stand for? "Pending"?

zulufoxtrot:
@torbjornaxelsson thank you!

In the meantime, I investigated the EU API.
In the EU API, nn order can go through at least 4 states:
- pending
- successful
- failed
- vehicle timeout

More info here: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/discussions/272#discussioncomment-4811173

I suspect it is the same for other regions.

I think we should probably handle all 4 states, they all have distinct business roles.
Right now `check_last_action_status` only returns a boolean so it can only support 2 states. And I'm not sure we all agree on what "successful" and "completed" mean so it could lead to inconsistencies across implementations.

We could create an OrderStatus Enum class that clearly defines each state, and return that instead of a bool.
We would have to make sure that all regions are compatibles with these states beforehand.

Another idea is to make `check_last_order_status` synchronous, meaning it waits (possibly regularly polling the API) until it gets a response other than "pending".

We could also add a boolean parameter named `synchronous` to `check_last_order_status`. If true then wait, if false then return status even if still pending.

All 3 ideas are compatible with each other. 

What do you think?

zulufoxtrot (changes requested):

hyundai_kia_connect_api/KiaUvoApiCA.py
torbjornaxelsson: The naming convention here is a bit confusing. The official API responds with P if the call has not yet been completed,  probably Pending or maybe Progress. It responds with C if the call completed successfully and F if the call failed. Along with C it also provides the latest vehicle data.

Since our API specification calls for check_last_action_status to respond with a boolean value, I have it responding True for both success and failure, and False for pending.

I should have probably structured it like this, a bit easier to follow:

```
        api_result = response["result"]["transaction"]["apiResult"]
        if api_result == "C":
          state = {"status": response["result"]["vehicle"]}
          self._update_vehicle_properties_base(vehicle, state)
        return api_result != "P"
```

cdnninja: Maybe we adjust away from boolean?

torbjornaxelsson:
> @torbjornaxelsson thank you!
> 
> In the meantime, I investigated the EU API. In the EU API, an order can go through at least 4 states:
> 
> * pending
> * successful
> * failed
> * vehicle timeout
> 
> More info here: [#272 (comment)](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/discussions/272#discussioncomment-4811173)
> 
> I suspect it is the same for other regions.
> 
> I think we should probably handle all 4 states, they all have distinct business roles. Right now `check_last_action_status` only returns a boolean so it can only support 2 states. And I'm not sure we all agree on what "successful" and "completed" mean so it could lead to inconsistencies across implementations.
> 
> We could create an OrderStatus Enum class that clearly defines each state, and return that instead of a bool. We would have to make sure that all regions are compatibles with these states beforehand.
> 
> Another idea is to make `check_last_order_status` synchronous, meaning it waits (possibly regularly polling the API) until it gets a response other than "pending".
> 
> We could also add a boolean parameter named `synchronous` to `check_last_order_status`. If true then wait, if false then return status even if still pending.
> 
> All 3 ideas are compatible with each other.
> 
> What do you think?

That is good news that the EU API seems to respond in a similar way to the Canadian one! This gives me hope that we can come up with a generic solution that works for all countries. Does the EU API also return vehicle data on a successful call?

How to manage all this logic is definitely a bit of a head scratcher. In my initial pull request I have the HA integration check status every X seconds, but I'm starting to think maybe this should all be handled internally by the API client. So that when you initiate a vehicle command from the HA integration (or any other app using our API client), you can supply a callback function that gets called called when the command has completed, with status, payload etc.

An additional challenge that @cdnninja pointed out is that we support multiple vehicles. So even though you can only have one call in progress per vehicle, you could theoretically have multiple calls in progress. I think this is another argument for having the API do the status checking.

Calls can take up to 30 seconds to complete or time out, maybe longer.. so I don't think making the call synchronous would be practical in the HA integration case.

hyundai_kia_connect_api/KiaUvoApiCA.py
torbjornaxelsson: I think that would make sense sense- it sounds like we have four potential valid states: pending, success, fail and timeout.

In CA we get a friendly message along with a failure as well, for example apiFailureReason: "We cannot process your request. Please verify that your vehicle's doors, hood and trunk are closed and locked", it may make sense to pass that along as well.

I don't know if CA has a timeout state, but I suppose in practice we would handle failure and timeout in a similar manner. Do we have anyone who can verify how things look in US?

zulufoxtrot: > it may make sense to pass that along as well.

In EU the message in the response is generic and quite useless: "cannot run command. please check vehicle status" ([source](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/discussions/272#discussion-4792651)). I'm not sure the CA message is much better?
If it's generic I dont think it's useful to return it.

zulufoxtrot:
> An additional challenge that @cdnninja pointed out is that we support multiple vehicles. So even though you can only have one call in progress per vehicle, you could theoretically have multiple calls in progress. I think this is another argument for having the API do the status checking.

Just to clarify, do you mean the API will reject any new action while another action is in progress, across vehicles of the same account?
And you want to prevent simultaneous actions?
For that, we would need to check if an action is pending, either with the server (if it supports it, and assuming we don't necessarily know the previous action's ID) or at the library level, before running any action.
We can think about it, but right now I think it's overkill. It can be handled at the application level. And the worst that can happen is a rejection by the API: in EU, it returns a "duplicate request" error.

zulufoxtrot:
> Does the EU API also return vehicle data on a successful call?

Nope.
However the vehicle does push fresh data and it can be pulled immediately from the server afterwards via `get_cached_vehicle_state`.

>  In my initial pull request I have the HA integration check status every X seconds, but I'm starting to think maybe this should all be handled internally by the API client

> I don't think making the call synchronous would be practical in the HA integration case

I'm not sure I follow. How can we handle checks internally and _not_ be synchronous?

In any case, there is always the option of making synchronous calls optional, with an argument. I think it gives the most flexibility for integrators.

I have begun writing my own implementation (EU + library interfaces). Would you like to take a look at it?

zulufoxtrot:
@torbjornaxelsson , I made a separate PR with the proposed functional changes: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/276
I suggest we review my PR first. Once we have agreed on the functional aspects, we can merge it then finalize yours.

torbjornaxelsson:
> @torbjornaxelsson , I made a separate PR with the proposed functional changes: #276 I suggest we review my PR first. Once we have agreed on the functional aspects, we can merge it then finalize yours.

Awesome! I'll have a look at your PR and your other comments shortly.

torbjornaxelsson:
> > An additional challenge that @cdnninja pointed out is that we support multiple vehicles. So even though you can only have one call in progress per vehicle, you could theoretically have multiple calls in progress. I think this is another argument for having the API do the status checking.
> 
> Just to clarify, do you mean the API will reject any new action while another action is in progress, across vehicles of the same account? And you want to prevent simultaneous actions? For that, we would need to check if an action is pending, either with the server (if it supports it, and assuming we don't necessarily know the previous action's ID) or at the library level, before running any action. We can think about it, but right now I think it's overkill. It can be handled at the application level. And the worst that can happen is a rejection by the API: in EU, it returns a "duplicate request" error.

I meant that the routine that checks status every X seconds could be instantiated per vehicle within our API client, so that there is  vehicles can have actions pending simultaneously and they both get tracked

Issue #145, opened 2022-11-24
Android Bluelink App language is reset to English when using hyundai_kia_connect_api


ZuinigeRijder:
* Hyundai / Kia Connect version: v1.40.11
* Python version: 3.9.13
* Operating System: Windows 10 and Raspberry Pi 11 (bullseye)
* Android Bluelink App: 2.0.7
 
### Description
When the Android Bluelink App language is configured for non-English, by using the hyundai_kia_connect_api the Android Bluelink App is reset to English. So somehow the authentication/login/api does reset the Android Bluelink App to English.

### What I Did

I've looked at the source code of hyundai_kia_connect_api and I thought it's the side effect of the following python statement in KiaUvoApiEU.py, line 68:
```
        if BRANDS[brand] == BRAND_KIA:
            auth_client_id = "f4d531c7-1043-444d-b09a-ad24bd913dd4"
            self.LOGIN_FORM_URL: str = (
                "https://"
                + self.LOGIN_FORM_HOST
                + "/auth/realms/eukiaidm/protocol/openid-connect/auth?client_id="
                + auth_client_id
                + "&scope=openid%20profile%20email%20phone&response_type=code&hkid_session_reset=true&redirect_uri="
                + self.USER_API_URL
                + "integration/redirect/login&ui_locales=en&state=$service_id:$user_id"
            )
        elif BRANDS[brand] == BRAND_HYUNDAI:
            auth_client_id = "64621b96-0f0d-11ec-82a8-0242ac130003"
            self.LOGIN_FORM_URL: str = (
                "https://"
                + self.LOGIN_FORM_HOST
                + "/auth/realms/euhyundaiidm/protocol/openid-connect/auth?client_id="
                + auth_client_id
                + "&scope=openid%20profile%20email%20phone&response_type=code&hkid_session_reset=true&redirect_uri="
                + self.USER_API_URL
                + "integration/redirect/login&ui_locales=en&state=$service_id:$user_id"
            )
```

The statement "ui_locales=en" could cause the side effect? Should it be "ui_locales=de" for Germany and "ui_locales=nl" for the Netherlands? But changing that, still the language setting in the Android Bluelink App is reset to English.

cdnninja:
[...]

PR #201, opened 2022-12-19
feat(EU): Android Bluelink App language reset to English #145

ZuinigeRijder:
Implemented issue:  Android Bluelink App language is reset to English when using hyundai_kia_connect_api #145 

Added optional language to constructor of VehicleManager, defaults to "en" and pass them to the implementation classes constructor and use the language where appropriate. 

Note that I could only test KiaUvoApiEU.py.

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
@@ -44,7 +44,9 @@ class HyundaiBlueLinkAPIUSA(ApiImpl):
         self,
         region: int,
         brand: int,
+        language: str
     ):
+        self.LANGUAGE: str = language
         self.BASE_URL: str = "api.telematics.hyundaiusa.com"
         self.LOGIN_API: str = "https://" + self.BASE_URL + "/v2/ac/"
         self.API_URL: str = "https://" + self.BASE_URL + "/ac/v2/"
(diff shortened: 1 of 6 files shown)


hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
```diff
@@ -44,7 +44,9 @@ def __init__(
         self,
         region: int,
         brand: int,
+        language: str
     ):
+        self.LANGUAGE: str = language
```
cdnninja, line 49: Since this isn't API specific but rather session I would think token would be a better place for it.

hyundai_kia_connect_api/KiaUvoAPIUSA.py
```diff
@@ -105,7 +107,7 @@ def api_headers(self) -> dict:
             "content-type": "application/json;charset=UTF-8",
             "accept": "application/json, text/plain, */*",
             "accept-encoding": "gzip, deflate, br",
-            "accept-language": "en-US,en;q=0.9",
+            "accept-language": self.LANGUAGE + ",en-US,en;q=0.9",
```
cdnninja, line 110: Since regions other than EU haven't been tested I would leave this as is personally.

cdnninja (changes requested):
I am thinking we should move session items to the token.  Thoughts?  Also hoping for some validation on what is passed in.  This would let me just pass the home assistant region into this and let the API handle it if not supported.  Do you know the list of supported languages?

ZuinigeRijder:
> I am thinking we should move session items to the token. Thoughts? Also hoping for some validation on what is passed in. This would let me just pass the home assistant region into this and let the API handle it if not supported. Do you know the list of supported languages?

I think Token is not the most elegant way to solve this, because token is filled by the implementation classes by the login and before that already the language should be known. The language is something enforced by the application, not something you get back from the API. So I think the constructor as parameter is for now the way to go.

hyundai_kia_connect_api/KiaUvoAPIUSA.py
ZuinigeRijder: I reverted the changes for non-EU, but still pass the language in the constructor, so this could be used in the future.

ZuinigeRijder:
> Do you know the list of supported languages?
I see 15 languages in my Bluelink App for Android in Europe. I added the supported languages in the README and check in the Europe implementation for the currently known supported languages and if it is not in the list, it will fallback to "en"

cdnninja:
Awesome! Looks great.  Thanks for doing this!

PR #202, opened 2022-12-21
fix(EU): Car Offline

cdnninja:
When two cars exist in the EU and a car has been offline for an extended period both cars fail due to no data for a single car.

https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/514

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -79,7 +79,7 @@ def _check_response_for_errors(response: dict) -> None:
         "4081": RequestTimeoutError,
         "5031": APIError,
         "5091": RateLimitingError,
-        "5921": NoDataFound
+        "5921": None
     }
 
     if not any(x in response for x in ["retCode", "resCode", "resMsg"]):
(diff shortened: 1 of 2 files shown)


hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -70,14 +70,16 @@ def _check_response_for_errors(response: dict) -> None:
...
     :param response: the API's JSON response
     """
 
     error_code_mapping = {
         "4004": DuplicateRequestError,
         "4081": RequestTimeoutError,
         "5031": APIError,
-        "5091": RateLimitingError
+        "5091": RateLimitingError,
+        "5921": NoDataFound
```
fuatakgun, line 82: Let's set this to None

hyundai_kia_connect_api/exceptions.py
```diff
@@ -18,6 +18,12 @@ class APIError(HyundaiKiaException):
     """
     pass
 
+class NoDataFound(APIError):
```
fuatakgun, line 21: Let's remove this, it is not used

PR #214, opened 2022-12-27
fix(EU): AMPM Time for Charging

cdnninja:
https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/539  @fuatakgun

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -256,8 +256,12 @@ class KiaUvoApiEU(ApiImpl):
             _LOGGER.debug(f"{DOMAIN} - last_updated_at - after {value}")
         return value
 
-    def _get_time_from_string(self, value) -> dt.datetime.time:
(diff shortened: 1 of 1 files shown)


cdnninja:
I don't have time to finish this just yet.  It is complaining around type of for the AM PM.

fuatakgun:
Thanks for raising.

One option, why don't we just pass timeSection to central function and doo all the pm/am logic inside? As we take care of unit based values.

cdnninja:
Yes could adjusted the new get_time_from_string method to do that. I am not sure why get_child_value is already detecting this as as a datetime.time object vs string.  As the following works in a separate test: 

```
import datetime as dt

def _get_time_from_string(value) -> dt.datetime.time:
    if value is not None:
        value = dt.datetime.strptime(value, '%I%M %p').time()
    return value

value = "0800"
value = _get_time_from_string(value + " PM")
print(value)
```

cdnninja:
Apparently the refactor it passes tests. I will merge for now.

Issue #226, opened 2023-01-01
Example code in README wrong, so the server is called too often, were less calls would be sufficient?


ZuinigeRijder:
* Hyundai / Kia Connect version: v1.52.13
* Python version: 3.9.13
* Operating System: Windows 10

### Description

The usage information (README) contains the following example:

````
vm = VehicleManager(region=2, brand=1, username="username@gmail.com", password="password", pin="1234")
vm.check_and_refresh_token()
vm.update_all_vehicles_with_cached_state()
print(vm.vehicles)
````

However, if I do not call "vm.update_all_vehicles_with_cached_state()" I still get the cached values.
So it looks like this call is not needed when you only want the cached server values.

Do I overlook thinks?
Note that I have an IONIQ 5 in Europe, so maybe there are side effects for other REGIONS not doing this call?

Especially because the number of calls to the server are restricted (200 in Europe), having less calls is important. Other regions are even more restricted, so even more important.

### What I Did

I used the following tool [debug.py](https://github.com/ZuinigeRijder/hyundai_kia_connect_monitor/blob/main/debug.py) and commented the line:
```
vm.update_all_vehicles_with_cached_state()
```

The output (except for the debug logging of calls) is exactly the same with or without the call. Note that I always run the script once (do not keep the program alive), so I always need to call: 
```
vm.check_and_refresh_token()
```

**Comment by cdnninja** [comment]
You are correct. I am actually think I am going to remove the call that is run when first setup as this fails if two cars exist and one of them is offline.  (Edge case)   Here is what is done when you initialize the vehicle manager: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/87a87b8131efd66e178bbb5bbaf886fcbe5e4595/hyundai_kia_connect_api/VehicleManager.py#L49-L55

As for check and refresh. That is correct if you only run once you don't need to run that.  It doesn't have harm though as it compares timestamps and only re-authenticates if aging out.

**Comment by cdnninja** [comment]
Here is the issue I am hoping to solve by not force calling cached data on setup: https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/514  It would let me offer "selection" of cars in home assistant and just call for updates on that car.  Getting around strange EU bugs.

**Comment by ZuinigeRijder** [comment]
I see that since 2.0.0 the call is now needed and not done double anymore:

```
vm.update_all_vehicles_with_cached_state()
```

[...]

PR #232, opened 2023-01-02
feat: Enable and Disable Cars

cdnninja:
BREAKING CHANGE:  No longer scans on initial setup in case a car needs to be disabled.


cdnninja:
Related to https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/226

@ZuinigeRijder Are you able to let me know of any concerns around this change?

ZuinigeRijder:
> Related to #226
> 
> @ZuinigeRijder Are you able to let me know of any concerns around this change?

No, I do not see any concerns. Good to go.

PR #235, opened 2023-01-03
fix: Correct typing as we parse data

cdnninja:
Make sure we are typing data away from a string.  @zulufoxtrot I noticed you do some typing in your code which got me thinking that we should attempt to ensure types are correct.    Any thoughts on this?

### DIFF
--- a/hyundai_kia_connect_api/VehicleManager.py
+++ b/hyundai_kia_connect_api/VehicleManager.py
@@ -122,8 +122,8 @@ class VehicleManager:
         return self.api.stop_charge(self.token, self.get_vehicle(vehicle_id))
 
     def set_charge_limits(self, vehicle_id: str, ac: int, dc: int) -> str:
-        ac = int(float(ac), 0)
-        dc = int(float(dc), 0)
+        ac = int(float(ac))
+        dc = int(float(dc))
         return self.api.set_charge_limits(self.token, self.get_vehicle(vehicle_id), ac, dc)
 
     def check_action_status(self, vehicle_id: str, action_id: str):
--- a/hyundai_kia_connect_api/VehicleManager.py
+++ b/hyundai_kia_connect_api/VehicleManager.py
@@ -122,8 +122,8 @@ class VehicleManager:
         return self.api.stop_charge(self.token, self.get_vehicle(vehicle_id))
 
     def set_charge_limits(self, vehicle_id: str, ac: int, dc: int) -> str:
-        ac = int(ac)
-        dc = int(dc)
+        ac = int(float(ac), 0)
+        dc = int(float(dc), 0)
         return self.api.set_charge_limits(self.token, self.get_vehicle(vehicle_id), ac, dc)
 
     def check_action_status(self, vehicle_id: str, action_id: str):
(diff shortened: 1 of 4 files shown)


hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -122,6 +122,8 @@ def stop_charge(self, vehicle_id: str) -> str:
         return self.api.stop_charge(self.token, self.get_vehicle(vehicle_id))
 
     def set_charge_limits(self, vehicle_id: str, ac: int, dc: int) -> str:
+        ac = int(float(ac), 0)
```
zulufoxtrot, line 125: What is the purpose of this line?
I have honestly never seen type casting with 2 arguments, I have no clue what it does.
But in any case, `ac` is already expected to be an int. Why cast it to float then back to int?
Same remark for `dc`

zulufoxtrot:
Hi @cdnninja 
Sorry I've not been around much. Busy with work.

I'm not sure I understand what this PR does. What problem does it solve?

>  noticed you do some typing in your code
I already forgot what I wrote, lol. I did type hinting for sure, but unfortunately it is not sufficient to ensure correct typing.

> we should attempt to ensure types are correct
Do you mean types from the data returned by the API? In that case yes, if the API has "unstable" types, I guess we must implement some validation.
If you mean types in our model/methods, then it depends. Python has weak typing, but manually implementing data validation can quickly make the code messy.
The tools I use to adress this:
- docstrings / type hinting
- unit tests
Still not as robust as a language with strong typing.

cdnninja:
Was meaning to tag @ZuinigeRijder. That being said the feedback is helpful!

hyundai_kia_connect_api/VehicleManager.py
cdnninja: I was finding in home assistant specifically sometimes a float was being passed.  What this does it ensures if 100.0 is passed it doesn't fail to set the charge limit which was happening prior.  float is a way to ensure we don't get an exception so:

```
value = 100.0
value = int(value)
```
Fails yet the above will succeed. 

As for the second value - that is a mistake.  I was trying to make it round off the decimals instead of the float approach.

ZuinigeRijder:
> Was meaning to tag @ZuinigeRijder. That being said the feedback is helpful!

I do not understand the changes:
````
 def set_charge_limits(self, vehicle_id: str, ac: int, dc: int) -> str:
        ac = int(float(ac))
        dc = int(float(dc))
````

ac and dc are already int, so you try to convert to float and then to int. But it was already int in the parameters. The caller of set_charge_limits must convert it to an int when the parameter type is int.

**Comment by fuatakgun** [comment]
Parameter typing do not enforce or convert them, they are used for hinting. In runtime, you can even supply a string to this function

File: hyundai_kia_connect_api/VehicleManager.py (hunk 1 above)
Comment (zulufoxtrot) [reply hyundai_kia_connect_api/VehicleManager.py]: I see. In my opinion, it would be better to update home assistant to pass the correct format, if possible. A library should not compensate for the shortcoming of its dependencies.
If not possible, then I think we should at least update the type hint to "accept" other formats (Union)

> Fails yet the above will succeed.

That is strange. It works for me:

```
i>>> int(100.0)
100
>>> int(100.1)
100
```

Maybe home assistant is passing a string instead:
```
>>> int("100.0")
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '100.0'
>>> int(float("100.0"))
100
```

fuatakgun: Agree on fixing on home assistant side, let's keep the library as lean as much as possible

cdnninja: Home assistant was passing a float but the above code was just to cover all possibilities. I have updated this PR to remove this. I will update HA to ensure an int is always passed.

PR #255, opened 2023-01-09
fix(apiEU): error when get ev_first_departure_time

Adi3000:
when `vehicle.ev_first_departure_time` is fetch _get_time_from_string isn't retreivable because of :
```
File "/usr/local/lib/python3.10/dist-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 587, in _update_vehicle_properties
vehicle.ev_first_departure_time = self._get_time_from_string(
File "/usr/local/lib/python3.10/dist-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 276, in _get_time_from_string
value = dt.datetime.strptime(value, "%I%M %p").time()
TypeError: 'NoneType' object is not callable
```

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -261,20 +261,21 @@ class KiaUvoApiEU(ApiImpl):
             _LOGGER.debug(f"{DOMAIN} - last_updated_at - after {value}")
         return value
 
-    def _get_time_from_string(self, value, timesection) -> dt.datetime.time:
-        if value is not None:
-            lastTwo = int(value[-2:])
+    def _get_time_from_string(self, dateToParse, timesection) -> dt.datetime.time:
+        if dateToParse is not None:
+            lastTwo = int(dateToParse[-2:])
             if lastTwo > 60:
-                value = int(value) + 40
-            if int(value) > 1260:
-                value = dt.datetime.strptime(str(value), "%H%M").time()
+                dateToParse = int(dateToParse) + 40
+            if int(dateToParse) > 1260:
+                dateToParse = dt.datetime.strptime(str(dateToParse), "%H%M").time()
             else:
                 if timesection == 0:
-                    value = str(value) + " AM"
+                    dateToParse = str(dateToParse) + " AM"
                 elif timesection == 1:
-                    value = str(value) + " PM"
-                value = dt.datetime.strptime(value, "%I%M %p").time()
-        return value
+                    dateToParse = str(dateToParse) + " PM"
+                dateToParse = dt.datetime.strptime(dateToParse, "%I%M %p").time()
+        return dateToParse
+
 
     def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> None:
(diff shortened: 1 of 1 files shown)


Adi3000:
Cannot rerun test, but I just changed variable name to avoid a potential override. This test error is a classical timeout

cdnninja:
This commit looks like it just renames the values.   Or am I missing something?

Adi3000:
It just do, and it works fine now. I was having the stack trace :

```
File "/usr/local/lib/python3.10/dist-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 587, in _update_vehicle_properties
vehicle.ev_first_departure_time = self._get_time_from_string(
File "/usr/local/lib/python3.10/dist-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 276, in _get_time_from_string
value = dt.datetime.strptime(value, "%I%M %p").time()
TypeError: 'NoneType' object is not callable
```

As value is the only variable implied, I try something after reading this : 
https://researchdatapod.com/python-nonetype-object-is-not-callable/

cdnninja:
Which function is it overlapping? This code is working for others. What version were you running when you encountered this error?

cdnninja:
Do you have logs of when the failure occurred? I suspect the time section is blank.

zulufoxtrot:
I don't think this PR will solve the original problem.

But I fail to see what causes the problem
Even if we fed the function with an empty, invalid or unset `value`, `strptime` should throw a `ValueError` or a `TypeError`.
In fact, the only way for this exact exception to be thrown is if we try to call a NoneType as a function (trailing `()`).
So it means either `strptime` is None or the generated `datetime` object is None.
And unless something is wrong with the imports, I don't see how it can happen.

@Adi3000 , I'll close this PR and open an issue for you. Sorry!

zulufoxtrot:
After a quick read of this file, I don't see anything wrong but maybe I missed it.
If you know how to use the debugger, you can try to narrow down the problem by putting a breakpoint at this line and evaluating `dt`, `dt.datetime`, `dt.datetime.strptime` and finally `dt.datetime.strptime("0000 AM, "%I%M %p").time()`

Adi3000:
Here the dump from my branch : I forgot to say that I use python 3.10 : 
[Kona_dump_fork_3.10.txt](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/files/10472273/Kona_dump_fork_3.10.txt)
I also install a new VM on another pc with python 3.9 (I use python 3.10 on my server)
[Kona_dump_main_3.9.txt](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/files/10472274/Kona_dump_main_3.9.txt)

We can see some diff on date/time  : they may be separated on 3.10 ^^ Is that something which could break some features ?
The dump is extracted by : 
```
vm = VehicleManager(region=1, brand=2, username=username, password=password, pin=pin)
vm.check_and_refresh_token()
pprint.PrettyPrinter(indent=2,compact=True,sort_dicts=True).pprint(vm.vehicles)
```

cdnninja:
Let's move this over to the issue for more.   I don't think printing of date time would break anything. Did you get an exception?

PR #256, opened 2023-01-10
fix(USA Kia):  Error code 1005

cdnninja:


hyundai_kia_connect_api/KiaUvoAPIUSA.py
```diff
@@ -73,7 +73,10 @@ def request_with_logging_wrapper(*args, **kwargs):
         if (
             response_json["status"]["statusCode"] == 1
             and response_json["status"]["errorType"] == 1
-            and response_json["status"]["errorCode"] == 1003
+            and (
+                response_json["status"]["errorCode"] == 1003
+                or response_json["status"]["errorCode"] == 1005
```
BigTuna94, line 78: this will work, but a more pythonic syntax is: `response_json["status"]["errorCode"] in [1003, 1005]`

BigTuna94 (changes requested):

Issue #267, opened 2023-01-20
Missing ServiceTemporaryUnavailable Exception in KiaUvoApiEU.py


ZuinigeRijder:
* Hyundai / Kia Connect version: v2.4.0
* Python version: 3.9.13
* Operating System: Windows 10

### Description

In my program I want to catch known exceptions and for unknown or too general exceptions I want to print a stacktrace.
Currently I do not distinguish between the different exceptions and so always print a stacktrace.

So I started with the following approach:
````
    def handle_exception(ex: Exception, retries: int, stacktrace=False):
    """
    If an error is found, an exception is raised.
    retCode known values:
    - S: success
    - F: failure
    resCode / resMsg known values:
    - 0000: no error
    - 4004: "Duplicate request"
    - 4081: "Request timeout"
    - 5031: "Unavailable remote control - Service Temporary Unavailable"
    - 5091: "Exceeds number of requests"
    - 5921: "No Data Found v2 - No Data Found v2"
    - 9999: "Undefined Error - Response timeout"
    """
    exception_str = str(ex)
    log(f"Exception: {exception_str}")
    if stacktrace:
        traceback.print_exc()
    retries -= 1
    sleep(retries)
    return retries



        try:

           ..... do the hyundai_kia_connect_api calls.....

        except exceptions.AuthenticationError as ex:
            retries = handle_exception(ex, retries)
        except exceptions.RateLimitingError as ex:
            retries = handle_exception(ex, retries)
        except exceptions.NoDataFound as ex:
            retries = handle_exception(ex, retries)
        except exceptions.DuplicateRequestError as ex:
            retries = handle_exception(ex, retries)
        except exceptions.RequestTimeoutError as ex:
            retries = handle_exception(ex, retries)
        except exceptions.InvalidAPIResponseError as ex:
            retries = handle_exception(ex, retries, True)
        except exceptions.APIError as ex:
            retries = handle_exception(ex, retries, True)
        except exceptions.HyundaiKiaException as ex:
            retries = handle_exception(ex, retries, True)
        except Exception as ex:  # pylint: disable=broad-except
            retries = handle_exception(ex, retries, True)
````

However, it appears that - 5031: "Unavailable remote control - Service Temporary Unavailable" is mapped to the generic **APIError:**

````
"5031": APIError,
````
I would expect a specific exception in exception.py, e.g. ServiceTemporaryUnavailable so I could catch that specific exception too and could add to my code:

````
[...]

PR #269, opened 2023-01-21
feat(EU): Missing ServiceTemporaryUnavailable Exception in KiaUvoApiEU.py #267

ZuinigeRijder:
It appears that in KiaUvoApiEU.py errorcode 5031: "Unavailable remote control - Service Temporary Unavailable" is mapped to the generic APIError:

````
"5031": APIError,
````

I expected a specific exception in exception.py, e.g. ServiceTemporaryUnavailable so I could catch that specific exception too and could add to my code:

````
        except exceptions.ServiceTemporaryUnavailable as ex:
            retries = handle_exception(ex, retries)
````


cdnninja:
Thank you for this!

PR #276, opened 2023-02-02
feat(EU): check action status

zulufoxtrot:
This PR is a follow up from https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/195

I implemented the following changes:
- EU API support: check last action with the /notifications endpoint
- EU API support: modify action functions to return action ID
- rename `check_last_action_status` to `check_action_status`
- synchronous checks (detailed below)

## Synchronous checks

The function takes 2 new arguments:
- `synchronous` (bool): if True, then poll the API until a final state (success/fail/timeout) is returned. if False, then perform just 1 request and return the status immediately, even if still pending. This gives flexibility to developers: they can rely on the library to return a final state, or they can implement this logic at a higher level if they prefer.
-  `timeout` (int, seconds): if call is synchronous, defines how long to poll the API before giving up and returning a TIMEOUT status.

Here's how the function works:
1 - check for action status
2 - if status is still pending, wait 5 seconds then check again in a loop
3 - exit loop once one of these conditions is met: status is final, or the check timed out.

## Tests
Tested today on Kia EU with variations of this code:

```python
        action_id = vehicle_client.vm.start_climate(vehicle_client.vehicle.id, options)
        status = vehicle_client.vm.check_action_status(vehicle_client.vehicle.id, action_id, synchronous=True,
                                                       timeout=60)
        if status == OrderStatus.SUCCESS:
            print "Climate control enabled"
```

It works fine. The final response takes about 20 seconds to arrive.

## What's left
Not included in this PR: implementations for other regions.

## To do
Make sure the states and logic are compatible with all regions.

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -56,7 +56,12 @@ class ApiImpl:
         pass
 
     def check_action_status(
-        self, token: Token, vehicle: Vehicle, action_id: str, synchronous: bool = False, timeout: int = 0
+        self,
+        token: Token,
+        vehicle: Vehicle,
+        action_id: str,
+        synchronous: bool = False,
+        timeout: int = 0,
     ) -> OrderStatus:
         pass
 
@@ -67,7 +72,6 @@ class ApiImpl:
     def update_geocoded_location(
         self, token: Token, vehicle: Vehicle, use_email: bool
     ) -> None:
-
         email_parameter = ""
         if use_email is True:
             email_parameter = "&email=" + token.username
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -6,10 +6,9 @@ from dataclasses import dataclass
 
 import requests
 
-from .const import *
 from .Token import Token
 from .Vehicle import Vehicle
-
+from .const import *
 from .utils import get_child_value
 
 _LOGGER = logging.getLogger(__name__)
@@ -56,11 +55,9 @@ class ApiImpl:
         """Get cached vehicle data and update Vehicle instance with it"""
         pass
 
-    def check_last_action_status(
-        self, token: Token, vehicle: Vehicle, action_id: str
-    ) -> bool:
-        """Check if a previous placed call was successful. Returns true if complete.
-        False if not.  Does not confirm if successful only confirms if complete"""
+    def check_action_status(
+        self, token: Token, vehicle: Vehicle, action_id: str, synchronous: bool = False, timeout: int = 0
+    ) -> OrderStatus:
         pass
 
     def force_refresh_vehicle_state(self, token: Token, vehicle: Vehicle) -> None:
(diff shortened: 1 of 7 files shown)


cdnninja:
I think this looks good.  A good architectural base for us to continue on from.   I suspect we will need a return type of "complete" for the enum in cases where regions don't tell us if successful just that it isn't running anymore.

zulufoxtrot:
@cdnninja thanks! Can I merge?

cdnninja:
Yes lets merge. Thanks for this work!

torbjornaxelsson:
Nice work! A couple of thoughts:

- Maybe OrderStatus should be called ACTION_STATUS for consistency with other constants and that we are referring to vehicle commands as actions?

- The synchronous flag is a nice addition, but it changes the behaviour enough that I would suggest breaking it out into a separate call. Perhaps the logic can be kept entirely in the VehicleManager class, something like this:

```
def track_action_result(
    self,
    token: Token,
    vehicle: Vehicle,
    action_id: str,
    timeout: int = 0,
) -> OrderStatus:
    end_time = datetime.datetime.now() + datetime.timedelta(seconds=timeout)
    while end_time > datetime.datetime.now():
        state = self.api.check_action_status(
            self.token, self.get_vehicle(vehicle_id), action_id, timeout
        )
        if state == OrderStatus.PENDING:
            # state pending: recheck regularly (until we get a final state or exceed the timeout)
            sleep(5)
        else:
            # any other state is final
            return state
    return OrderStatus.TIMEOUT
```

I will align my PR with this implementation, if you agree I can add the above suggested changes as well.

zulufoxtrot:
@torbjornaxelsson 

> Maybe OrderStatus should be called ACTION_STATUS

I'm OK with ActionStatus. Just leave it in CamelCase in compliance with pep8 (it's the other constants that are non-compliant)

> Perhaps the logic can be kept entirely in the VehicleManager class

I'm not against using a separate function, but is there a good reason to move it to the upper layer?

torbjornaxelsson:
> @torbjornaxelsson
> 
> > Maybe OrderStatus should be called ACTION_STATUS
> 
> I'm OK with ActionStatus. Just leave it in CamelCase in compliance with pep8 (it's the other enum constants that are non-compliant)

Good point, let's stick to CamelCase!

> 
> > Perhaps the logic can be kept entirely in the VehicleManager class
> 
> I'm not against using a separate function, but is there a good reason to move it to the upper layer?

The concept of polling for status every X seconds and then reporting the result seems to be universal for all supported regions (US/CA/EU). The only api-specific part is the the api.check_action_status command itself, so I think it makes sense to have the common polling logic at a higher level.

I'm just wrapping up my PR which also includes a separate async version of the polling command. Will let you know when I've pushed the latest changes, would love your feedback.

cdnninja:
Can I just say you are both awesome.  Thank you for all of this.  You are both more advanced than I so really appreciate the work you are putting into this.  This work benefits so many Kia and Hyundai owners.

zulufoxtrot:
> The concept of polling for status every X seconds and then reporting the result seems to be universal for all supported regions (US/CA/EU). The only api-specific part is the the api.check_action_status command itself, so I think it makes sense to have the common polling logic at a higher level.

Sounds good! 👍

torbjornaxelsson:
> Can I just say you are both awesome. Thank you for all of this. You are both more advanced than I so really appreciate the work you are putting into this. This work benefits so many Kia and Hyundai owners. As a note from count home assistant analytics alone show over 900 users and that is opt in which is estimated at 1/3 of the user base. This code is most likely used by well over 3000 users between home assistant and the other items being written against it.

Cheers! You're the one deserving credit though, maintaining a project like this takes skill and dedication. Happy to hear it's being used by so many owners.

PR #285, opened 2023-02-14
USA Hyundai force refresh

### LINKED ISSUE #31: USA Hyundai Missing Force Refresh
USA Hyundai doesn't have force refresh implemented.

https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/9c846174d64cf63ef755feb47e72997b2ff7706d/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py#L367-L368

greglook:
I recently got a new Hyundai (in the US) and found this library via the `kia_uvo` Home Assistant integration. I found that the BlueLink data was often very stale, which makes sense given the lack of a force-refresh implementation. I've taken a shot at adding that to the `HyundaiBlueLinkAPIUSA` file. I incorporated some of the changes from #244 as part of a bit of refactoring of the various internal API call methods.

Fixes #31 

I tested this by:
- Setting up a Home Assistant dev environment
- Manually installing the `kia_uvo` integration
- Manually installing this library in the HA virtualenv with `pip3 install -e ../hyundai_kia_connect_api`
- Starting HA with pip skipping this library with `hass --skip-pip-packages hyundai-kia-connect-api -c config`
- Enabling debug logging per the docs in this repo

I was able to see that the integration was successfully refreshing the data at the configured interval, whereas before it would sit stale:
coding-home-assistant-bluelink-updates

As far as tests go, I tried running `flake8` and it found a ton of existing warnings - I addressed some of the ones that this change touches, but fixing all of them seemed out of scope. I also ran the existing tests, some of which passed and others did not. :man_shrugging: I'm not primarily a Python developer, so if you'd like additional testing, I could use some pointers.

Hoping this looks good and I can start using it in my main Home Assistant instance soon!

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
@@ -126,47 +126,77 @@ class HyundaiBlueLinkAPIUSA(ApiImpl):
             valid_until=valid_until,
         )
 
-    def _get_cached_vehicle_state(self, token: Token, vehicle: Vehicle) -> dict:
+    def _get_vehicle_details(self, token: Token, vehicle: Vehicle):
+        url = self.API_URL + "enrollment/details/" + token.username
+        headers = self._get_authenticated_headers(token)
+        response = self.sessions.get(url, headers=headers)
+        _LOGGER.debug(f"{DOMAIN} - Get Vehicles Response {response.text}")
+        response = response.json()
+        for entry in response["enrolledVehicleDetails"]:
+            entry = entry["vehicleDetails"]
+            if entry["regid"] == vehicle.id:
+                return entry
+
+    def _get_vehicle_status(self, token: Token, vehicle: Vehicle, refresh: bool) -> dict:
         # Vehicle Status Call
         url = self.API_URL + "rcs/rvs/vehicleStatus"
-        headers = self.API_HEADERS
-        headers["accessToken"] = token.access_token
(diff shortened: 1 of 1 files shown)


cdnninja:
Amazing work! Thank you for doing this.  I will merge and release this into HA.  Could I ping you if we see user issues afterward for this region?

hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
```diff
@@ -117,47 +126,79 @@ def login(self, username: str, password: str) -> Token:
...
-                vehicle_status["vehicleLocation"] = None
-        else:
-            vehicle_status["vehicleLocation"] = self.get_location(token, vehicle)
-        return vehicle_status
+                if (
+                    response_json.get("errorCode", 0) == 502
+                    and response_json.get("errorSubCode", "") == "HT_534"
+                ):
+                    _LOGGER.warn(
+                        f"{DOMAIN} - get vehicle location rate limit exceeded."
```
cdnninja, line 189: Longer term we have exceptions this should throw that are defined as standard.  Since we don't have a dev with cars in this region it lags behind.

greglook:
> Could I ping you if we see user issues afterward for this region?

Sure thing, I'm using this code now in my Home Assistant install so I'm certainly motivated to keep it working. :wink: Hopefully I'll have some time to look into the other features missing in the US region as well.

PR #286, opened 2023-02-23
fix:  location_last_updated_at and corrected/suppressed warnings

ZuinigeRijder:
Issue root cause described here:
https://github.com/ZuinigeRijder/hyundai_kia_connect_monitor/issues/48#issuecomment-1436588493

The root cause is in the hyundai_kia_connect_api library.
In Vehicle.py _location_last_set_time is not correctly filled, only for KiaUvoApiEU.py this is done correctly.

You see the missing conversion to a datetime in HyundaiBlueLinkAPIUSA.py.
````
    vehicle.location = (
            get_child_value(state, "vehicleStatus.vehicleLocation.coord.lat"),
            get_child_value(state, "vehicleStatus.vehicleLocation.coord.lon"),
            get_child_value(state, "vehicleStatus.vehicleLocation.time"),
        )
````
The last statement should be converted to a datetime, e.g. with get_last_updated_at.
Note that the same is wrong for KiaUvoAPIUSA.py and KiaUvoApiCA.py.

here are 2 problems in get_last_updated_at for HyundaiBlueLinkAPIUSA.py:
- time can be None as input, and then the error: TypeError: expected string or bytes-like object
- the other datetime last_update_at string replaces characters before calling get_last_updated_at

Changes:

HyundaiBlueLinkAPIUSA.py
KiaUvoAPIUSA.py
KiaUvoApiCA.py
KiaUvoApiEU.py
- Fix location_last_updated_at to convert to correctly datetime or to 2000-01-01
- added doc strings
- moved suppression of pylint warnings to second line, after doc string
- fixed pylint warnings or suppressed specific pylint warnings
- fixed flake8 warnings, "line to long warnings" suppressed at end of line with # noqa

ApiImpl.py
Token.py
Vehicle.py
VehicleManager.py
__init__.py
const.py
exceptions.py
utils.py
- added doc strings
- moved suppression of pylint warnings to second line, after doc string
- fixed pylint warnings or suppressed specific pylint warnings
- fixed flake8 warnings, "line to long warnings" suppressed at end of line with # noqa

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -1,5 +1,6 @@
-# pylint:disable=unnecessary-pass,missing-class-docstring,invalid-name,missing-function-docstring,wildcard-import,unused-wildcard-import,unused-argument
 """ApiImpl.py"""
+# pylint:disable=unnecessary-pass,missing-class-docstring,invalid-name,missing-function-docstring,wildcard-import,unused-wildcard-import,unused-argument,missing-timeout
+
 import datetime as dt
 import logging
 from dataclasses import dataclass
--- a/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
@@ -62,6 +62,9 @@ class HyundaiBlueLinkAPIUSA(ApiImpl):
         ).total_seconds()
         utc_offset_hours = int(utc_offset / 60 / 60)
 
+        origin: str = "https://" + self.BASE_URL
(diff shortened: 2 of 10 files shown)


cdnninja:
Merged some larger changes to the hyundai blue link.  Could you resolve the conflict and then we can merge this one too?

ZuinigeRijder (commented):
I did merge the file from main manually in my local repository and committed again.  Note that I found a problem in the commit to MAIN, namely that in lock_action method, the call to _get_vehicle_headers the parameter vehicle was forgotten. That is also why it is important to use pylint and flake8.

ZuinigeRijder:
I think something did go wrong while merging. So I will create a new Pull Request, so this Pull Request can be removed.

PR #288, opened 2023-02-28
fix(BluelinkUSA): always cached location used

ZuinigeRijder:
HyundaiBlueLinkAPIUSA.py always fills location with cached vehicle location, also when findMyCar is called. This results in the situation that _location_last_set_time is never filled with a datetime for the Bluelink USA implementation!

Another problem encountered is the inconsistent filling of vehicle.timezone. This is only filled by KiaUvoApiCA.py in the constructor of Vehicle. 

Issue analyzed here:
https://github.com/ZuinigeRijder/hyundai_kia_connect_monitor/issues/48#issuecomment-1447709689

Vehicle.timezone is not consistent used. Only in KiaUvoApiCA.py this is set, in the other implementations HyundaiBlueLinkAPIUSA.py, KiaUvoApiEU.py and KiaUvoAPIUSA.py these are NOT initialized when Vehicle is instantiated, while internally they use UTC.

Easy fix is changing the default timezone in Vehicle.py line 95:
However, more consistent is that (also) the implementations fill the timezone when calling the constructor of Vehicle.

Change:
````
timezone: datetime.timezone = None
```` 

Into:
````
timezone: datetime.timezone = datetime.timezone.utc  # default UTC
````

First, in the following get_cached_vehicle_status response you see only the VehicleLocation 'lat' and 'lon', not 'time' (I did a pretty print):
````
DEBUG:hyundai_kia_connect_api.HyundaiBlueLinkAPIUSA:hyundai_kia_connect_api - get_cached_vehicle_status response {
  "hataTID": "zqCe14XFQhOQHksA6zYsZA",
  "vehicleStatus": {
    "dateTime": "2023-02-27T20:24:30Z",
    "acc": false,
    "defrostStatus": "false",
    "transCond": true,
    "doorLockStatus": "false",
    "doorOpen": {
      "frontRight": 0,
      "frontLeft": 0,
      "backLeft": 0,
      "backRight": 0
    },
    "washerFluidStatus": false,
    "battery": {
      "batSoc": 80,
      "sjbDeliveryMode": 1
    },
    "hazardStatus": 0,
    "vehicleLocation": {
      "coord": {
        "lon": -79.123456,
        "type": 0,
        "lat": 39.123456
      }
    },
....
````

Thereafter a call is done, using findMyCar , where 'time' is available (agian in pretty print):
````
DEBUG:urllib3.connectionpool:https://api.telematics.hyundaiusa.com:443 "GET /ac/v2/rcs/rfc/findMyCar HTTP/1.1" 200 None
DEBUG:hyundai_kia_connect_api.HyundaiBlueLinkAPIUSA:hyundai_kia_connect_api - Get Vehicle Location {
  "head": 267,
  "coord": {
    "alt": 0,
    "lon": -79.123456,
    "type": 0,
    "lat": 39.123456
  },
  "accuracy": {
    "pdop": 14,
    "hdop": 8
  },
  "time": "2023-02-27T20:33:27Z",
  "speed": {
    "unit": 1,
    "value": 0
  }
}
````

And there I think I spotted the problem, why _location_last_set_time is not filled. 
There are two state entries in the dictionary:
- vehicleStatus.vehicleLocation (filled by get_cached_vehicle_status response)
- vehicleLocation (filled by _get_vehicle_location)

So _update_vehicle_properties should check first vehicleLocation (if filled), then vehicleStatus.vehicleLocation. 
Because in the current implementation always vehicleStatus.vehicleLocation is used to fill location and location_last_updated_at.

The implementation should be something like:

````
        if get_child_value(state, "vehicleLocation.coord.lat"):
            vehicle.location = (
                get_child_value(state, "vehicleLocation.coord.lat"),
                get_child_value(state, "vehicleLocation.coord.lon"),
                self.get_last_updated_at(
                    get_child_value(state, "vehicleLocation.time")
                ),
            )
        elif get_child_value(state, "vehicleStatus.vehicleLocation.coord.lat"):
            vehicle.location = (
                get_child_value(state, "vehicleStatus.vehicleLocation.coord.lat"),
                get_child_value(state, "vehicleStatus.vehicleLocation.coord.lon"),
                self.get_last_updated_at(
                    get_child_value(state, "vehicleStatus.vehicleLocation.time")
                ),
            )
````

This Pull Request solves both problems.

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
@@ -393,7 +393,16 @@ class HyundaiBlueLinkAPIUSA(ApiImpl):
         vehicle.fuel_level_is_low = get_child_value(state, "vehicleStatus.lowFuelLight")
 
         vehicle.fuel_level = get_child_value(state, "vehicleStatus.fuelLevel")
-        if get_child_value(state, "vehicleStatus.vehicleLocation.coord.lat"):
+
+        if get_child_value(state, "vehicleLocation.coord.lat"):
+            vehicle.location = (
+                get_child_value(state, "vehicleLocation.coord.lat"),
+                get_child_value(state, "vehicleLocation.coord.lon"),
+                self.get_last_updated_at(
+                    get_child_value(state, "vehicleLocation.time")
+                ),
+            )
+        elif get_child_value(state, "vehicleStatus.vehicleLocation.coord.lat"):
             vehicle.location = (
                 get_child_value(state, "vehicleStatus.vehicleLocation.coord.lat"),
                 get_child_value(state, "vehicleStatus.vehicleLocation.coord.lon"),
@@ -442,6 +451,7 @@ class HyundaiBlueLinkAPIUSA(ApiImpl):
                 VIN=entry["vin"],
                 model=entry["modelCode"],
                 registration_date=["enrollmentDate"],
+                timezone=self.data_timezone,
             )
             result.append(vehicle)
 
--- a/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
@@ -204,6 +204,9 @@ class HyundaiBlueLinkAPIUSA(ApiImpl):
                 f"{DOMAIN} - Get vehicle location failed: {e}", exc_info=True
             )
 
+        _LOGGER.debug(f"{DOMAIN} - Get Vehicle Location result is None")
+        return None
+
     def _update_vehicle_properties(self, vehicle: Vehicle, state: dict) -> None:
         vehicle.last_updated_at = self.get_last_updated_at(
             get_child_value(state, "vehicleStatus.dateTime")
@@ -394,15 +397,7 @@ class HyundaiBlueLinkAPIUSA(ApiImpl):
 
         vehicle.fuel_level = get_child_value(state, "vehicleStatus.fuelLevel")
 
-        if get_child_value(state, "vehicleLocation.coord.lat"):
-            vehicle.location = (
-                get_child_value(state, "vehicleLocation.coord.lat"),
-                get_child_value(state, "vehicleLocation.coord.lon"),
-                self.get_last_updated_at(
-                    get_child_value(state, "vehicleLocation.time")
-                ),
-            )
-        elif get_child_value(state, "vehicleStatus.vehicleLocation.coord.lat"):
+        if get_child_value(state, "vehicleStatus.vehicleLocation.coord.lat"):
             vehicle.location = (
                 get_child_value(state, "vehicleStatus.vehicleLocation.coord.lat"),
                 get_child_value(state, "vehicleStatus.vehicleLocation.coord.lon"),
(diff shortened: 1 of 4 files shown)


ZuinigeRijder:
Is the Test Build failure below because kiaconnect.ca is not reachable?
````
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='www.kiaconnect.ca', port=443): Max retries exceeded with url: /tods/api/lgn (Caused by SSLError(SSLError(1, '[SSL: WRONG_SIGNATURE_TYPE] wrong signature type (_ssl.c:1129)')))
````

ZuinigeRijder:
> They changed ssl. So currently ca doesn't work. I need to find time to adjust what type of tls we allow.

Has this something to do with this in the logging:

````
requests.exceptions.SSLError: HTTPSConnectionPool(host='www.kiaconnect.ca', port=443): Max retries exceeded with url: /tods/api/lgn (Caused by SSLError(SSLError(1, '[SSL: WRONG_SIGNATURE_TYPE] wrong signature type (_ssl.c:1129)')))
````
and
````
server_hostname = 'www.kiaconnect.ca', ssl_version = None, ciphers = None
````

Is this maybe the same problem, older TLS version?
https://stackoverflow.com/questions/61631955/python-requests-ssl-error-during-requests

cdnninja:
Yes you are going down the right path.  

USA had this happen months ago.  They rolled back to unsuccessfully ciphers that are typically disabled.  So in some regions we enable ciphers.  Just need to copy that code over from the USA side and test it out.

cdnninja:
#289   is my quick and dirty fix for the Canadian issue.

cdnninja:
Good find.   The other solution for the vehicle location mix match would be changing any line in this where location updates is called. https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/4656a4c4be04141443375db6be538a9bcbe62a02/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py#L413-L427
from:

`state["vehicleLocation"] = self._get_vehicle_location(token, vehicle)`

to: 

`state["vehicleStatus"]["vehicleLocation"] = self._get_vehicle_location(token, vehicle)`

as that is where we expect the data to sit at all times. 

Thoughts?

cdnninja:
Actually my CA issue still exists for Linux. It works on windows.  More work required!

ZuinigeRijder:
> Good find. The other solution for the vehicle location mix match would be changing any line in this where location updates is called.
> 
> https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/4656a4c4be04141443375db6be538a9bcbe62a02/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py#L413-L427
> 
> from:
> 
> `state["vehicleLocation"] = self._get_vehicle_location(token, vehicle)`
> 
> to:
> 
> `state["vehicleStatus"]["vehicleLocation"] = self._get_vehicle_location(token, vehicle)`
> 
> as that is where we expect the data to sit at all times.
> 
> Thoughts?

Yes, I think this is better. But that means more changes, because you only want to overrule vehicleLocation if findMyCar is successful. I also see that the method _get_vehicle_location does not explicit return None, although that is standard behavior of Python. I added a debug statement there too. 

So I implemented your proposal with extra checks and debug statements.

ZuinigeRijder:
@cdnninja If you have time, can you review the changes done after your comment?

PR #289, opened 2023-03-02
fix(CA): Login SSL Error

cdnninja:


ZuinigeRijder:
In HyundaiBlueLinkAPIUSA.py I changed:

````
from requests.packages.urllib3.util.ssl_ import create_urllib3_context
````

Into:
````
from urllib3.util.ssl_ import create_urllib3_context
````

Otherwise the python script would not be interpreted by Visual Studio Code on Windows 10.
So I would propose to do the same in KiaUvoApiCA.py.

PR #295, opened 2023-03-11
fix: add requirements from requirements.txt

fhempy:
Hi,

the setup.py file is missing a lot of the requirements from requirements.txt. I updated the code to include all the requirements from requirements.txt.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,10 +9,12 @@ with open("README.rst") as readme_file:
 
 with open("HISTORY.rst") as history_file:
     history = history_file.read()
+    
+with open("requirements.txt") as f:
+    requirements = f.read().splitlines()
 
 long_description = readme + "\n\n" + history
 long_description = readme
-requirements = ["curlify", "bs4"]
 
 test_requirements = [
     "pytest>=3",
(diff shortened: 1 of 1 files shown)


fhempy:
I just saw that there are too many requirements in the requirements.txt document. It would make sense to cleanup requirements.txt before merge.

As I'm not sure which workflows rely on the requirements.txt I didn't change anything there. Would be great if someone could have a look at it. Thx!

cdnninja:
Do you mean you see some that aren't used?  Not sure what you mean by too many.

fhempy:
requirements.txt normally includes only the packages which are required for the installation. In this case it includes things like coverage, pytest, flake8, etc.. This should be only included in the requirements_dev.txt I think.

cdnninja:
I have shrunk the list by a good bit.   It could probably go a bit further. Could you review and let me know if you are okay with me merging this?

fhempy:
Thanks for the cleanup! I would say that following dependencies could be removed as well
```
Click==8.1.3
watchdog==3.0.0
```
I searched the source code but couldn't find an import of those.

cdnninja:
Done! I will merge now.

fhempy:
Great, thx!

PR #311, opened 2023-04-07
fix(EU):  Login update

cdnninja:


cdnninja:
@fuatakgun @zulufoxtrot  I think this is passing tests for EU.   However I don't think it should.  @fuatakgun account doesn't seem impacted. Could one of you take a look to see if this solves the issue?  Here is the PR the fixed it for bluelinky: https://github.com/Hacksore/bluelinky/commit/a6717a4fc8ea98b4142d3fb1f635dc20d4564bd8#diff-76574552cff659c5fd28ce4da059ce15a2043b1d097f22aec9267d0aac768f01

I haven't added the follow redirect deny yet.  Just the inUserId, which also will need cleanup since we don't need to prep the value either.

ValentinPettmann:
Hi, I tried this PR and I had to comment the now unused lines 1240 and 1243 to get rid of the KeyError: 'int_user_id' message.
now I have a keyError on "redirectUrl" instead
  File "C:\Python310\lib\site-packages\hyundai_kia_connect_api\KiaUvoApiEU.py", line 1257, in _get_authorization_code_with_form
    parsed_url = urlparse(response["redirectUrl"])
KeyError: 'redirectUrl'

Hope that helps :)

PR #313, opened 2023-04-08
Fix Hyundai login EU

Danielhiversen:
This fixes the integration in Home Assistant for my Hyundai Ioniq


fuatakgun (approved):
Thanks, is it working locally for you?

Issue #316, opened 2023-04-08
requirements import causes build to fail


cdnninja:
Release doesn't run due to the requirements change: 

https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/actions/runs/4645478713

I reverted it for now.

cdnninja:
@fhempy can you take a look? I had to revert your change to get builds working.  I reverted as was urgent with the other region issues.

cdnninja:
Yes I looked at fixing it but couldn't understand the issue.  The setup file references the readme in the same location no issue.

PR #328, opened 2023-04-18
fix: use requirements.txt for requirements in setup (#316)

fhempy:
This should fix the issue when building wheels.


cdnninja:
That worked and makes perfect sense.  Thanks!

PR #329, opened 2023-04-23
feat: China region support

toraidl:
Add initial support for Bluelink&Kia Connect

| Features | BlueLink | Kia |
| :--: | :--: | :--:|
|Sigin | ✅|✅|
|Climate on/off  | ✅ |not tested|
|Door lock/unlock | ✅ |not tested|
|Vehicle Status | ✅ |not tested|

Other features needed to be tested, as Sonata being used only has limited features available.


cdnninja:
This is awesome! Thank you so much for doing this.  

I'll take a read through this and provide any feedback in the coming days.

toraidl:
Thank you!
The Bluelink API here is very similar to the EU , so the codes are basically the same as KiaUvoApiEU.py
There are only minor differences, such as  authorization part, and vehicle status response json data structure.
 
Based on the captured HTTPS traffic, it appears that the latest version of the Bluelink app uses the V2 API to both fetch the status of  and control the vehicle. 
However, V1 API is still working and has not been abandoned yet. 

Please take your time to go through the ontent and let me know if you have any questions or feedback.

zulufoxtrot:
Looks good to me. 👍

toraidl:
@fuatakgun  That sounds promising. Initially, my plan was to inherit the EU version, but since I'm not very familiar with Python, I was concerned about introducing bugs. As a result, I opted for the easiest approach.

cdnninja:
As a whole looks good!  For the tests to work you would need to add your username and password as a secret to this repository.   It does have a certain level of risk.  Your call if you would like to do that.  If not the test will need to be removed.

cdnninja:
Are you a home assistant user? I plan to add this region to the ha integration and someone to test or assist would be great!

toraidl:
> Are you a home assistant user? I plan to add this region to the ha integration and someone to test or assist would be great!

Yes, I am HA user.

LitchiCherry:
是不是新车型默认只能用新的api，导致现在无法登录

PR #349, opened 2023-07-01
fix(EU): App ID updates

cdnninja:
This can't be merged until stamp app ids are updated too.  Dependent on https://github.com/neoPix/bluelinky-stamps/pull/8. 

Also will need some tweaks to notification code.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -1084,7 +1084,7 @@ class KiaUvoApiEU(ApiImpl):
         url = self.SPA_API_URL + "notifications/register"
         payload = {
             "pushRegId": registration_id,
-            "pushType": "GCM",
+            "pushType": "APNS",
             "uuid": str(uuid.uuid4()),
         }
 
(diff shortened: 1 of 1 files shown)


hyundai_kia_connect_api/KiaUvoApiEU.py
ZuinigeRijder, line 1: For Hyundai on Android, the APP_ID is still 014d2225-8495-4735-812d-2616334fd15d. If I changed it to 1eba27d2-9a5b-4eba-8ec7-97eb6c62fb51, the code will no longer work with also changing the pushRegIDd.

fuatakgun:
I just initiated the tests with latest stamps generated and tests are still failing for my EU KIA account.
@ZuinigeRijder , you are saying that we should not update Hyundai APP ID, right?

ZuinigeRijder:
> I just initiated the tests with latest stamps generated and tests are still failing for my EU KIA account. @ZuinigeRijder , you are saying that we should not update Hyundai APP ID, right?

Yes, it works without changing the Hyundai App ID. In the Hyundai Bluelink App on Android, the current App ID is used (I sniffed it). Apparently for IOS the Hyundai App ID is different, but the Hyundai Android App ID works perfectly now.

Issue #354, opened 2023-07-03
R3.3.10 makes that EU Hyundai no longer works compared to R3.3.8 because of "pushType": "APNS"


ZuinigeRijder:
* Hyundai / Kia Connect version: R3.3.10
* Python version: 3.9.13
* Operating System: Windows 10

### Description

V3.3.8 fixed the EU Hyundai for this issue: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/348
However, the changes to fix also EU Kia (V3.3.10) makes that it no longer works for EU Hyundai.

### What I Did

"pushType": "APNS" is the problem for EU Hyundai, if I change it back to "pushType": "GCM", then it works again.
Apparently the pushType should be set dependent on EU Kia (APNS) or EU Hyundai (GCM).

Something like:
````
      if BRANDS[brand] == BRAND_KIA:
            self.PUSHTYPE : str = "APNS"
            .....
        elif BRANDS[brand] == BRAND_HYUNDAI:
            self.PUSHTYPE : str = "GCM"

.....
       "pushType": self.PUSHTYPE,
````

Below the exception for EU Hyundai when pushType is APNS.

```
DEBUG:urllib3.connectionpool:https://raw.githubusercontent.com:443 "GET /neoPix/bluelinky-stamps/master/hyundai-014d2225-8495-4735-812d-2616334fd15d.v2.json HTTP/1.1" 200 3107
DEBUG:hyundai_kia_connect_api.KiaUvoApiEU:hyundai_kia_connect_api - get_stamp 2023-07-03 13:01:59.949000+00:00 120000 46 998 46.495045125
[...]

PR #355, opened 2023-07-03
fix(eu): notifications: use GCM for hyundai and APNS for Kia (#354)

zulufoxtrot:
Should fix https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/354

Tests:
No regression on my Kia EU account.
Needs a test for Hyundai.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -122,8 +122,9 @@ class KiaUvoApiEU(ApiImpl):
             _LOGGER.warning(f"Unsupported language: {language}, fallback to en")
             language = "en"  # fallback to English
         self.LANGUAGE: str = language
+        self.brand: int = brand
 
-        if BRANDS[brand] == BRAND_KIA:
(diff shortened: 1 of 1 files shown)


hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -122,21 +122,24 @@ def __init__(self, region: int, brand: int, language: str) -> None:
             _LOGGER.warning(f"Unsupported language: {language}, fallback to en")
             language = "en"  # fallback to English
         self.LANGUAGE: str = language
+        self.brand: int = brand
```
fuatakgun, line 125: do we need to store this here?

fuatakgun (commented):
looks good to me, just one small comment, thanks

hyundai_kia_connect_api/KiaUvoApiEU.py
zulufoxtrot: I need it in `_get_device_id`. Since that method is part of the class, I find it more natural to use class attributes.

PR #356, opened 2023-07-08
fix(EU): Auth error on bad device id

cdnninja:
This will do a few things:

1. Log the error code itself to make this easier for me in the future.
2. Throw an device ID exception which is a child of Authentication Error.   My understanding is when a device ID error occurs we should re-login.  The Home assistant integration does this if it catches auth error.  So this will automatically take care of it in theory.  I can't test as I am not in EU. This should be removed in the future if we find out how to not invalidate the device id. 
https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/672

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -88,6 +88,7 @@ def _check_response_for_errors(response: dict) -> None:
     - 5091: "Exceeds number of requests"
     - 5921: "No Data Found v2 - No Data Found v2"
     - 9999: "Undefined Error - Response timeout"
+    - Unknown:  "Invalid request body - invalid deviceId", relogin oftend required. 
     :param response: the API's JSON response
     """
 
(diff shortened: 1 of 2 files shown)


ZuinigeRijder:
> This will do a few things:
> 
>     1. Log the error code itself to make this easier for me in the future.
> 
>     2. Throw an device ID exception which is a child of Authentication Error.   My understanding is when a device ID error occurs we should re-login.  The Home assistant integration does this if it catches auth error.  So this will automatically take care of it in theory.  I can't test as I am not in EU. This should be removed in the future if we find out how to not invalidate the device id.
>        [[Kia/Hyundai EU] Integration stops working after some time (invalid device id) kia_uvo#672](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/672)

The changes look ok. But....
Should the hyundai_kia_connect_api not try at least once to re-login? Because now all the programs using the kia_connect_api needs to be adapted to re-login? E.g. my monitor tool does not re-login. Is it also clear when the re-login is triggered (root cause), because I have not seen this yet in my monitor tool? 

Note that my monitor tool runs and then exits, so it could be that it does not impact my tool (maybe only an extra retry later)?

cdnninja:
Yes your tool relogins each use.   Our integration keeps running making calls on a schedule for the life of the tokens but a push notification to the native app invalidates our tokens. 

This shouldn't impact your code.

PR #362, opened 2023-07-19
Set ssl.OP_LEGACY_SERVER_CONNECT in order to be able to connect to unsafe SSL servers that OpenSSL 3.0.0 prohibits by default now.

borski:
Update SSL context to enable ssl.OP_LEGACY_SERVER_CONNECT, which isn't in Python yet (see python/cpython#89051, it'll be in 3.12), but in OpenSSL its value is 0x4 in the bitfield.

Fixes Hyundai-Kia-Connect/kia_uvo#678 and Hyundai-Kia-Connect/kia_uvo#679.

Once this is merged, you'll need to do a version bump of this and then bump the version of the required API in Hyundai-Kia-Connect/kia_uvo, obviously.

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
@@ -32,14 +32,20 @@ class cipherAdapter(HTTPAdapter):
     A HTTPAdapter that re-enables poor ciphers required by Hyundai.
     """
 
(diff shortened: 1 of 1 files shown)


fuatakgun:
Thanks for quick fix

cdnninja:
Looks like for some reason the semantic pull CI didn't run on this.   Since it wasn't semantic the release pipeline didn't run.  I will push it manually.

Issue #370, opened 2023-07-29
Infinite loop because of bluelinky-stamps generation outdated?


ZuinigeRijder:
* Hyundai / Kia Connect version:  v3.3.13
* Python version: 3.9+
* Operating System: Windows + Raspberry Pi

### Description

Since today the hyundai_kia_connect_api gets in an infinite loop.
Infinite loop is on line 1086 of KiaUvoApiEU.py: 
        if (position * 100.0) / stamp_count > 90:
            _LOGGER.debug("_get_stamp: (position * 100.0) / stamp_count > 90")
            self.stamps = None
            return self._get_stamp()

Added the _Logger.debug statement there, so recursion till fails.
Am I the only one? Is the bluelinky-stamps generation no longer valid since today?
So there is now an infinite loop....

Looks like the timestamps are 2 days old (it is 2023-07-29 today):
get_stamp 2023-07-27 13:01:47.588000+00:00

### What I Did

Added debug commands, run hyundai_kia_connect_monitor test program debug.py:

```
monitor>python debug.py
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): raw.githubusercontent.com:443
DEBUG:urllib3.connectionpool:https://raw.githubusercontent.com:443 "GET /neoPix/bluelinky-stamps/master/hyundai-014d2225-8495-4735-812d-2616334fd15d.v2.json HTTP/1.1" 200 3112
DEBUG:hyundai_kia_connect_api.KiaUvoApiEU:hyundai_kia_connect_api - get_stamp 2023-07-27 13:01:47.588000+00:00 120000 1335 998 1335.6188112333334
DEBUG:hyundai_kia_connect_api.KiaUvoApiEU:_get_stamp: (position * 100.0) / stamp_count > 90
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): raw.githubusercontent.com:443
DEBUG:urllib3.connectionpool:https://raw.githubusercontent.com:443 "GET /neoPix/bluelinky-stamps/master/hyundai-014d2225-8495-4735-812d-2616334fd15d.v2.json HTTP/1.1" 200 3112
DEBUG:hyundai_kia_connect_api.KiaUvoApiEU:hyundai_kia_connect_api - get_stamp 2023-07-27 13:01:47.588000+00:00 120000 1335 998 1335.6206157083334
[...]

PR #371, opened 2023-07-29
feat(EU): local stamp generation #370

ZuinigeRijder:
Added local stamp generation for EU as discussed on discord.


andig:
Thank you @ZuinigeRijder copied your approach over to evcc. I'm wondering why we've ever relied on server-provided stamps given it's so simple to generate them locally?

ZuinigeRijder:
> Thank you @ZuinigeRijder copied your approach over to [evcc.io](https://evcc.io). I'm wondering why we've ever relied on server-provided stamps given it's so simple to generate them locally?

I took over this code from @PierreLevres so thanks to him.
I asked this question earlier on discord. And @fuatakgun answered with:

Some history: last year, we were running all these with static stamps and everything was ok. Later on. App is updated and kia/hyundai came up with this dynamic stamp generation. We were lucky as @neoPixel22 worked out on a solution and started publishing time limited stamps into github repository. It relies on docker image, running commands on it and writing it daily. Bluelinky and our python library are using it. Later on, inside bluelinky, a beta solution was added to ditch docker image and generate the stamps locally. And @PierreLevres 's code is replicating it.

PR #381, opened 2023-08-09
feat: add australia region support

bitnimble:
This PR adds Australia region support. It's effectively a copy paste of the China + EU regions, with the stamps updated for the Australia CFB and everything else irrelevant removed.

I've tested most of it but can't test all of it myself, so I think this is good enough to merge now and we can merge fixes in as we I test it more day-to-day and get some other people testing as well :)


bitnimble:
There are a couple things that I noticed that I'd like to clean up, but I'll merge them in as separate PRs later 👍

tests/au_login_test.py
```diff
@@ -0,0 +1,22 @@
+import logging
+import os
+
+from hyundai_kia_connect_api.VehicleManager import VehicleManager
+
+
+def test_AU_login():
+    username = os.environ["KIA_AU_USERNAME"]
```
cdnninja, line 8: Are you interesting in adding your login details as a secret for the repo? This will be needed so GitHub tests can run.  All good either way! 

If interested I'll grant you access to the repo to add the secrets.  Each PR after this will run a test to confirm login still functions for regions we have this setup for.

cdnninja (changes requested):

cdnninja:
Amazing work! Thanks for doing this. A few quick comments.

setup.py
```diff
@@ -44,6 +44,6 @@
     test_suite="tests",
     tests_require=test_requirements,
     url="https://github.com/fuatakgun/hyundai_kia_connect_api",
-    version="3.5.0",
+    version="3.6.0",
```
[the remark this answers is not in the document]
bitnimble: Ah gotcha, didn't realise it was automatic! Yep will revert this.

tests/au_login_test.py
bitnimble: Hm... probably not now but I'll consider it later. I'm still testing a lot of stuff locally and I've locked myself out of my account multiple times now due to spamming the login endpoint and hitting the login timeout 😅 so that wouldn't make for a very good test right now. 

Happy to once this is all more stable though!

bitnimble: Will comment this test out for now similar to the CA test.

zappoo:
I still get "Kia is not supported in Australia yet" when I try - has AU support only been added for Hyundai?

PR #384, opened 2023-08-11
fix:  Adjust device ID error and fix url error

cdnninja:
Adjust Device ID error to be API error as it isn't directly auth.


fuatakgun:
Do you want me to try this? Or we know it will work?

cdnninja:
> Do you want me to try this? Or we know it will work?

It should be okay.  This actually caught something in the CI that I did yesterday that I am fixing as well.

PR #386, opened 2023-08-12
fix: add wrapping status layer for Australia vehicle parsing

bitnimble:
Fixes the missing sensors - looks like Australia's API response skips the extra `status` layer. 
I've rewrapped it at the call site instead because I'm looking to start sharing the vehicle parsing code across regions instead. 

With this I'm now able to get the battery level correctly :) 

image


bitnimble:
@cdnninja PTAL, thanks!

PR #387, opened 2023-08-12
fix(AUS): return location correctly

bitnimble:
Also (maybe) fixes "actions" which required more authentication with the PIN, they were previously using the wrong headers.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiAU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiAU.py
@@ -228,8 +228,15 @@ class KiaUvoApiAU(ApiImpl):
         return value
 
     def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> None:
-        state = self._get_cached_vehicle_state(token, vehicle)
-        self._update_vehicle_properties(vehicle, {"status": state})
+        status = self._get_cached_vehicle_state(token, vehicle)
+        location = self._get_location(token, vehicle)
+        self._update_vehicle_properties(
+            vehicle,
+            {
+                "status": status,
+                "vehicleLocation": location,
+            },
+        )
 
         if vehicle.engine_type == ENGINE_TYPES.EV:
             try:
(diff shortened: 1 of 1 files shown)


bitnimble:
@cdnninja (is it possible for me to be able to assign you as a reviewer properly? haha)

hyundai_kia_connect_api/KiaUvoApiAU.py
```diff
@@ -228,8 +228,15 @@ def _get_time_from_string(self, value, timesection) -> dt.datetime.time:
         return value
 
     def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> None:
-        state = self._get_cached_vehicle_state(token, vehicle)
-        self._update_vehicle_properties(vehicle, {"status": state})
+        status = self._get_cached_vehicle_state(token, vehicle)
+        location = self._get_location(token, vehicle)
```
cdnninja, line 232: I think in eu we were able to also get location in a vehicle status as a cached value. This uses up an api call each refresh and wakes the car up which can kill the 12v battery.

cdnninja (changes requested):

hyundai_kia_connect_api/KiaUvoApiAU.py
bitnimble: Ah this was actually a mistake tbh, I accidentally had it here just for testing.

That being said...

I think it's different in Australia - I'm not sure if it's a different available API, or if the backend is actually the same and the Australian app is just calling different endpoints.

1) Our cached vehicle status (`/status/latest`) only returns vehicle data, and does not return odometer or location information. Does your `/status/latest` return location info?
2) In Australia we call `/location/park`, not `/location`
3) `/location/park` resolves instantly with the same latency as `/status/latest`. Also, my car is currently in the basement with no 4G connection - `/status` does not return at all for me, but `/status/latest` and `/location/park` do, which suggests that `/location/park` is pulling a cached location value and not actually waking the car. 

Could you try `/location/park` in EU and seeing if that works for you?

cdnninja: I am actually located in Canada.  @fuatakgun  is EU so I can't directly test this.   How about I merge this for now based on your above theory.   If we find it is hitting the car later please adjust as it will kill your 12v battery and use up API calls.

cdnninja: /status/latest does indeed return location for EU as a note, could be different though.   Each region seems to have tweaks, some larger than others.

bitnimble: Merging for now sounds good 👍 

Do you know if there's a way to know if the car is waking often? I'd like to track it to confirm

cdnninja: Not sure - I find you can tell based on response time like you mentioned above.  I have heard some cars you can also hear relays click when it occurs.

PR #403, opened 2023-09-06
feat: add support for opening and closing windows

bitnimble:
Tested on my Ioniq 6 and seems to work well (minus the whole device ID invalidation thing which is unrelated)

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiAU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiAU.py
@@ -696,7 +696,7 @@ class KiaUvoApiAU(ApiImpl):
         _check_response_for_errors(response)
         return response["msgId"]
 
-    def set_windows(
+    def set_windows_state(
         self, token: Token, vehicle: Vehicle, options: WindowRequestOptions
     ) -> str:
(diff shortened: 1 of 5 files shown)


hyundai_kia_connect_api/KiaUvoApiAU.py
```diff
@@ -695,6 +696,25 @@ def lock_action(
...
+    def set_windows_state(
+        self, token: Token, vehicle: Vehicle, options: WindowRequestOptions
+    ) -> str:
+        url = self.SPA_API_URL_V2 + "vehicles/" + vehicle.id + "/control/windowcurtain"
+
+        payload = {
+            "backLeft": options.back_left,
+            "backRight": options.back_right,
+            "frontLeft": options.front_left,
+            "frontRight": options.front_right,
```
bitnimble, line 708: The API schema seems to theoretically support it, but I didn't test opening different combinations of windows yet - only all open, or all closed. 

The Bluelink app only has open/close all windows as well, so... I'll test that later and see how it goes. Will adjust the function signature if it's not possible to do combinations.

dbolger:
Would these changes be better off in the BlueLink API?

PR #404, opened 2023-09-07
refactor(Hyundai USA): create Hyundai start/stop engine API methods

dbolger:
I've changed the start/stop climate methods to more intuitive start/stop engine methods.
`engine_control_action` behaves similarly to `lock_action` - that is, whatever action is passed in is interpreted.

I wrote a test around this and the US Login, and both worked. I still need to test `start_climate` and `stop_climate` to ensure no regressions exist.

Vehicle used to test: 2020 Hyundai Santa Fe


dbolger:
Looks like the build failed due to my environment variables. Looking into it

dbolger:
I see that `python-tests.yml` is configured to use some defined secrets. I don't feel comfortable sharing my secrets here, so I'm commenting on the test for now.

dbolger:
Run black failed for something out of the reach of my code

cdnninja:
Start_climate is named the way it is because it is used by EVs but the same endpoint works for ICE.  In both cases it's starting climate.   I do understand the confusion.   

As a note the way it currently is created it would be a breaking change for usa since integration using this library expect start_climate to work for ICE and EV.

dbolger:
We could name it something more generic like `VehicleManager.start(...)` for both types of vehicles

cdnninja:
You need climate parameters otherwise it will error out.    This depends on the country though.    If a strong desire to rename I am okay with this if done across the board in a refactor as a breaking change.    Seems like a alot of work considering impacts on others who use this too.  Could do something to keep support if needed.

dbolger:
I haven't forgotten about this PR. My ideal approach was not to introduce breaking changes immediately but rather to deprecate the climate methods and eventually kill them. This would give consumers of the API some time to update their code and not put unneeded stress on developers.

dbolger:
There are a lot of changes that need to be caught up - I'm going to start with a fresh fork.

PR #413, opened 2023-09-25
Enable Genesis path for USA

pkarimov:
This change enables handling of Genesis vehicles in the USA region to be handled by HyundaiBlueLinkAPIUSA module. In limited testing this was successful in retrieving vehicle details.


fuatakgun (approved):
Thanks, looks good to me

PR #415, opened 2023-10-04
fix(eu): Add 12h using datetime

agx:
Rather than relying on am/pm parsing (which can fail based on locale) rather add 12h using the datetime module.

I hit this when using this module embedded in a tiny [Rust application](https://gitlab.gnome.org/guidog/phosh-ev/)
with pyo3.

Thanks a lot for providing this very useful module!


cdnninja:
Thank you for this!

PR #417, opened 2023-10-05
feat(EU): add power_regenerated

WillemD61:


cdnninja:
Quick question is this total power generated?  As in a running tally?  Just trying to keep naming the same as the others.

WillemD61:
yes, this is total power regenerated.

cdnninja:
Thank you for this!

PR #419, opened 2023-10-05
fix(EU): correction to get total power consumed from driving info

WillemD61:
total power consumed is obtained from drivingInfo, not drivingInfoDetail

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -1035,18 +1035,19 @@ class KiaUvoApiEU(ApiImpl):
             drivingInfo = responseAlltime["resMsg"]["drivingInfo"][0]
 
             drivingInfo["dailyStats"] = []
-            for day in response30d["resMsg"]["drivingInfoDetail"]:
-                processedDay = DailyDrivingStats(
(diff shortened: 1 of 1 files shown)


cdnninja:
@ZuinigeRijder could you review this? I think this may impact your integration.

ZuinigeRijder:
> @ZuinigeRijder could you review this? I think this may impact your integration.

I will have a closer look later today or tomorrow.

ZuinigeRijder (commented):
I would expect on line 1038 an extra check to avoid an error on possibly missing drivingInfoDetail:
     if get_child_value(responseAlltime, "resMsg.drivingInfoDetail.0"):

WillemD61:
I assume you want to have that check so that the calculation of daily driving stats that follows is succesfull.

But therefore you will have to test for presence of drivingInfoDetail in the response30d, not in the responseAlltime.

So line 1038 should be
if get_child_value(response30d, "resMsg.drivingInfoDetail.0"):
And line 1039 to 1050 should be subject to this condition.

If agreed, I will make the additional change in this PR.

WillemD61:
adjusted now included extra if statement

ZuinigeRijder (approved):

fuatakgun (approved):
LGTM

PR #432, opened 2023-10-26
fix(USA): Cast set_temp to string

zphaze:


zphaze:
Home Assistant still errors, from debug logs appears that it sends set temp as a float instead of string that the Hyundai API wants.

Issue #440, opened 2023-11-06
No `totalPwrCsp` available


jwillemsen:
* Hyundai  Connect version: 2.0.16
* Hyundai Kona 2024 Full Electric

### Description

I see in the log that `totalPwrCsp` is not available. When using the code from your example 

```
vm.update_all_vehicles_with_cached_state()
print(vm.vehicles)
```

I get the following errror

```
Traceback (most recent call last):
  File "/home/johnny/ACE/github/hyundai_kia_connect_api/hyundai_kia_connect_api/KiaUvoApiEU.py", line 330, in update_vehicle_with_cached_state
    state = self._get_driving_info(token, vehicle)
  File "/home/johnny/ACE/github/hyundai_kia_connect_api/hyundai_kia_connect_api/KiaUvoApiEU.py", line 1059, in _get_driving_info
    drivingInfoItem["totalPwrCsp"]
KeyError: 'totalPwrCsp'
```

The data printed is the following (removed id and registration data)

```
[...]

PR #445, opened 2023-11-24
fix(EU): Driving Info Error

cdnninja:
Safer method to get values.


cdnninja:
@WillemD61 could you confirm this code works for you before I merge? This is to try resolve https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/440

WillemD61:
I tested it and it still works for me.

But I doubt whether this is going to solve #440 since the problem is not about getting the right fields out of the total response but getting a proper response in the first place. The response that [](https://github.com/jwillemsen) received was valid but totally empty, so you will still not get a value for the field totalPwrCsp.

Focus should be on getting a response with contents filled first. Once that is done the fields can be picked out.

cdnninja:
Agreed something else is going on data wise.  This will just handle the exception in theory.

PR #459, opened 2023-12-23
adding function for heat/ventilated seats in USA region

chasebolt:
i dont know how to properly test this. but I am 99.999% confident this is good.

i did test this with curl, so I know the API json data is correct and works.


cdnninja:
Wow this is awesome!!  Thank you! 

Two ways to test:

1. Using code from readme you can run this just as python in just a few commands. 
2. I merge it which generates a release version number.  Put the number into your manifest file in home assistant custom component folder.  Restart HA and give it a go. 

For this one I'll merge since low risk and others have to update version in manifest to take effect.   Let me know if tests okay.  If so we bump the version in the other repo to role this out.

cdnninja:
I missed title to cause a release so I forced one.  3.11.0 is being built right now and should be good by the time you read this.

PR #464, opened 2023-12-23
fix(USA): remap seat values

chasebolt:
options value passed is not the expected value by the api. creating a new mapping with the values adjusted to what is expected.


chasebolt:
@cdnninja this one didnt create a release version

cdnninja:
Oops you are correct.  Fixed. For next maybe try have that title as the commit title as well as the PR.

PR #467, opened 2023-12-26
fix: Add login token lifetime to consts and use it in all API clients

freaksdotcom:
fix: Add the token lifetime to `const.py` and use the value across all API endpoints.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoAPIUSA.py
+++ b/hyundai_kia_connect_api/KiaUvoAPIUSA.py
@@ -18,11 +18,12 @@ from .ApiImpl import ApiImpl, ClimateRequestOptions
 from .Token import Token
 from .Vehicle import Vehicle
 from .const import (
-    DOMAIN,
-    VEHICLE_LOCK_ACTION,
-    TEMPERATURE_UNITS,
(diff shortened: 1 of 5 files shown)


cdnninja:
I don't think the USA token is valid for 24 hours.  Have you tested to see if they are now valid for that long?    Historically they were much shorter than other regions.

freaksdotcom:
> I don't think the USA token is valid for 24 hours. Have you tested to see if they are now valid for that long? Historically they were much shorter than other regions.

Can you point out where this was discussed? I'd like to know how the lifetime was determined.

I've been running it on my local instance for a few hours and the integration has not encountered any problems with fetching data, but it's been less than the full 23 hours.

During the time that it's been running, the car has not been exhibiting the periodic wake ups that I'm trying to fix.

cdnninja:
Looking back at some of the code I think a short token was used to cover when the token was invalidated. Back when this was written if you used the mobile app it would invalidate the token.  The code now can recover from this.   USA Kia code is largely unmaintained as we don't have very many active USA devs here.  Have had some recent interest though! 

Since you tested out I am happy to merge this.   Worst case we revert!

PR #469, opened 2023-12-30
fix: convert heat/cool seating option from string to int

chasebolt:
these values are set as a string when you configure these properties with the UI. this is a dirty hack for now. is it possible to have the UI configurator set these are ints instead of strings? then we couldn't need this.


cdnninja:
Thoughts on doing this in Vehicle manager so all regions benefit? I have run into this on the Canadian region too.

chasebolt:
is it possible to have HA return these options values as ints instead of strings? that would be ideal

cdnninja:
Yes could just move this to the service call.   Good point.

cdnninja:
Thoughts on this? https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/793

Issue #473, opened 2024-01-05
Feature: [US] Expose EV Trip Details


DanielRussell:
* Hyundai / Kia Connect version: 5.1.8
* Python version: 3.11
* Operating System: macOS / Linux

### Description

The MyHyundai app shows some details about recent trips (Vehicle Status -> EV Trip Details).  I would like to expose these stats to arbitrary applications, HomeAssistant, etc.

### What I Did

Setup Charles Proxy.  Launch the MyHyundai app and navigate to the EV Trip Details.  Copy the cURL request from Charles and verify its functionality via the CLI.  Then figure out how to add this API call and resulting data to the codebase.

ZuinigeRijder:
[See this issue](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/248), where I integrated this in KiaUvoApiEU.py for Hyundai Bluelink and Kia Uvo. Maybe this helps with a similar approach?

PR #474, opened 2024-01-05
feat(USA): Expose EV Trip Details (#473)

DanielRussell:
The US version of the MyHyundai app shows some details about the most recent trips; extract and expose this.

This change uses the same API call to pull the trip data and add to `vehicle.daily_stats`. The data in the Hyundai API response matches exactly with the existing EU `Vehicle.DailyDrivingStats`, so I reused that data structure despite the slight name difference.

A simple example:

```python
#!/usr/bin/env python3

import pprint

vm = VehicleManager(region=3, brand=2, username="me@example.com", password="hunter2", pin="1234")
vm.check_and_refresh_token()

for v in vm.vehicles:
    pprint.pprint(vm.vehicles[v].name)
    for s in vm.vehicles[v].daily_stats:
        # Avoid DIV0
        distance = (s.distance if s.distance > 0 else 1)
        # Total watt hours per mile
        whpmi = int(s.total_consumed / distance)
        # "Overhead" (anything except engine) Watt hours per mile
        whpmi_overhead = int((s.climate_consumption + s.onboard_electronics_consumption + s.onboard_electronics_consumption) / distance)
        # Regeneration as a percentage of total consumption
        regen_pct = int((s.regenerated_energy / s.total_consumed) * 100)

        print(f"{s.date} - {s.distance} miles - {s.total_consumed} watt hours - {whpmi} Wh/mi - {whpmi_overhead} Wh/mi overhead - {regen_pct}% regen pct")

    pprint.pprint(vm.vehicles[v].daily_stats)
```

And output:

```
$ python3 example.py
hyundai_kia_connect_api - get vehicle location rate limit exceeded.
'2023 IONIQ 5'
2024-01-04 21:11:45 - 1 miles - 509 watt hours - 509 Wh/mi - 320 Wh/mi overhead - 82% regen pct
2024-01-04 20:44:01 - 1 miles - 751 watt hours - 751 Wh/mi - 335 Wh/mi overhead - 72% regen pct
```

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
@@ -21,7 +21,10 @@ from .const import (
 from .utils import get_child_value, get_float
 from .ApiImpl import ApiImpl, ClimateRequestOptions
 from .Token import Token
-from .Vehicle import Vehicle
+from .Vehicle import (
+    DailyDrivingStats,
+    Vehicle
+)
 
 
 CIPHERS = "DEFAULT@SECLEVEL=1"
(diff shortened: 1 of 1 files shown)


cdnninja:
Awesome! Thanks for doing this.  If it appears to be an EV only call maybe it is worth adding an if statement to only place the api call it engine type is ev?

cdnninja:
Great! Thanks for doing this.

PR #475, opened 2024-01-08
fix: Fix temperature range to include 82 degrees as max

nprez83:
I changed the temperature_range from `range(62, 82)` to `range(62, 83)` so the range includes 82 degrees, which is the max temp setting supported by the USA api prior to changing to HIGH. As it is currently, when the car reads HIGH, the api reads 81, which is incorrect. In reality even 82 is technically incorrect, as 82 comes before HIGH, but I think it's reasonable to cap the display to that in order to avoid mixing numbers and strings (if one were to try and actually display "HIGH").


cdnninja:
Thanks for this!

nprez83:
Thank you for all the hard work with this outstanding integration!

PR #480, opened 2024-01-15
feat(EU): New KIA models

Bubo08:


Bubo08:
Not everything is implemented, yet. Only the most important attributes from the vehicle status request are integrated.

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -74,6 +74,8 @@ class Vehicle:
     year: int = None
     VIN: str = None
     key: str = None
+    protocolType: int = None
```
fuatakgun, line 77: Can you follow current naming convention?

fuatakgun (commented):
Looks good to me, please allow me a couple of days to test locally as an older car owner

hyundai_kia_connect_api/Vehicle.py
Bubo08: Do you mean instead of Camel Case notation I should use underscore? Good point, I will fix it.

cdnninja: If protocol type isn't used in an if statement maybe we don't record it?

Bubo08: I fixed the naming and removed the attribute as proposed above. Additionaly, I continude with the mapping of the attributes.

cdnninja:
I can't figure out how to post the review from mobile app.  

Only thought is maybe renaming the ccs variable in vehicle to support multiple regions? I doubt other regions will call it that but still have versions.   Thoughts on making it more universal?

Bubo08:
> Only thought is maybe renaming the ccs variable in vehicle to support multiple regions? I doubt other regions will call it that but still have versions. Thoughts on making it more universal?

I think we can still refactor it, if it is the case. For me it looks a bit like a universal attribute but I might be wrong.

cdnninja:
Each region is totally different. Ev9s haven't had issues so far for other regions as an example.

Bubo08:
I did see in older logs that the attribut Ccuccs2protocolsupport is set to 0 and in my sniffed code it is set to 1. So my interpretation was, that ccs2 was disabled in the past and now - in some newer cars - it is enabled. I think other region might also have to switch sooner or later to the ccs2 protocol. Interestingly, this attribut is also used in the header of other type of calls (vehicles, register, ...).

But it is only an interpretation which might be completely wrong. The attribut could also mean Combined Charging System (CCS) but then why would it be included in all calls?

cdnninja:
Interesting. So far I haven't seen this value in any other regions.  I have seen a version flag though.   Not sure why each region is so different.  Stamps also don't exist in ca and USA. They do exist in Australia though.

fuatakgun:
Austria is following EU style, right? Similar to IN and CN.

cdnninja:
Yes.

Bubo08:
I head to add the attribute Ccuccs2protocolsupport to the header. The value is taken from the result of the vehicle call. It would be good if somebody could test it for the old cars in EU.

jwillemsen:
I have a 2024 Kona and just tried this PR using a small test script, I do see a lot more data now being returned, including a lot of driving statistics, great work!

Bubo08:
> I have a 2024 Kona and just tried this PR using a small test script, I do see a lot more data now being returned, including a lot of driving statistics, great work!

Thanks! I mapped some additional attributes and documented the still missing ones.

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -123,6 +123,7 @@ class KiaUvoApiEU(ApiImpl):
 
     def __init__(self, region: int, brand: int, language: str) -> None:
         # Users were complaining about the warning message.   Stating it is already english.  The below handles this but still throws warnings for non english items.
+        self.ccu_ccs2_protocol_support = None
```
cdnninja, line 126: This should be moved up a line.  As that comment is explaining why it looks for the en and coverts variants to EN.

cdnninja (approved):
Good with this being merged.  @fuatakgun  let us know when you test on your car.

hyundai_kia_connect_api/KiaUvoApiEU.py
Bubo08: Done. Thanks for the review.

Bubo08:
> I tried to map some more values. Now it shows 43 entitys for EV9.

All are integrated now. Thanks for your help.

ristoavila:
> > I tried to map some more values. Now it shows 43 entitys for EV9.
> 
> All are integrated now. Thanks for your help.

Some fixes for previous mapping. I also mapped is_locked to driver door lock. Would be nice to get sniffing of the open lock and setting of the limits.
[0001-Fix-value-states-for-added-attributes.patch](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/files/13999122/0001-Fix-value-states-for-added-attributes.patch)

Bubo08:
I added the proposed changes and also changed the is_locked. All door are checked and only if all four are closed then I set the attribute to True.

ristoavila:
> I added the proposed changes and also changed the is_locked. All door are checked and only if all four are closed then I set the attribute to True.

Btw there is Lock status that is inverse its different from open. Property is Cabin.Door.Row1.Driver.Lock its 1 when its open and 0 when its locked.

cdnninja:
Thank you for this! I am going to merge and release this.  Another PR can be done for the other work that is on the go. This will help test the current progress out in a wider audience.

jwillemsen:
With the pre release 2.19.0 I do see much more data, but `KONA AC Charging Limit` and `KONA DC Charging Limit` but show a slider with 80, but in the app I see 100, also the UI is strange
Screenshot_20240123_132912

jwillemsen:
I don't have FrontLeftWindow and FrontRightWindow in HA, only the door, for the back I do see the door and window status

jwillemsen:
See https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/483 for the window status

PR #484, opened 2024-01-24
fix(EU): New models, convert location_last_updated_at to datetime

ZuinigeRijder:
@cdnninja [hyundai_kia_connect_monitor](https://github.com/ZuinigeRijder/hyundai_kia_connect_monitor#introduction-hyundai_kia_connect_monitor) uses Property Vehicle.location_last_updated_at. A German Kona model 2024 user tried out the support of the new EU models (v3.13.0). 

Property Vehicle.location_last_updated_at was not given back a datetime for the EU new models (e.g. Kia EV9, Hyundai Kona 2024 model) but a dictionary. Converted the dictionary to a datetime, like this is done for the other EU models.
 
Vehicle._location_last_set_time is now also converted to a datetime for the EU new models.

Other improvements:
- all the supported languages list are two-letters, so no distinguishing of e.g. en-gb or en-us, so just use the first two letters to avoid errors
- fixed style warning: comparison == False

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -123,9 +123,10 @@ class KiaUvoApiEU(ApiImpl):
 
     def __init__(self, region: int, brand: int, language: str) -> None:
         self.ccu_ccs2_protocol_support = None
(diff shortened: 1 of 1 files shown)


cdnninja:
Thanks for this! I think new models are getting much closer.

Issue #488, opened 2024-01-29
Missing data for ccs2 EU


jwillemsen:
* Hyundai / Kia Connect version: 2.19.0
* Python version:
* Operating System:

### Description

After upgrading to 2.19.0 using a kona 2024 EV which uses the new ccs2 support the following fields are missing
- heating/cooling all 4 seats
- smart key battery warning
- defrost
- back window heater

Also AC/DC Charging Limit looks strange, in the bluelink app both are set to 100, but HA shows 80
![Screenshot_20240123_132912](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/assets/47353/69ce1725-b966-4fca-afcb-42f19ecd6868)

### What I Did

Upgrade to 2.19.0

jwillemsen:
ccs2 support was added as part of https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/480

ZuinigeRijder:
I think this info is available, but @Bubo08 did already a great job in providing most of the other information. I got a debug log of a German Kona 2024 with my[ tool debug.py](https://github.com/ZuinigeRijder/hyundai_kia_connect_monitor/blob/main/debug.py)  from [hyundai_kia_connect_monitor](https://github.com/ZuinigeRijder/hyundai_kia_connect_monitor).

heating/cooling all 4 seats
````
"Cabin": {
                    "Door": {
                        "Row2": {
                            "Right": {"Lock": 0, "Open": 0},
                            "Left": {"Lock": 0, "Open": 0},
                        },
                        "Row1": {
                            "Driver": {"Open": 0, "Lock": 0},
                            "Passenger": {"Lock": 0, "Open": 0},
                        },
                    },
                    "HVAC": {
                        "Row1": {
                            "Driver": {
                                "Temperature": {"Value": "OFF", "Unit": 0},
                                "Blower": {"SpeedLevel": 0},
                            }
                        },
                        "Vent": {
                            "AirCleaning": {"Indicator": 1, "SymbolColor": 0},
                            "FineDust": {"Level": 0},
                        },
                        "Temperature": {"RangeType": 1},
                    },
                    "RestMode": {"State": 0},
                    "Seat": {
                        "Row2": {
                            "Left": {"Climate": {"State": 2}},
                            "Right": {"Climate": {"State": 2}},
                        },
                        "Row1": {
                            "Passenger": {"Climate": {"State": 2}},
[...]

PR #489, opened 2024-01-30
feat(EU): smart key battery warning and seat heating new cars

Bubo08:
Bug Fix for Missing data for ccs2 EU #488 

I did also add three additional battery value to the vehicle information:
- SOH state of health in percentage
- Capacity as int value
- Remain (Capacity) as int value

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -577,13 +577,13 @@ class KiaUvoApiEU(ApiImpl):
                 )
             ],
         )
-        vehicle.ev_first_departure_enabled = bool(get_child_value(
-            state, "Green.Reservation.Departure.Schedule1.Enable"
-        ))
+        vehicle.ev_first_departure_enabled = bool(
+            get_child_value(state, "Green.Reservation.Departure.Schedule1.Enable")
+        )
 
-        vehicle.ev_second_departure_enabled = bool(get_child_value(
-            state, "Green.Reservation.Departure.Schedule2.Enable"
-        ))
+        vehicle.ev_second_departure_enabled = bool(
+            get_child_value(state, "Green.Reservation.Departure.Schedule2.Enable")
+        )
 
         # TODO: vehicle.ev_first_departure_days --> Green.Reservation.Departure.Schedule1.(Mon,Tue,Wed,Thu,Fri,Sat,Sun)
         # TODO: vehicle.ev_second_departure_days --> Green.Reservation.Departure.Schedule2.(Mon,Tue,Wed,Thu,Fri,Sat,Sun)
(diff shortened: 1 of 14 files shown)


hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -415,11 +422,34 @@ def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None
...
+        flSeatHeatState = get_child_value(state, "Cabin.Seat.Row1.Driver.Climate.State")
+        if flSeatHeatState in [0, 2]:
+            vehicle.front_left_seat_status = False
+        elif flSeatHeatState == 1:
+            vehicle.front_left_seat_status = True
+
+        frSeatHeatState = get_child_value(
+            state, "Cabin.Seat.Row1.Passenger.Climate.State"
+        )
+        if frSeatHeatState in [0, 2]:
```
cdnninja, line 435: This appears to be status like all other regions not is_on.  As such should be using the constants. 

https://github.com/Bubo08/hyundai_kia_connect_api/blob/7bf5c6444cd7c0049b065df98685d8280d7f31a6/hyundai_kia_connect_api/const.py#L38-L49

Other states most likely exist for cars as per the above.

cdnninja (changes requested):

hyundai_kia_connect_api/KiaUvoApiEU.py
Bubo08: Good point. I did not realize that there were status in the code. I fixed the above. 
I think I will have to check some other attributes.

cdnninja: Sounds good.  What I have found in the past is anything that's a number in the api could be more than true false.  We just don't know it yet. 

We hit that on seats.  For a long time it was 0 and 1. Until we started seeing other options for more advanced cars.  We can cross that bridge when we come to it though.

Bubo08: The code is now more readable. So this is good change in my opinion :-)

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -598,9 +613,13 @@ def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None
         vehicle.air_control_is_on = get_child_value(
             state, "Cabin.HVAC.Row1.Driver.Blower.SpeedLevel"
         )
-        # TODO: vehicle.smart_key_battery_warning_is_on = get_child_value(
-        # TODO:     state, "status.smartKeyBatteryWarning"
-        # TODO: )
+        smart_key_battery_warning_is_on = get_child_value(
```
ZuinigeRijder, line 616: This if-else can be better written as:
````
        smart_key_battery_warning_is_on = get_child_value(
            state, "Electronics.FOB.LowBattery"
        ) != 0
````

Bubo08: As python newbie I realized that I even can use the bool function :-)

PR #491, opened 2024-02-01
fix: EU distance unit were not correctly set

Bubo08:


Bubo08:
I think my pull request is not correct. Only the last commit should be in this pull request.

Bubo08:
old commits are in this pull request

PR #492, opened 2024-02-02
fix: Distance unit fixed

Bubo08:
Should fix the kia_uvo problem: Not valid unit for device class distance #812

### DIFF
--- a/hyundai_kia_connect_api/Vehicle.py
+++ b/hyundai_kia_connect_api/Vehicle.py
@@ -249,6 +249,10 @@ class Vehicle:
     def total_driving_range(self):
         return self._total_driving_range
 
+    @property
+    def total_driving_range_unit(self):
+        return self._total_driving_range_unit
+
     @total_driving_range.setter
     def total_driving_range(self, value):
         self._total_driving_range_value = value[0]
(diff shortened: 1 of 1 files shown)


cdnninja:
This should match the property approach of the rest of the file with a setter. On my phone so hard to link to an example but I can later.

cdnninja:
Here is an example of unit being used.

Values being stored here: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/2ca8b233c0c012d702f511702bb40991914f42c1/hyundai_kia_connect_api/Vehicle.py#L143-L145

and setter getter: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/2ca8b233c0c012d702f511702bb40991914f42c1/hyundai_kia_connect_api/Vehicle.py#L258-L266

Plus from the EU API this must be set correctly.

cdnninja (changes requested):
Marking as requested changes per above.

Bubo08:
I took the following code as template as the odometer is not throwing this kind of error:

    @property
    def odometer(self):
        return self._odometer

    @property
    def odometer_unit(self):
        return self._odometer_unit

    @odometer.setter
    def odometer(self, value):
        float_value = get_float(value[0])
        self._odometer_value = float_value
        self._odometer_unit = value[1]
        self._odometer = float_value

Therefore I thought the unit setter function is missing and I added them. Why do you think this is not correct?

cdnninja:
Oops you are correct. The rest was in before this PR.

PR #496, opened 2024-02-05
fix: support air_temperature with ccs2

jwillemsen:
Currently not tested, just looked at the json data as attached
[kona2024data.txt](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/files/14166280/kona2024data.txt)

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -406,7 +406,9 @@ class KiaUvoApiEU(ApiImpl):
 
         vehicle.engine_is_running = get_child_value(state, "DrivingReady")
 
-        # TODO: vehicle.air_temperature = get_child_value(state, "Cabin.HVAC.Driver.Temperature.Value")
+        vehicle.air_temperature = get_child_value(
+            state, "Cabin.HVAC.Row1.Driver.Temperature.Value"
+        )
 
         defrost_is_on = get_child_value(state, "Body.Windshield.Front.Defog.State")
         if defrost_is_on in [0, 2]:
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -406,20 +406,13 @@ class KiaUvoApiEU(ApiImpl):
 
         vehicle.engine_is_running = get_child_value(state, "DrivingReady")
 
-        if get_child_value(state, "Cabin.HVAC.Row1.Driver.Temperature.Value"):
-            tempIndex = get_hex_temp_into_index(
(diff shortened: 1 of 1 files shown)


cdnninja (changes requested):
Unit should be set as well.

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -406,7 +406,20 @@ def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None
 
         vehicle.engine_is_running = get_child_value(state, "DrivingReady")
 
-        # TODO: vehicle.air_temperature = get_child_value(state, "Cabin.HVAC.Driver.Temperature.Value")
+        if get_child_value(state, "Cabin.HVAC.Row1.Driver.Temperature.Value"):
+            tempIndex = get_hex_temp_into_index(
```
cdnninja, line 410: To confirm is your raw data in hex? I got the impression it wasn't. This is only needed for cars that don't show the decimal value.

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -406,7 +406,13 @@ def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None
 
         vehicle.engine_is_running = get_child_value(state, "DrivingReady")
 
-        # TODO: vehicle.air_temperature = get_child_value(state, "Cabin.HVAC.Driver.Temperature.Value")
+        air_temp = get_child_value(
+            state,
+            "Cabin.HVAC.Row1.Driver.Temperature.Value",
```
cdnninja, line 411: Looks like no unit set now.

cdnninja (changes requested):

hyundai_kia_connect_api/KiaUvoApiEU.py
jwillemsen: This is just the retrieval into a local variable, it is set on line 415

hyundai_kia_connect_api/KiaUvoApiEU.py
jwillemsen: You are correct, it is decimal, just found also that when the car is off the value is set to "OFF"

hyundai_kia_connect_api/KiaUvoApiEU.py
cdnninja: Line 415 looks to be adjusting for off.

hyundai_kia_connect_api/KiaUvoApiEU.py
cdnninja: If the value is off we will need to set to that a number as the data type isn't a string.

PR #498, opened 2024-02-07
fix(EU): sort daily stats for same behaviour between cc2 and non-cc2

ZuinigeRijder:
For non cc2 the behaviour was that the daily stats were sorted by the bluelink/Connect server:
- increasing yyyymmss
- decreasing hhmmss

For cc2 the order is undefined. Fixed by always sorting when the Bluelink/Connect server returns the daily stats.

This will solve the issue: [Daily driving stats not sorted on date #820](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/820)

Also a minor style change and flake line length suppression warnings.


Bubo08:
cool, thanks.

PR #499, opened 2024-02-10
feat: India region support

gowthamgowtham:
Works for India region. I haven't managed to add new tests. I am sure some cleanup will be necessary.


cdnninja:
Thanks for doing this! I'll run through feedback this weekend.   Are you able to cleanup merge conflicts and potentially rebase?  It shows a lot of changes that may not be real.  It will help with review. 


Version commit also needs to be backed out.  The system does that.

india.py
cdnninja, line 1: This should either:

1. Removed.
2. Moved to the test folders and we load your creds into a secret in the repo so these tests are run like EU.   No pressure though.

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -45,7 +57,15 @@ class DayTripInfo:
     """Day Trip Info"""
 
     yyyymmdd: str = None
-    summary: TripInfo = None
```
cdnninja, line 48: Be removing this it will break other regions.   I also think adding the below will break other regions.  Can we conform to existing?

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -11,16 +12,27 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+@dataclass
```
cdnninja, line 15: Location is already handled and used in other values.  No need for a new class.

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -1,5 +1,6 @@
 # pylint:disable=missing-class-docstring,missing-function-docstring,wildcard-import,unused-wildcard-import,invalid-name
 """Vehicle class"""
+import decimal
```
cdnninja, line 3: We use floats.

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -11,16 +12,27 @@
...
+class Location:
+    lat: decimal = None
+    lon: decimal = None
+
+
 @dataclass
 class TripInfo:
     """Trip Info"""
 
-    hhmmss: str = None  # will not be filled by summary
```
cdnninja, line 18: Removing fields will break other regions and not work for kia_uvo.   Please try to use the existing fields if possible.

hyundai_kia_connect_api/HyundaiApiIN.py
```diff
@@ -0,0 +1,968 @@
...
+            # if iterate the whole notifications list and
+            # can't find the action, raise an exception
+            raise APIError(f"No action found with ID {action_id}")
+
+    def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> None:
+        state = self._get_cached_vehicle_state(token, vehicle)
+        self._update_vehicle_properties(vehicle, state)
+        state = self._get_maintenance_alert(token, vehicle)
+        self._update_vehicle_maintenance_alert(vehicle, state)
+        state = self._get_location(token, vehicle)
```
cdnninja, line 761: This needs protective logic.  By doing this in a cached call you will kill the vehicle battery.  Cached calls shouldn't hit the car each call.

cdnninja (changes requested):

hyundai_kia_connect_api/KiaUvoApiAU.py
cdnninja, line 1: Could you remove these changes just to keep the PR clean?

cdnninja (changes requested):

hyundai_kia_connect_api/HyundaiApiIN.py
```diff
@@ -0,0 +1,968 @@
...
+                second=int(m.group(6)),
+                tzinfo=self.data_timezone,
+            )
+
+        _LOGGER.debug(f"{DOMAIN} - last_updated_at - after {value}")
+        return value
+
+    def force_refresh_vehicle_state(self, token: Token, vehicle: Vehicle) -> None:
+        # FIXME: Gowtham - No force refresh
+        _LOGGER.error(f"Disabling force refresh")
```
cdnninja, line 963: Add domain like other logging to clarify what is throwing this.    I assume this exists since force refresh command doesn't work yet?

cdnninja (changes requested):

CodeWithShreyans:
any updates on this?

nishant713:
Hi @gowthamgowtham any update?

ohuc:
any updates

nishant713:
@gowthamgowtham any update?

nishant713:
Hi @cdnninja @fuatakgun 
Using @gowthamgowtham 's branch and the india.py I am able to get information from my car using python on my local machine (not HA). 

Is it possible to be of some help remotely so as to resolve the issues and finish Indian bluelink integration?

I'm not much good at programming or gift, but can run python on my machine

cdnninja:
Is that when you first connect or on an action?

cdnninja:
If you also create a PR to this repo it will let me comment on specific parts of the code.

PR #528, opened 2024-03-21
feat: add support for ccs2 vehicles in Australia

bitnimble:
This PR:
 * Consolidates some datetime parsing code between all of the regions
 * Moves CCS2 parsing out of the EU implementation, and into Vehicle.py instead
 * Adds support for CCS2 vehicles in Australia

When this is ready to merge, please merge it with "Squash and merge" :)


hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -57,10 +57,6 @@ def refresh_vehicles(self, token: Token, vehicles: list[Vehicle]) -> None:
         Required for Kia USA as key is session specific"""
         return vehicles
 
-    def get_last_updated_at(self, value) -> dt.datetime:
```
bitnimble, line 60: Moved this into `utils` instead, as there was a duplicate implementation for every region

hyundai_kia_connect_api/KiaUvoApiAU.py
```diff
@@ -657,18 +664,6 @@ def _update_vehicle_drive_info(self, vehicle: Vehicle, state: dict) -> None:
         vehicle.power_consumption_30d = get_child_value(state, "consumption30d")
         vehicle.daily_stats = get_child_value(state, "dailyStats")
 
-    def _get_cached_vehicle_state(self, token: Token, vehicle: Vehicle) -> dict:
```
bitnimble, line 660: This was only used in `update_vehicle_with_cached_state` so I just inlined it instead to make switching over ccs2 vehicles easier

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -396,281 +377,10 @@ def force_refresh_vehicle_state(self, token: Token, vehicle: Vehicle) -> None:
             else:
                 self._update_vehicle_drive_info(vehicle, state)
 
-    def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None:
```
bitnimble, line 399: Moved into Vehicle.py - the CCS2 schema seems largely the same, if not identical between Aus and EU so I'm opting to just have a shared implementation of the parser instead.

cdnninja: It isn't used for the other regions and starts to break the structure.  I think we are better off creating a stamp based class that inherits for EU / AU.   So apiimpl -->  stampimpl --> kiaUvoAPIAU   this way you can override as required but most code would be in stampimpl.

cdnninja (changes requested):

hyundai_kia_connect_api/KiaUvoApiEU.py
bitnimble: This is for vehicle state parsing, not stamps 🤔 

Happy to pull out to a separate class anyway, though, if it differs from NA. ooi, do we know if NA ccOS vehicles use this schema as well? or still different?

cdnninja: Stamps is best way I can describe the api method.  Canada and USA are massively different api approach and parsing.

cdnninja: Open to other names for this general api approach.  Just can't think of another name to classify the eu / au approach Kia uses.

bitnimble: Ah gotcha. I'm aware that for existing vehicles, NA/CA have a different schema to EU/AU/CN, but I'm not sure if that's the case for the newer ccOS vehicles.
In EU/AU/CN, non-ccOS and ccOS vehicles _also_ have a different API, so I'm wondering if the new ccOS schema is just global now... but maybe that's just wishful thinking. It would be good to verify that though as it'd certainly make things more consistent on our side.

For now I'll go with what you've described and pull it out of vehicle 👍 other random names off the top of my head (none of them good either lol):
* Just call it EuAuCnVehicleParser.py vs NaCaVehicleParser.py, boring name but it's extremely obvious how they're different / where they're used 🤷 
* Just call it "type 1" or "type 2" or something like that, and have a comment on each one

Happy to go with whatever you prefer :)

cdnninja: Maybe we just go with type 1?   CA, USA Hyundai, USA Kia are all different so they would remain as is.   Tough to name.   Reason I called it stamps is that CA, USA Hyundai and USA kia don't use stamps at all for authentication where as all the others do.  I did a quick draft of what I meant for structure I will create a PR.  Just purely as example. I could also merge it to build upon if it helps. Happy to rename to whatever works as well.

bitnimble: sgtm! we can always rename it later if we come up with a better name as well :)

cdnninja: In #530  I have renamed to ApiImplType1.  If it looks okay I can merge it and this PR can be changed to align with it?   I noticed China doesn't use stamps but rest seems to match AU/EU.   I don't think china works anymore though. Something changed and the person who added it hasn't come back to fix it.

bitnimble: 👌 happy to go with that, will update this PR, cheers

cdnninja: I merged my PR and resolved the conflict here. I think last step would be moving the css2 code from vehicle back into the apiimpltype1 file.  Let me know once ready for next review.

zappoo: The above merge @cdnninja i think has caused a break - my HA is now showing this error:

`Setup failed for custom integration 'kia_uvo': Unable to import component: cannot import name 'stampApiImpl' from 'hyundai_kia_connect_api.ApiImplType1' (/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/ApiImplType1.py)`

cdnninja:
I took the liberty of copy pasting some code around.    @bitnimble  let me know if this works for you.  Someone will need to test it for AU.  The CI already ran a test build for EU successfully.  

Ignore the pre-commit failure. Something I am trying to move this repo over to but lots of things to fix to comply with ruff.

bitnimble:
Fixed up and working again now: 
image

cdnninja:
Just looking on my phone.  I think the call from eu api to parsing method will need to be tweaked to match.   I bet it would fail right now.

bitnimble:
Good catch! fixed that up. I'll also look at doing a separate follow up PR to start lifting the common stuff out of EU/AU/CN and into the Type1 impl.

PR #539, opened 2024-03-29
feat: add start/stop charge and charge limits to Hyundai USA

stephenjamieson:


stephenjamieson:
By the way, start/stop charge seems to require pin

cdnninja:
Thanks for doing this!!

PR #547, opened 2024-04-24
fix(EU): missing function

asychow:


cdnninja:
Thank you for this!  Looks like I made a mistake in merging.

PR #548, opened 2024-04-26
feat(EU): schedule charging and climate

asychow:


hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -729,6 +796,23 @@ def _update_vehicle_properties(self, vehicle: Vehicle, state: dict) -> None:
             ):
                 vehicle.ev_off_peak_charge_only_enabled = False
 
+        if (
```
cdnninja, line 799: Could this be added to constants to follow the current structure?  I think this will currently error if this value comes back as "None".

cdnninja (changes requested):
Amazing work! Thanks for adding this!

hyundai_kia_connect_api/KiaUvoApiEU.py
asychow: I suppose you are referring to `const.py`? There isn't a mapping for a boolean attribute so far. Happy to add that (perhaps as a generic mapping from `[None,0,1]` to `[None,False,True]`) anyway.

Regarding the issue when `None` is returned,

- `None == 0` returns `False`
- `None == 1` returns `False`

Therefore if `None` (or a `str`, a `dict` or a number not `[0,1]`) is returned from `get_child_value()`, `vehicle.ev_schedule_charge_enabled` remains `None`. And there should be no error?

In my opinion, the existing comparision here is redundant.
https://github.com/asychow/hyundai_kia_connect_api/blob/196dafdc09b21d534bb1d9c358bd7be876d87d72/hyundai_kia_connect_api/KiaUvoApiEU.py#L778C9-L781C11

asychow:
Thanks for reviewing the code. I have replied to your comment.

hyundai_kia_connect_api/KiaUvoApiEU.py
cdnninja: Why is existing redundant? I'm good with either approach however the code can't crash if None is returned for that value.

cdnninja: You are correct.  I thought comparing to None would fault but tested and it doesn't.

PR #567, opened 2024-06-04
fix: kia usa climate start

whinis:
I looked into what was being sent on the web portal for USA and noticed that remoteClimate only had temp and unit, I rearranged that on my version of home assistant and climate action started working again so I am submitting this pull request with the rearranged body.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoAPIUSA.py
+++ b/hyundai_kia_connect_api/KiaUvoAPIUSA.py
@@ -677,44 +677,44 @@ class KiaUvoAPIUSA(ApiImpl):
 
         body = {
             "remoteClimate": {
-                "airCtrl": options.climate,
                 "airTemp": {
                     "unit": 1,
                     "value": str(options.set_temp),
                 },
-                "defrost": options.defrost,
-                "heatingAccessory": {
-                    "rearWindow": int(options.heating),
-                    "sideMirror": int(options.heating),
-                    "steeringWheel": int(options.heating),
-                },
-                "ignitionOnDuration": {
-                    "unit": 4,
-                    "value": options.duration,
-                },
-                "heatVentSeat": {
-                    "driverSeat": {
-                        "heatVentType": front_left_heatVentType,
-                        "heatVentLevel": front_left_heatVentLevel,
-                        "heatVentStep": 1,
-                    },
-                    "passengerSeat": {
-                        "heatVentType": front_right_heatVentType,
-                        "heatVentLevel": front_right_heatVentLevel,
-                        "heatVentStep": 1,
-                    },
-                    "rearLeftSeat": {
-                        "heatVentType": rear_left_heatVentType,
-                        "heatVentLevel": rear_left_heatVentLevel,
-                        "heatVentStep": 1,
-                    },
-                    "rearRightSeat": {
-                        "heatVentType": rear_right_heatVentType,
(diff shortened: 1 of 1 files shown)


cdnninja:
Thank you for doing this!

chasebolt:
do you have heated or ventilated seats?

whinis:
I have both, I can test that

chasebolt:
here is a dump from the API call to my 2024 Kia Sportage (US). you will notice everything is nested under `remoteClimate`.

```json
{
   "action":"ACTION_EXEC_REMOTE_CLIMATE_ON",
   "remoteClimate":{
      "airTemp":{
         "value":"68",
         "unit":1
      },
      "airCtrl":true,
      "defrost":false,
      "ventilationWarning":false,
      "ignitionOnDuration":{
         "value":5,
         "unit":4
      },
      "heatingAccessory":{
         "steeringWheel":0,
         "steeringWheelStep":null,
         "sideMirror":0,
         "rearWindow":0
      },
      "heatVentSeat":{
         "driverSeat":{
            "heatVentType":2,
            "heatVentLevel":4,
            "heatVentStep":1
         },
         "passengerSeat":{
            "heatVentType":2,
            "heatVentLevel":4,
            "heatVentStep":1
         },
         "rearLeftSeat":{
            "heatVentType":0,
            "heatVentLevel":1,
            "heatVentStep":0
         },
         "rearRightSeat":{
            "heatVentType":0,
            "heatVentLevel":1,
            "heatVentStep":0
         }
      }
   }
}
```

chasebolt:
i find it strange that your API call has everything outside of `remoteClimate`. can you verify this is correct and output the json payload that is being sent?

cdnninja:
I wouldn't doubt it if different per car.  Other regions have some of this crazy complexity.  It means you need to figure out when is it different and why, attaching that to if statements.

whinis:
So I went a step further because I was getting 2 different results based on what node I was on in the browser. So I MITM the kia app using httptoolkit and got this from the app
```{
  "remoteClimate": {
    "airCtrl": true,
    "airTemp": {
      "unit": 1,
      "value": "72"
    },
    "defrost": false,
    "heatingAccessory": {
      "rearWindow": 0,
      "sideMirror": 0,
      "steeringWheel": 0
    }
  }
}```

Which seems similar to what already existed but currently returns a schema error for me when not via the app

whinis:
Also, regardless of where I put it I cannot get the climate to start. With it this way I get a success on the first call but 5 minutes later get a climate start failed message

whinis:
So after more testing it seems to be the `"heatVentSeat" ` key causes the error. I have heated and vented seats but the ev6 appears to have no ability to set them via API currently. Going to make a new PR with this new information and figure out a way to only add that key if the option is enabled

whinis:
I updated my branch if you would like to test @chasebolt, Since my car has them but cannot set them via API it seems to cause the issue. I hope this new version is a bit better

whinis:
Unable to test further as I am getting a `unable to get local issuer certificate` on home assistant and on my test environment. Not sure if its related to me MITM my app or not, its an entirely different device so shouldn't be affected but is.

whinis:
pushed #568 which should also fix the ssl issue, not sure why its an issue now

chasebolt:
sorry i have been MIA. i did try your original commit, where you moved the options outside of `remoteClimate` and it did break for me. i will try the latest merge change you did and see if that has it working again.

PR #568, opened 2024-06-08
fix: usa hyundai remote start fails

whinis:
fix: USA ssl connection by importing certs from certifi
Update KiaUvoAPIUSA.py

After doing more research, it was the heatedventedseat option that cause the issue. Added a if statement to only include the block if passed to start climate


cdnninja:
Thanks for doing this!

whinis:
Ya no problem, I very much enjoy having this in home assistant

PR #577, opened 2024-06-19
fix: KiaUVOApiCA check_action_status

rjantoine:
Adds synchronicity (run until complete) and returns OrderStatus instead of simply returning True.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -665,7 +665,7 @@ class KiaUvoApiCA(ApiImpl):
         elif response["result"]["transaction"]["apiResult"] == "C":
             return OrderStatus.SUCCESS
         elif response["result"]["transaction"]["apiResult"] == "P":
-            if synchronous == False:
+            if not synchronous:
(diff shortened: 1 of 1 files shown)


rjantoine:
I have tested this and it works as expected.

vm.check_action_status(vehicle_id, action_id, False) only runs once and returns either OrderStatus.FAILED, SUCCESS or PENDING.

vm.check_action_status(vehicle_id, action_id, True, timeout) runs every 10 seconds until return status is OrderStatus.FAILED, SUCCESS or timeout is reached and returns TIMEOUT.

cdnninja:
Awesome! I haven't looked at other regions.  I assume eu appears to have same behaviour?

rjantoine:
Although it’s implemented differently, yes, this code behaves the same way as the EU check_action_status and the same way as specified in coordinator.

cdnninja:
Thanks for confirming!

rjantoine:
The only region that doesn’t behave this way is the US. BluelinkApiUS doesn’t even have a check_action_status.

KiaUvoApiUSA has check_action_status but it simply returns True. It does not have a synchronous option implemented and does not return the OrderStatus constants.

PR #580, opened 2024-06-24
feat(EU): Support setting the charging current

jrnk:
Hi! I noticed this feature missing and I have a need for it (including others apparently: https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/864).

I sniffed the traffic on my EU app, so I don't know if other regions also have support for this, the above mentioned issue has screenshots in German so that is EU too :)

With this implemented we can set the charging current to 60%, 90% or 100%


cdnninja:
This is great! Thank you for doing this.

PR #589, opened 2024-07-08
fix(CA): remote start for gv60 

hurricane7719:
Code change to address "Canada - Unable to update location #906" in kia_uvo

Change in Vehicle location url is conditional to avoid possibly breaking the feature with other brands/models.


cdnninja:
This has your username and password in it.  

Could you simplify to smaller PRs? It will help make easier review.

hurricane7719:
Please reject this.

I was not intending to push this to the main repository.  Just testing some
things and still trying to understand git hub

cdnninja:
I can do that.  As a note that won't remove your password or pin from your repo which is public.

PR #597, opened 2024-07-22
Genesis Canada Location fix

hurricane7719:
Update to resolve issue #906 in kia_uvo

Updated URL for location request.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -1,6 +1,7 @@
 """KiaUvoApiCA.py"""
 
 # pylint:disable=unused-argument,missing-timeout,logging-fstring-interpolation,bare-except,invalid-name,missing-function-docstring
+#Master
 
 import time
 import datetime as dt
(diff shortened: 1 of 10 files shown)


hyundai_kia_connect_api/KiaUvoApiCA.py
```diff
@@ -422,11 +422,11 @@ def _update_vehicle_properties_service(self, vehicle: Vehicle, state: dict) -> N
     def _update_vehicle_properties_location(
         self, vehicle: Vehicle, state: dict
     ) -> None:
-        if get_child_value(state, "coord.lat"):
+        if get_child_value(state, "gpsDetail.coord.lat"):
```
cdnninja, line 425: Wouldn't this break other brands within Canada?

hurricane7719:
Not sure.  Be interesting to know if it's the same across Hyundai and Kia.
Could make it conditional on brand.  That would point to them diverging
across brands.

cdnninja:
How about we scream test? We can make the change and see if we get reports of issues.

cdnninja:
Please comment if still interested in merging. Closing for now.

Issue #617, opened 2024-08-24
HyundaiBlueLinkAPIUSA Vehicle.daily_stats NOT consistent with the other region implementations


ZuinigeRijder:
-   Hyundai / Kia Connect version: 3.22.7
-   Python version: 3.9.13
-   Operating System: Windows

### Description

Region Europe was the first one ([by me](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/248)) implementing Vehicle._daily_stats, Vehicle.month_trip_info and  Vehicle.day_trip_info.
Later on, also implementations followed for region Australia and China. Also HyundaiBlueLinkAPIUSA  has partly implemented Vehicle.daily_stats, however the implementation is NOT consistent with the other region implementations.

The Bluelink API is very different for HyundaiBlueLinkAPIUSA, see also[ this discussion in hyundai_kia_connect_monitor](https://github.com/ZuinigeRijder/hyundai_kia_connect_monitor/discussions/62). 

1. vehicle.daily_stats is not sorted, only sorted in KiaUvoApiEU.py, same problem for other regions.
2. In USA you have daily stats information for all trips (so no cumulative day value, but for each trip the actual values), in Europe (and the other regions?) only the cumulative latest value per day.
3. method update_day_trip_info is NOT overruled/implemented in HyundaiBlueLinkAPIUSA.py, although the information seems available for the last day?

### What I Did

1. made a local fix for issue 1, so the sorting is done in setters of Vehicle.daily_stats, Vehicle.month_trip_info and Vehicle.day_trip_info.
2. checked with [hyundai_kia_connect_monitor/debug.py](https://github.com/ZuinigeRijder/hyundai_kia_connect_monitor/blob/main/debug.py) what information is returned for KiaUvoApiEU.py and for HyundaiBlueLinkAPIUSA.py (latter provided by @gyveri  )

See next posts for the API for getting daily_stats, month_trip_info, Vehicle.day_trip_info from the debug.py output.

ZuinigeRijder:
## Relevant debug.py output for KiaUvoApiEU.py.


### get_driving_info responseAlltime 
````
{
    "periodTarget": 1
}
POST /api/v1/spa/vehicles/<id>/drvhistory

{
    'retCode': 'S',
     'resCode': '0000',
     'resMsg': 
    {
        'drivingInfo': 
        [
            
            {
                'drivingPeriod': 0,
[...]

PR #618, opened 2024-08-24
fix: sort daily_stats for all implementations in setter of Vehicle.daily_stats

ZuinigeRijder:
Only KiaUvoApiEU.py was sorting Vehicle.daily_stats, Vehicle.month_trip_info and Vehicle.day_trip_info, other implementations did not.

Moved the sorting to Vehicle in each setter.
Changed the Europe only comment.
Fixed or suppressed flake and lint warnings.

This partly solves (problem 1) this issue: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/617

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,16 +89,16 @@ For a list of language codes, see here: https://www.science.co.il/language/Codes
 - "pt" Portuguese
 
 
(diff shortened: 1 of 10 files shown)


cdnninja:
Thank you as always!  I no longer own one of these cars and greatly appreciate the work you put into this.

PR #620, opened 2024-08-30
fix: set daily_stats distance_unit

ZuinigeRijder:
See this issue: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/617#issuecomment-2315105745

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -1,4 +1,4 @@
-""" "ApiImplType1.py"""
+"""ApiImplType1.py"""
 
 import datetime as dt
 from typing import Optional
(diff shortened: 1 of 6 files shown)


cdnninja:
I am confused on the intent of this.  If not getting the unit from the API and just copying the unit we already retrieved doesn't this just increase memory consumption? If something downstream needs the unit from this service could you just use vehicle.odometer_unit?

ZuinigeRijder:
> I am confused on the intent of this. If not getting the unit from the API and just copying the unit we already retrieved doesn't this just increase memory consumption? If something downstream needs the unit from this service could you just use vehicle.odometer_unit?

Yes, as afterthought, this would have been better. But currently the memory consumption is already there, because it was filled always with "km", which is wrong for some regions. And the clients of the Vehicle.daily_stats might use the distance_unit and removing it is a breaking change in the API. At least hyundai_kia_connect_monitor is using this field......

cdnninja:
Ah yes I see the dataset already has this.   I'll merge. Not worth the effort to change.

PR #625, opened 2024-09-03
fix: Removed possible decimals from get_hex_temp_into_index

rjantoine:
The Canadian website (and surely others) sometimes returns a decimal for temperatures. Since the expected value is an integer, it crashes everything. See crash report below.

> Failed setup, will retry: Config Not Ready: Error communicating with API: Traceback (most recent call last): File "/config/custom_components/kia_uvo/coordinator.py", line 126, in _async_update_data await self.hass.async_add_executor_job( File "/usr/local/lib/python3.12/concurrent/futures/thread.py", line 58, in run result = self.fn(*self.args, **self.kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 115, in check_and_force_update_vehicles self.update_vehicle_with_cached_state(vehicle_id) File "/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 90, in update_vehicle_with_cached_state self.api.update_vehicle_with_cached_state(self.token, vehicle) File "/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/KiaUvoApiCA.py", line 180, in update_vehicle_with_cached_state self._update_vehicle_properties_base(vehicle, state) File "/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/KiaUvoApiCA.py", line 251, in _update_vehicle_properties_base tempIndex = get_hex_temp_into_index( ^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/utils.py", line 39, in get_hex_temp_into_index value = int(value, 16) ^^^^^^^^^^^^^^ ValueError: invalid literal for int() with base 16: '24.5' During handling of the above exception, another exception occurred: Traceback (most recent call last): File "/config/custom_components/kia_uvo/coordinator.py", line 135, in _async_update_data await self.hass.async_add_executor_job( File "/usr/local/lib/python3.12/concurrent/futures/thread.py", line 58, in run result = self.fn(*self.args, **self.kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 85, in update_all_vehicles_with_cached_state self.update_vehicle_with_cached_state(vehicle_id) File "/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 90, in update_vehicle_with_cached_state self.api.update_vehicle_with_cached_state(self.token, vehicle) File "/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/KiaUvoApiCA.py", line 180, in update_vehicle_with_cached_state self._update_vehicle_properties_base(vehicle, state) File "/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/KiaUvoApiCA.py", line 251, in _update_vehicle_properties_base tempIndex = get_hex_temp_into_index( ^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/local/lib/python3.12/site-packages/hyundai_kia_connect_api/utils.py", line 39, in get_hex_temp_into_index value = int(value, 16) ^^^^^^^^^^^^^^ ValueError: invalid literal for int() with base 16: '24.5'

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -247,7 +247,10 @@ class KiaUvoApiCA(ApiImpl):
         )
         # Converts temp to usable number. Currently only support celsius.
         # Future to do is check unit in case the care itself is set to F.
-        if get_child_value(state, "status.airTemp.value") != "OFF" and get_child_value(state, "status.airTemp.value")[-1] == "H":
+        if (
+            get_child_value(state, "status.airTemp.value") != "OFF"
+            and get_child_value(state, "status.airTemp.value")[-1] == "H"
+        ):
             tempIndex = get_hex_temp_into_index(
                 get_child_value(state, "status.airTemp.value")
             )
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -247,7 +247,7 @@ class KiaUvoApiCA(ApiImpl):
         )
         # Converts temp to usable number. Currently only support celsius.
         # Future to do is check unit in case the care itself is set to F.
-        if get_child_value(state, "status.airTemp.value") != "OFF":
+        if get_child_value(state, "status.airTemp.value") != "OFF" and get_child_value(state, "status.airTemp.value")[-1] == "H":
             tempIndex = get_hex_temp_into_index(
                 get_child_value(state, "status.airTemp.value")
             )
--- a/hyundai_kia_connect_api/utils.py
+++ b/hyundai_kia_connect_api/utils.py
@@ -35,7 +35,7 @@ def get_float(value):
 
 def get_hex_temp_into_index(value):
     if value is not None:
-        value = value.replace("H", "")
+        value = value.replace("H", "").split(".")[0]
         value = int(value, 16)
         return value
     else:
--- a/hyundai_kia_connect_api/utils.py
+++ b/hyundai_kia_connect_api/utils.py
@@ -35,7 +35,7 @@ def get_float(value):
 
 def get_hex_temp_into_index(value):
     if value is not None:
-        value = value.replace("H", "").split(".")[0]
+        value = value.replace("H", "")
         value = int(value, 16)
         return value
     else:


rjantoine:
@cdnninja Is this something that can be added to the API shortly? It's often causing everything for my vehicles in Home Assistant to become unavailable due to a crash when the API receives decimals from the Kia interface. Thanks 😃

cdnninja:
Thanks for the ping. I was meaning to run through some test cases on this to confirm it doesn't break hex values.

cdnninja:
I don't see how this code would work.  If you run this code and pass 24.5 to it the response will be 36.  It won't crash but won't give you the right value.   I would think this code should detect if hex, if it isn't pass it through.   If hex convert to the value.

cdnninja:
I would think swap the if statement to:

`    if value is not None and value[-1] == 'H':
`

As well as the return on the else to be return value.

rjantoine:
Turning 24 to 36 was the way the original code was intended to work, changing hex to int.

The issue is that newer vehicles will sometimes send 24.5, which converts to 36.3125 and crashes when calling ```int(24.5, 16)``` since ```int()``` crashes when the input isn't an integer.

To avoid this, the ```.split(`.`)[0]``` code I added just strips any decimals from the hex code before parsing it. I don't believe there is a need to check if it's hex or not, since this function is meant to convert hex to integer. The only issue is that the original code assumed all hex values were whole numbers, and newer vehicles return decimal values, crashing the function.

cdnninja:
Now that I comment I realized my proposed code won't work either for you. 

Looks like your car is returning degrees not a hex list value.    Older cars would return 0BH in the temp field as an example.  That code would convert to 11 after you strip the H.   That 11 is position 11 in a list of possible temps.  That temp list changes by year.    

It was really clunky but not our choice, rather the APIs.  It looks like they finally just return the real value for newer cars. 

Some of this is outlined here: https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/220

Here is the Canadian temp ranges that are compared to the list: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/07d765dfe4b5a5bd5860e06be7fd270c0e77e32c/hyundai_kia_connect_api/KiaUvoApiCA.py#L66-L68

I think the right fix is to add the if statement I mentioned to this block of code: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/07d765dfe4b5a5bd5860e06be7fd270c0e77e32c/hyundai_kia_connect_api/KiaUvoApiCA.py#L251-L263

So if the value isn't hex, rather a temp just place the temp into the temp field.

rjantoine:
I see what you mean. You are correct, my vehicle is returning straight decimal as opposed to hex. When I started my vehicle and set the temperature to 20.5C, I later got ```"airTemp": { "value": "20.5", "unit": 0 }``` . I'll adjust the code accordingly. Thanks.

cdnninja:
Looks like now the code wouldn't crash but it doesn't map the value so you won't have the sensor in ha.  An else statement would be needed that does that when it isn't hex.

rjantoine:
Isn't it mapped at line 301?

```
vehicle.air_temperature = (
    get_child_value(state, "status.airTemp.value"),
    TEMPERATURE_UNITS[0],
)
```

cdnninja:
I'll merge this.  As a note I suspect you will have remote start issues since it translates it to hex for that too.   

Thanks for the work on this!

rjantoine:
Thanks. I will test the remote start later, but I did snoop what was being sent to the server, and it is in HEX for the remote start request.

PR #642, opened 2024-10-04
fix: handle EU server time issue

cdnninja:
Sometimes EU reports old time stamps

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -49,7 +49,7 @@ class ApiImplType1(ApiImpl):
     def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None:
         if get_child_value(state, "Date"):
             # `Date` field is in UTC time
-            if vehicle.last_updated_at:
+            if not vehicle.last_updated_at:
                 vehicle.last_updated_at = parse_datetime(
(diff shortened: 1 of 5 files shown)


cdnninja:
@ZuinigeRijder  you are thinking move this logic to the vehicle setting?

ZuinigeRijder:
Yes, I can make a proposal (new) Pull request?

cdnninja:
Of course! This is as much yours as mine! I don't have one of these cars anymore so can't say I spend much time on it.  I just increased your access to admin on this repo as well.

hyundai_kia_connect_api/ApiImplType1.py
```diff
@@ -49,7 +49,7 @@ def _get_authenticated_headers(
     def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None:
         if get_child_value(state, "Date"):
             # `Date` field is in UTC time
-            if vehicle.last_updated_at:
+            if not vehicle.last_updated_at:
                 vehicle.last_updated_at = parse_datetime(
                     get_child_value(state, "Date"), dt.timezone.utc
```
ZuinigeRijder, line 54: I think this was wrong, someone else already reported that this was wrong for ccs2, it should be self.data_timezone

ZuinigeRijder (changes requested):

PR #645, opened 2024-10-05
fix(EU): update default of ScheduleChargingClimateRequestOptions::DepartureOptions::days to non-empty to avoid API error

asychow:
Server rejects schedule charge/climate request when day is empty


cdnninja:
Thank you for this!

PR #664, opened 2024-10-30
chore: Update python; Remove travis and tox; trim .gitignore

maruel:
- Bump minimum python to 3.10 (from 3.6). Test in python 3.12 as HA currently uses 3.12 and will bump to 3.13 soon. I kept 3.10 because it's the default on Ubuntu 22.04 and I've yet to upgrade.
- I didn't see travis still being used so removed .travis.yml.
- tox couldn't succeed locally. The login tests require accounts in both Europe and Canada, which nobody can achieve. Remove all references to tox and guide contributors to leverage GitHub PR checks instead.
- Clean up .gitignore. It was full of random stuff that was not related to the project. Include references to help the user set up their environment properly instead.


maruel:
I realize that tox is not really used and the workflow https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/master/.github/workflows/python-tests.yml simply call flake8 and pytest directly. Should I remove tox references instead?

cdnninja:
You most likely know this better than I do.  Feel free to remove it.  

As for min version isn't that the min version to run not the required version?  Does it not run on older?

maruel:
Home Assistant only supports one python version at a time and closely tracks the latest version. See https://github.com/home-assistant/architecture/blob/HEAD/adr/0020-minimum-supported-python-version.md

So there's little point to supporting really old versions. 3.10 should be "old enough" here.

I tried running tox locally and mostly failed, since it requires the username/password to live test the API. It also parses the stdlib file that get injected which messed flake8 up. Since tox cannot work in the current state, I'll update this PR to remove tox.ini.

maruel:
I tweaked the github actions workflow but it won't fully take effect until it's merged. It's kinda annoying to test from this perspective. :/

.github/workflows/python-tests.yml
```diff
@@ -11,10 +11,11 @@ on:
 jobs:
     build:
         runs-on: ubuntu-latest
+        name: "test py${{matrix.python-version}}"
         strategy:
             fail-fast: false
             matrix:
-                python-version: ["3.10"]
+                python-version: ["3.10", "3.12", "3.13"]
```
cdnninja, line 18: Will this cause three sets of the build tests to run? If so that worries me since this queries a real car and can max out api calls.

maruel: Thanks I didn't realize it could become a real problem. I left 3.12 since it's what matters at the moment and removed 3.10 (which was the previous value) and 3.13.

cdnninja:
Excellent.  Thanks for doing this!

PR #666, opened 2024-10-31
docs: Update fuatakgun to Hyundai-Kia-Connect

maruel:


cdnninja:
Good catch. Love you are digging into this.  I don't own a Hyundai or Kia anymore so don't spend much time on it.  fuatakgun  also has some other things on the go!

maruel:
You're welcome! I saw the project had been well set up but needed a little bit of love. There's more real issues I'd like to dig down later (I'm seeing exceptions in my HA logs) but wanted to pick the easy cleanup things first.

cdnninja:
Amazing!  That would be great! I don't even see those exceptions anymore since I don't run the code myself.

PR #668, opened 2024-11-03
fix: made heating parameter binary in KiaUvoAPIUSA

renatkh:


cdnninja:
At one time this was correct.   This may break older cars or other models.  Let's merge and see what happens!

PR #672, opened 2024-11-10
feat(EU): support valet mode

asychow:
prerequisite of https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/966


emtunc:
I've tested against my Hyundai IONIQ 5 and it works 👍

PR #674, opened 2024-11-11
feat(USA hyundai): Add EV charge rate support

ADeeds:
Bluelink in America started supporting the EV charge rate in kW a couple weeks ago.

The JSON exposes two fields named batteryStndChrgPower and batteryDischrgPower whose values appear to be equivalent. I've used StndChrgPower to populate `vehicle.ev_charging_current`.

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
@@ -386,6 +386,9 @@ class HyundaiBlueLinkAPIUSA(ApiImpl):
         vehicle.ev_battery_is_plugged_in = get_child_value(
             state, "vehicleStatus.evStatus.batteryPlugin"
         )
+        vehicle.ev_charging_current = get_child_value(
+            state, "vehicleStatus.evStatus.batteryStndChrgPower"
+        )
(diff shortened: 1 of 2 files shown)


RuneNyhuus:
EU version have the same data.

hyundai_kia_connect_api/HyundaiBlueLinkAPIUSA.py
```diff
@@ -386,6 +386,9 @@ def _update_vehicle_properties(self, vehicle: Vehicle, state: dict) -> None:
         vehicle.ev_battery_is_plugged_in = get_child_value(
             state, "vehicleStatus.evStatus.batteryPlugin"
         )
+        vehicle.ev_charging_current = get_child_value(
+            state, "vehicleStatus.evStatus.batteryStndChrgPower"
```
asychow, line 390: shouldn't this be vehicleStatus.evStatus.batteryPower.batteryStndChrgPower?
according to https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/961

cdnninja: Good catch. Worth confirming what region that screenshot is from.

Issue #564, opened 2024-05-27
"No action found with ID..." when calling check_action_status (Kia/EU)


sjonez:
I'm successfully sending actions to lock/unlock/etc my car which returns an Action ID, but passing this ID to **check_action_status** always throws "No action found with ID..."

After looking at the code for check_action_status I manually tried navigating to **https://prd.eu-ccapi.kia.com:8080/api/v1/spa/notifications/[vehicleid]/records** which just returned:

{"retCode":"S","resCode":"0000","resMsg":[],"msgId":"4a94f720-f648-4303-83f9-f399233c3b9a"}

I assume the empty "resMsg" is the problem - any idea why I can't retrieve any action status? The actions are completing successfully, but I don't seem to have any way of tracking them.

I'm testing on a Kia EV6 in the UK

Thanks!

Sab44:
Same here, action completes successfully but getting this error in the logs.

Strixx76:
Same here when sending "start_climate". The action works just fine, and the climate and defrosting) starts, but I can see this exceptions in the logs.

Bob33700:
Hi, I'm testing with a E-NIRO (2019) and 3.32.0
In every test case the action (lock) succeed.

with original 3.32.0: **fail**
- **check_action_status** returns an empty list of actions => error

if I disable `token.device_id = self._get_device_id(self._get_stamp())` in **lock_action**: **a little better**
- **check_action_status** returns 1st:
`{'retCode': 'S', 'resCode': '0000', 'resMsg': [{'deviceId': 'fb400962-...-0aea80bb7a42', 'vehicleId': 'cfa84308-...-0d4dc534901d', 'recordId': 'b426c960-...-7b5dddf47bf2', **'confirmedYn': False**, 'action': 'uvo://control/door/close', **'result': None**, **'record': None**, 'createdAt': '2025-02-16 11:32:29.122', 'updatedAt': '2025-02-16 11:32:29.122'}], 'msgId': 'b45c0a30-...-8c6551e5350f'}`
then: 
[...]

PR #679, opened 2024-11-12
fix(EU): Remove token device_id refresh for some actions

asychow:
Potential fix for https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/564

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -913,7 +913,6 @@ class KiaUvoApiEU(ApiImplType1):
         ).json()
         _LOGGER.debug(f"{DOMAIN} - Lock Action Response: {response}")
         _check_response_for_errors(response)
-        token.device_id = self._get_device_id(self._get_stamp())
         return response["msgId"]
 
     def charge_port_action(
@@ -928,7 +927,6 @@ class KiaUvoApiEU(ApiImplType1):
         ).json()
         _LOGGER.debug(f"{DOMAIN} - Charge Port Action Response: {response}")
         _check_response_for_errors(response)
-        token.device_id = self._get_device_id(self._get_stamp())
         return response["msgId"]
 
     def start_climate(
@@ -1433,7 +1431,6 @@ class KiaUvoApiEU(ApiImplType1):
         ).json()
         _LOGGER.debug(f"{DOMAIN} - Schedule Charging and Climate Response: {response}")
         _check_response_for_errors(response)
-        token.device_id = self._get_device_id(self._get_stamp())
         return response["msgId"]
 
     def valet_mode_action(


cdnninja:
Those were added as a work around.   Our device id was being invalidated for some reason. Maybe it needs to be moved in the code though.  

Would need some decent testing before merging.

asychow:
I have performed two preliminary tests.

1. Capture 20 hours of BL traffic, with requests sent, wifi reconnection and app force-close & reopen multiple times during the period. Neither deviceId nor vehicleId is changed
2. Simulate BL requests using the API with breaks in between, e.g. lock => wait for success => sleep 10 seconds => wait for success => unlock => wait for success => sleep 3 minutes => close charge door => wait for success. All the commands ran successfully.

Please correct me if I am wrong or the tests are flawed, but the test results suggest that refreshing deviceId is no longer needed (maybe some changes on the server?). I'm happy to test this in the HA integration for an extended period, but ideally we need someone who experienced the deviceId issue in https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/672 to test too.

fuatakgun:
thanks for comprehensive test, vehicle id is not supposed to change, but for device id, can you please run it further more than 20 hours? let's see if it still requires to be updated after initial token expiration.

cdnninja:
This speaks to why this was added: https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/672

It is believed to be caused by us not registering push notifications correctly.  So when we run a command that triggers a push it errors server side and invalidates our device_id.   If you are finding that isn't the case when using these methods and using mobile app same day we can try remove this and roll it out.  Worse case we revert it.

vinnyspb:
I've done a few tests on Kia in EU and device ID changes after some time for me (not very long, minutes).
My impression is that Kia's API is not meant to support a long-living connection. If I login and then execute `start_climate()` without device_id refresh, everything works sound.
If I have a device_id refresh as part of `start_climate()` (as in the current master), I get the same behavior as @asychow mentioned - task is not being listed in the subsequent `check_action()` requests.
Would it make sense to re-login every time a 'write' operation (like `start_climate`) is supposed to be executed?

cdnninja:
Maybe we look at getting a new device id at the start of the call instead of end?  My impression is the token is still valid so not worth doing a full login.

cdnninja:
Maybe we look at getting a new device id at the start of the call instead of end?  My impression is the token is still valid so not worth doing a full login.

vinnyspb:
I tried it now, it looks like the device_id changes mid-action.

I moved the `_get_device_id()` call to the beginning of `stop_climate()` function.

Device ID is updated before the climate call, as expected.

Climate call returns a message ID.

Then several recursive calls to check the action status return pending status like `{'retCode': 'S', 'resCode': '0000', 'resMsg': [{'deviceId': '<...>', 'vehicleId': '<...>', 'recordId': '<...>', 'confirmedYn': False, 'action': 'uvo://control/temperature/stop', 'result': None, 'record': None, 'createdAt': '2024-12-28 17:58:10.063', 'updatedAt': '2024-12-28 17:58:10.063'}], 'msgId': '<...>'}`

However after 4th call to check the action status the response changes to this: `{'retCode': 'F', 'resCode': '4002', 'resMsg': 'Invalid request body - Invalid deviceId. ', 'msgId': '<...>'}`

Not sure how to deal with it tbh.

cdnninja:
Previous theory was that our push notification registration wasn't correct.  

So when it completes it sends a push notification to the device id.  It recognizes it isn't valid and invalidates it. 

Since we didn't have someone to review that that's the case it was done the way you see today.   It isn't the right way but got us by. 


Maybe sniff the push registration to see how we compare?

PR #680, opened 2024-11-12
feat: implement script 'bluelink'

maruel:
The tool `bluelink` dumps information live from the API and optionally stores it locally to JSON. This makes iterating on features much easier.

I had to work around Vehicle class idiosyncrasies. Not changing this class as it would be a significant breaking change.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,9 +6,9 @@ I no longer have a Kia or Hyundai so don't maintain this like I used to.  Others
 Introduction
 ============
 
-This is a Kia UVO, Hyundai Bluelink, Genesis Connect(Canada Only) written in python.  It is primary consumed by home assistant.  If you are looking for a home assistant Kia / Hyundai implementation please look here: https://github.com/Hyundai-Kia-Connect/kia_uvo.  Much of this base code came from reading bluelinky (https://github.com/Hacksore/bluelinky) and contributions to the kia_uvo home assistant project.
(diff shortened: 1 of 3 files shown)


maruel:
This is to help with https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/710#issuecomment-2470834826. Sometimes the tool would output:
```
Climate                    
  air_temperature: OFF °C
```
and I believe it's what causes the problem in Home Assistant.

cdnninja:
Yes climate sometimes returns a word vs a number and needs code to handle it.  

Vehicle.data has the raw json.  Wouldn't recreating this manually present errors? If helpful for debug I'm open to it though.  I always found debug logs and vehicle.data were enough.

maruel:
The Home Assistant integration uses the Vehicule object, not the raw data. That's why I wanted this printed out. I was thinking of only printing the raw data with --verbose but --verbose prints a lot of output. I really don't have a strong stance, whatever you feel is the most appropriate.

cdnninja:
Home assistant prints raw to the data entity attribute. It also prints raw api response when debug is turned on.  

I'm just struggling to understand the problem to solve?

maruel:
I am looking at refactoring ApiImpl to reduce the code duplication across all the implementations but that requires faster iteration. Testing on my local HA is quite disruptive for me because I do use my home assistant installation to control my home. I wanted to add action commands later, that's why I the current implementation is done as a subcommand. I don't know if it would be useful for others not using HA but that's there.

PR #681, opened 2024-11-13
fix: clarify ApiImpl which methods are required or not

maruel:
Self-document better.

Make update_geocoded_location() a staticmethod since it doesn't depend on self.

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -73,11 +73,11 @@ class ApiImpl:
 
     def login(self, username: str, password: str) -> Token:
         """Login into cloud endpoints and return Token"""
-        pass
+        raise NotImplementedError("required")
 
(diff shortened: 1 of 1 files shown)


hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -139,50 +144,59 @@ def lock_action(
...
         """Stops charge. Returns the tracking ID"""
-        pass
+        raise NotImplementedError("required")
 
     def set_charge_limits(
         self, token: Token, vehicle: Vehicle, ac: int, dc: int
     ) -> str:
         """Sets charge limits. Returns the tracking ID"""
-        pass
+        raise NotImplementedError("required")
```
cdnninja, line 171: Shouldn't this match set_charging_current?  Same with the other items above.

cdnninja (changes requested):

cdnninja:
Closing due to lack of response. Feel free to reopen if you would like.

PR #699, opened 2024-12-23
fix: seat heat/cool settings for USA vehicles

joeshaw:
This sets the `heatVentStep` to the appropriate value.  It also does
away with the math on `heatVentLevel`, which I found confusing.  It
now treats the options value more like an enum.

The values are based on the empirical data from
Hyundai-Kia-Connect/kia_uvo#718 and confirmed on my own vehicle.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiUSA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiUSA.py
@@ -620,6 +620,59 @@ class KiaUvoApiUSA(ApiImpl):
 
         return response.headers["Xid"]
 
+    def _seat_settings(self, level) -> dict:
+        # See const.SEAT_STATUS for the list and descriptions of levels.
+        #
+        # The values were determined empirically, see https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/718
+        if level == 8:    # High heat
+            return {
+                "heatVentType": 1,
+                "heatVentLevel": 4,
+                "heatVentStep": 1,
+            }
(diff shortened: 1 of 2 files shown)


hyundai_kia_connect_api/KiaUvoApiUSA.py
```diff
@@ -620,6 +620,59 @@ def lock_action(self, token: Token, vehicle: Vehicle, action) -> str:
...
+                "heatVentType": 2,
+                "heatVentLevel": 2,
+                "heatVentStep": 3,
+            }
+        elif level == 1:  # Generically on, let's assume high heat
+            return {
+                "heatVentType": 1,
+                "heatVentLevel": 4,
+                "heatVentStep": 1,
+            }
```
joeshaw, line 668: This is the only thing that is really different from before.  If this value was set to 1 before, the seat would turn off.  But `const.SEAT_STATUS` says that it should be "On", so this just sets it to the highest setting.

cdnninja: As I recall some cars can't set level.  Purely on off.  So this may break cars that don't support levels via the app.   I'm okay to merge this and do a scream test though.   It could have changed since that time.

joeshaw: The previous iteration of the code didn't handle this value at all, so I think we'll be no worse off, and maybe it'll even work!  If you'd like I can add a comment that this is untested, though.

PR #701, opened 2024-12-24
fix: add steeringWheelStep: 2

joeshaw:
This is needed for vehicles with multiple heating levels on the
steering wheel.  This doesn't allow for customization of the level
(it's always on high), but it's better than not being set.

Tested to work on a 2024 Kia EV9, but I am unsure if it'll cause
problems on vehicles without multiple levels.


cdnninja:
I see this is draft.  Is it waiting on something?

joeshaw:
Nope, I'm willing to try it if you are!

PR #705, opened 2025-01-09
fix: EU Seats

cdnninja:


hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -939,35 +939,79 @@ def charge_port_action(
...
+            )
 
-        hex_set_temp = get_index_into_hex_temp(
-            self.temperature_range.index(options.set_temp)
-        )
+            payload = {
+                "drvSeatLoc": "R",
+                "tempUnit": "C",
+                "seatClimateInfo": {
+                    "rrSeatClimateState": 6,
```
cdnninja, line 1001: Right now these are hard coded to confirm this code even runs.  Once that works we will change this.

PR #708, opened 2025-01-12
feat: Adds steeringWheel option separate from options.heating

erikn13:
relates to PR: https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/1010

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -34,6 +34,7 @@ class ClimateRequestOptions:
     front_right_seat: int = None
     rear_left_seat: int = None
     rear_right_seat: int = None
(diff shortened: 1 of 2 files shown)


cdnninja:
This would be a breaking change but can't remember how to flag that.   I think that's okay based on who uses it.

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -34,6 +34,7 @@ class ClimateRequestOptions:
     front_right_seat: int = None
     rear_left_seat: int = None
     rear_right_seat: int = None
+    steeringWheel: int = None
```
cdnninja, line 37: Please follow naming convention of other parameters.

cdnninja (changes requested):

PR #727, opened 2025-02-21
fix: osm geocoded location 403 error

robinlee09201:
403 error due to missing user-agent when requesting geocode from osm api. 
Add curl agent the the problem is solved.


cdnninja:
Thank you for this!

PR #733, opened 2025-02-24
fix: not working force vehicle state update

bkakilli:
Current logic needs `vehicle.last_updated_at` value not to be `None`. However, it stays `None` until we call `self.update_vehicle_with_cached_state(vehicle_id)` . Therefore it never actually force updates the vehicle state and always falls back to `update_vehicle_with_cached_state` call.

This PR fixes the issue by calling the `update_vehicle_with_cached_state` before checking the `last_updated_at` time so that the rest of the logic works.

### DIFF
--- a/hyundai_kia_connect_api/VehicleManager.py
+++ b/hyundai_kia_connect_api/VehicleManager.py
@@ -107,6 +107,7 @@ class VehicleManager:
         # Otherwise runs a cached update.
         started_at_utc: dt = dt.datetime.now(pytz.utc)
         vehicle = self.get_vehicle(vehicle_id)
+        self.update_vehicle_with_cached_state(vehicle_id)
         if vehicle.last_updated_at is not None:
             _LOGGER.debug(
                 f"{DOMAIN} - Time differential in seconds: {(started_at_utc - vehicle.last_updated_at).total_seconds()}"  # noqa
@@ -115,10 +116,6 @@ class VehicleManager:
                 started_at_utc - vehicle.last_updated_at
             ).total_seconds() > force_refresh_interval:
                 self.force_refresh_vehicle_state(vehicle_id)
-            else:
-                self.update_vehicle_with_cached_state(vehicle_id)
-        else:
-            self.update_vehicle_with_cached_state(vehicle_id)
 
     def force_refresh_all_vehicles_states(self) -> None:
         for vehicle_id in self.vehicles.keys():


cdnninja:
This would cause an additional API call for each check on force update which will trigger rate limits sooner.   If you want to just call a force update use: `force_refresh_vehicle_state`. 

I assume you are using this outside of home assistant?

cdnninja:
Why would it never call that?  After first scan interval it will be set via cached. 

Only time it should be none is on first call, which happens on boot up of HA.  

I'm trying to understand if I am missing something on the logic.     If last update isn't set within a regions cached update that would be a bug though.

hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -107,6 +107,7 @@ def check_and_force_update_vehicle(
         # Otherwise runs a cached update.
         started_at_utc: dt = dt.datetime.now(pytz.utc)
         vehicle = self.get_vehicle(vehicle_id)
+        self.update_vehicle_with_cached_state(vehicle_id)
         if vehicle.last_updated_at is not None:
```
bkakilli, line 111: @cdnninja how can we make sure this value is set without making another API call?

bkakilli:
> Only time it should be none is on first call, which happens on boot up of HA.

I see. Is it being logged in HA by default? I can double check whether it is calling the force update or not.

cdnninja:
Yes,  two aspects for this one.  If you enable debug on the integration it will automatically put this library into debug mode.   So anything that says logger.debug will be placed into the home assistant logs.  

You should see this comparison thrown into logs: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/b927c3e9646bc2ac703ca0f82c697e395e7219eb/hyundai_kia_connect_api/VehicleManager.py#L111-L113

Followed by the library saying what it is doing within the regional files.  

What region and car type do you have?  It is possible time zone comparisons aren't working well for force updates.

The other item is the sensor of last updated is exactly that value - if it populated you know it isn't none.  To confirm time zones are correct trigger a force update,  that sensor should show near your current time right after that assuming success.

bkakilli:
I'm in the US region, the vehicles are Hyundai Palisade and Hyundai Kona EV.

Datetime now returns: 2025-02-24 03:25:30.934450+00:00
`vehicle.last_updated_at`: 2025-02-24 03:24:45+00:00

So there does not seem to be a timezone issue and the comparison should be fine.

I will check HA debug logs and follow up.

bkakilli:
Verified on HA that it works as expected. Closing the PR as it only affecting the first call to the API as @cdnninja mentioned.

PR #743, opened 2025-03-08
feat: google as geocode option

cdnninja:
Add google as a geocode option.  Using a new library as well that looks to support many options!


cdnninja:
@gyveri could you test this out? I don't have one of these cars anymore so can't test it.

cdnninja:
@ZuinigeRijder FYI.

Issue #739, opened 2025-03-03
AU - No Odometer reading for Niro


SNilon:
When requesting the latest status for my 2024 Niro I don't get back a value for the odometer. This is strange because in the KIA Connect app the odometer value is displayed. 

Looking at the vehicle status response there is no odometer value:

> DEBUG:hyundai_kia_connect_api.KiaUvoApiAU:hyundai_kia_connect_api - get_cached_vehicle_status response: 
{'retCode': 'S', 'resCode': '0000', 
	'resMsg': {
		'airCtrlOn': False, 
		'engine': False, 
		'doorLock': True, 
		'doorOpen': {'frontLeft': 0, 'frontRight': 0, 'backLeft': 0, 'backRight': 0}, 
		'trunkOpen': False, 'airTemp': {'value': '02H', 'unit': 0, 'hvacTempType': 1}, 
		'defrost': False, 
		'lowFuelLight': False, 
		'acc': False, 
[...]

PR #745, opened 2025-03-09
Added work around to get Odometer value for AU Kia Niro

SNilon:
Recently I raised an issue that my AU region Niro didn't have an odometer value. This was due to it being missing in the status response from Kia. The Niro normally requests the non-CCS2 status, however if the CCS2 status is request the response will contain a valid odometer value.

I have implemented a workaround so that the CCS2 status will be request and the odometer value extracted if the non-CCS2 status doesn't contain a valid odometer value.

I have tested this on my car but it would be good if someone else in the AU region were to test it.


cdnninja:
In theory this increases calls to the API which could lead to rate limits.  I am good to merge and scream test though.  @fuatakgun  thoughts?

fuatakgun:
thanks for great work, given that this is a cached called, I do not have a problem with rate limits, while I am sure that I am wrong :) There are some users out, reporting that cached calls are even consuming 12v battery.

furthermore, we should somehow eliminate the repeating behavior of the code.

Can you dump your regular non ccs2 and ccs2 response?

SNilon:
There is an example of each response attached to issue #739, let me know if you want more.

The extra API call only happens if the non-CCS2 status is missing the odo value, so this should only result in more API calls for vehicles that are affected by this issue. More testing with other vehicles is probably necessary to confirm this tho.

PR #749, opened 2025-03-20
fix(CA): kia start climate

cdnninja:
For: https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1055

Extends the logic of this as kia has the same issue: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/703


cdnninja:
#[Shereef](https://github.com/Shereef) could you confirm this works for your car still?

cdnninja:
@JB12134 not sure.   Please test and see if it works.

Issue #687, opened 2024-11-23
start climate doesnt work with Canadian Kia EV9


sethp45:
-   Hyundai / Kia Connect version: 3.29.1
-   Python version: 3.12.4
-   Operating System: homeassistantOS

### Description
Calling start_climate with a Canadian Kia EV9 fails and returns error code 15109.

The API builds the payload like this for EV cars:

`
payload = {
    "hvacInfo": {
        "airCtrl": int(options.climate),
        "defrost": options.defrost,
        "heating1": options.heating,
        "airTemp": {
            "value": hex_set_temp,
            "unit": 0,
            "hvacTempType": 1,
        },
    },
    "pin": token.pin,
}
`

I dont know for other EV vehicules but at least for the Canadian EV9, Ive found out that it should be

`
payload = {
    "remoteControl": {
        "airCtrl": int(options.climate),
        "defrost": options.defrost,
        "heating1": options.heating,
        "igniOnDuration": options.duration,
        "ims": 0,
        "airTemp": {"value": hex_set_temp, "unit": 0, "hvacTempType": 0},
        "seatHeaterVentCMD": {
            "drvSeatOptCmd": options.front_left_seat,
            "astSeatOptCmd": options.front_right_seat,
            "rlSeatOptCmd": options.rear_left_seat,
            "rrSeatOptCmd": options.rear_right_seat,
        },
    },
    "pin": token.pin,
}
`

This is how Ive fixed it locally.
Hope this help

cdnninja:
Could you create a pull request?

sethp45:
I used to be a dev in my old days, that's how I managed to find how to a fix this but I'm not familiar with GitHub and pull request mecanism and I dont want to make a mess. So Id rather share the fix and let someone else make the change

cdnninja:
Where do we stand on this one? Is it still an issue?

sethp45:
This has been fixed

PR #750, opened 2025-03-22
fix(CA): Trying to revert change to climate command for non EV9

JB12134:
-EV9 KIA.

https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/749

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -596,18 +596,32 @@ class KiaUvoApiCA(ApiImpl):
                 },
             }
             if BRANDS[self.brand] == BRAND_KIA:
-                payload["remoteControl"] = climate_settings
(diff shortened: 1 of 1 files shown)


JB12134 (commented):
Trying to fix the Canada non-EV9 issues. Left the EV9 command the same, but if the Kia in Canada is not an EV9, it uses the old format. I'm new to git and "properly" doing dev work, be gentle on me.

cdnninja:
Do we have evidence the EV9 uses a different format?

JB12134:
I think so? I think this issue shows that.

https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/687

 I thought I saw it somewhere else, but I can't find it now. I thought that's what caused the change for KIAs in Canada.

cdnninja:
Good find! I will merge.

cdnninja:
Thank you for this!

PR #769, opened 2025-04-13
fix: obey timezones for CCS cars

cdnninja:
This fixes a key bug that would impact Australia being a +10 offset.  I would like eyes on this.  Only touches ccs cars since I don't know if offset exists for non ccs cars or if they still live in UTC for data.   If someone has a non ccs car would be great to check the data in the API for "offset".


cdnninja:
This fixes a key bug that would impact Australia being a +10 offset.  I would like eyes on this.  Only touches ccs cars since I don't know if offset exists for non ccs cars or if they still live in UTC for data.   If someone has a non ccs car would be great to check the data in the API for "offset".

cdnninja:
@fuatakgun  could you check if you non CCS car has "offset" in the data now? If so I should do this same change for the non CCS code.

cdnninja:
Some merges came from another branch.  Going to reopen as new.

fuatakgun:
Nope, doesn't exist for me

PR #770, opened 2025-04-13
fix: obey ccs timezones

cdnninja:
This fixes a key bug that would impact Australia being a +10 offset. I would like eyes on this. Only touches ccs cars since I don't know if offset exists for non ccs cars or if they still live in UTC for data. If someone has a non ccs car would be great to check the data in the API for "offset".

@fuatakgun  could you check if your CCS car now reports offset?


fuatakgun:
Just checked it, doesn't exist for me.

cdnninja:
Perfect thanks for checking.  I will merge based on that.

PR #781, opened 2025-04-17
fix: V2l for CCS vehicles

rutj87:
Updated V2L location to get state


cdnninja:
Awesome work!  For reference the title of the PR is the release notes and sets the version number.  So fix is x.x.1 release feat being feature so x.2.x.

For this change shouldn't electric be in it? Like you mentioned:  ['Green']['Electric']['SmartGrid']['VehicleToLoad']['DischargeLimitation']['SoC']


This link is also good for taking the log output and making it readable: https://paji-toolset.net/en/py-dict-to-json-converter

cdnninja:
Looks good to me.   Thanks for doing this.  Looking at the list around those items in the file I bet others aren't correct either.  Part of teaching you this is that I don't have much time to support anymore and the more people that know how the better!

PR #787, opened 2025-04-20
feat: Add climate support for Indian Region 

nishant713:
Support climate start/stop service for Indian Region

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -30,8 +30,6 @@ from .const import (
     TEMPERATURE_UNITS,
     VEHICLE_LOCK_ACTION,
     ORDER_STATUS,
-    REGIONS,
-    REGION_INDIA,
 )
 
 from .exceptions import (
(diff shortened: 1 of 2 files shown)


cdnninja:
Could you try use the _get_authenticated_headers from apiimpltype1?  It may fix it without duplicating the actions.

nishant713:
> Could you try use the _get_authenticated_headers from apiimpltype1? It may fix it without duplicating the actions.

I get validation error if I do that

cdnninja:
It is probably removing the import since you haven't deleted the method within the file.  

1. delete the method within the file.
2. Import from the parent.  
3. Adjust all spots it is called to pass the two parameters the parent uses.   
4. remove you start_climate, test if it works using the parent version.

nishant713:
> 1. delete the method within the file.
Did this

> 2. Import from the parent.
This is not needed since the method is inside ApiImplType1 class which is already imported

> 3. Adjust all spots it is called to pass the two parameters the parent uses.
Maybe this is the part I didn't understand? All places where it's called already passed token and optionally ccs2 support 

> 4. remove you start_climate, test if it works using the parent version.
Did this

But the integration fails to load with response code 4017 from server, "Invalid Application Id"

nishant713:
Closed by mistake, creating new PR

nishant713:
> It is probably removing the import since you haven't deleted the method within the file.
> 
> 1. delete the method within the file.
> 2. Import from the parent.
> 3. Adjust all spots it is called to pass the two parameters the parent uses.
> 4. remove you start_climate, test if it works using the parent version.

Ok finally found the issue. _get_authenticated_headers has 2 additional entries compared to the one in KiaUvoApiIN therefore login fails.

But even after changing the header in ApiImplType1 to match it to existing format, climate requests fail with the error-
Server returned: '4005' 'Unsupported control request - Vehicle is not supported (vehicleId: 51f7eb1e-ffb0-41bc-8fb4-b3c109e4235a)'

nishant713:
Found the problem. Latest commit fixes climate, but will break other regions (EU I guess).

I don't have much expertise in coding but may try to give an attempt to fix it later. Or do you want me to fall back to the earlier solution with separate method in KiaUvoApiIN?

nishant713:
@cdnninja I've reverted to the earlier implementation with separate method for _get_authenticated_headers and climate start/stop in KiaUvoApiIN. Can we merge this? I see that CN has a similar implementation.

PR #806, opened 2025-05-07
fix: CA login

cdnninja:


robinlee09201:
Did the MR fix the CA login issue? I still couldn't login though.

cdnninja:
Nope.  Was a hope but needed a release to test.

PR #827, opened 2025-06-27
fix: Update start_climate for older US Hyundai vehicles

mattsch:
This should fix the issues in Hyundai-Kia-Connect/kia_uvo#1135. Wasn't sure of the best method to determine when to include the extra options so based it off of `vehicle.generation`. Happy to update if there's a better way to do this but I was able to test successfully with both a 2025 Ioniq 6 (generation 2) and a 2025 Ioniq 5 (generation 3).


cdnninja:
Thank you for this!

PR #833, opened 2025-07-06
fix: ca cloudscraper version

robinlee09201:
Finally, sorry multiple tries. Not very experienced in the version manager.


cdnninja:
Looks like pypi doesn't like it.   It reference it being an issue with a link to this: https://packaging.python.org/en/latest/specifications/core-metadata/

PR #842, opened 2025-07-19
fix(CA): Add deviceId to login call

dr1rrb:
Add deviceId to login call as discussed here https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1163#issuecomment-3092457705


cdnninja:
Thanks for this!

PR #849, opened 2025-08-02
fix: Hyundai EU API Login

Juinjer:
This fixes the api for Hyundai, let me know if it breaks the Kia api


cdnninja:
Lots of duplicate code. Could you set the variable for the URL up top in the other if statements as a variable and use it below instead?

cdnninja:
Hopefully okay I tried reducing code here. https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/850  Happy to have you submit though for credit.  Could you test this out for Hyundai?

Juinjer:
Thanks @cdnninja, #850 is indeed working for Hyundai, was a little bit slower with the clean up of the duplicate code then you :)

PR #850, opened 2025-08-02
fix:  EU Hyundai Login

cdnninja:
Credit to @Juinjer   #849


cdnninja:
Finally think I got this working.  Genesis probably doesn't.

PR #857, opened 2025-08-10
feat: Add a retry for requests

alphalneo:
From time to time, my login fails with error 104 (reset by peer).
I’m not sure if it’s something related to my network, but this only happens during the login process.
I want to check whether it’s worth pushing the change or keeping it on standby to see if others experience the same problem.
I added the class to the file for the reason I mentioned earlier.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -61,6 +61,7 @@ class KiaUvoApiCA(ApiImpl):
         self.old_vehicle_status = {}
         self.API_URL: str = "https://" + self.BASE_URL + "/tods/api/"
         self.API_HEADERS = {
+            "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:141.0) Gecko/20100101 Firefox/141.0",
             "content-type": "application/json",
             "accept": "application/json",
             "accept-encoding": "gzip",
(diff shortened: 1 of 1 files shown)


cdnninja:
I would think this retry needs to look at error?   In Canada we are fighting with cloud flare blocking, this would make it worse if the block is hit and retried.

alphalneo:
For sure, it should retry only on error 104 (definitely not on any 4xx errors), which was the error I got. I checked the issue section and didn't notice anyone else reporting this problem. It seems that I am the only one experiencing it, as I don't have it when retrieving the data.
I notice from the logs and from the historical graph

Screenshot 2025-08-10 at 5 47 10 PM

If any one report this kind of error it is useless, since it is something on my side

cdnninja:
I have seen others in other regions report it.   I’ll have to search to find it. On Aug 10, 2025, at 2:48 PM, alphalneo ***@***.***> wrote:﻿alphalneo left a comment (Hyundai-Kia-Connect/hyundai_kia_connect_api#857)
For sure, it should retry only on error 104 (definitely not on any 4xx errors), which was the error I got. I checked the issue section and didn't notice anyone else reporting this problem. It seems that I am the only one experiencing it, as I don't have it when retrieving the data.
I notice from the logs and from the historical graph
Screenshot.2025-08-10.at.5.47.10.PM.png (view on web)
If any one report this kind of error it is useless, since it is something on my side

alphalneo:
Only to let you know, I did the mod

Issue #855, opened 2025-08-07
EU login error (again)


Jesus-M:
Hello,

apparently there is again a problem authenticating, same symptoms than https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/847:

```
# Kona-odo.py
Traceback (most recent call last):
  File "/cinemateka/scripts/Kona-odo.py", line 13, in <module>
    vm.check_and_refresh_token()
  File "/usr/lib/python3.12/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 147, in check_and_refresh_token
    self.initialize()
  File "/usr/lib/python3.12/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 82, in initialize
    self.token: Token = self.api.login(self.username, self.password)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 194, in login
    _, access_token, authorization_code, expires_in = self._get_access_token(
                                                      ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 1245, in _get_access_token
    token_type = response["token_type"]
                 ~~~~~~~~^^^^^^^^^^^^^^
KeyError: 'token_type'
#
```

And the module is updated:

```
# pip list | grep hyundai
hyundai_kia_connect_api       3.43.5
#
```

panzerdev:
I get the follwing error msg from the failing call 

```
{'retId': '1e24f681-9c2f-457e-b7c2-caa6ab2735d0', 'retCode': 4000, 'retMsg': 'Bad Request', 'retSubMsg': '"clientSecret is wrong. requested clientId=6d477c38-3ca4-4cf3-9557-2a1929a94654 requested clientSecret=secret"', 'retValue': None}
```

Juinjer:
There was an app update of bluelink yesterday, maybe they disabled access for the old version’s client id.
I’ll try to take a look in the evening

ZuinigeRijder:
[...]

PR #859, opened 2025-08-11
fix: Use old login methods for hyundai in EU

web-dc:
It looks like Hyundai rolled back authentification methods in EU to previous versions.

Therefore I added some condition to differentiate between Hyundai and Kia/Genesis during Authentication process

Potential fix for https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1231 and https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/855


cdnninja:
Thank you for this! Much appreciated!

sqw534:
Many thanks for your support! Great job

PR #863, opened 2025-08-15
fix: Should fix Hyundai Canada Probelm

alphalneo:
I was not able to get someone to test it, I don't know if you want to accept the PR or if you can make a BETA

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,4 @@ python-dateutil
 pytz>=2021.3
 certifi>=2024.6.2
 geopy >= 2.2.0
-cloudscraper @ git+https://github.com/VeNoMouS/cloudscraper.git@master
+cloudscraper >= 1.2.71
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -11,4 +11,4 @@ twine==6.1.0
 watchdog==6.0.0
 wheel==0.45.1
 geopy>=2.2.0
-cloudscraper @ git+https://github.com/VeNoMouS/cloudscraper.git@master
+cloudscraper >= 1.2.71


hyundai_kia_connect_api/KiaUvoApiCA.py
```diff
@@ -39,43 +39,6 @@
...
-    def __init__(self, max_retries=3, delay=2, backoff=2):
-        super().__init__()
-        self.max_retries = max_retries
-        self.delay = delay
-        self.backoff = backoff
-
-    def post(self, url, **kwargs):
-        return self._request_with_retry("POST", url, **kwargs)
-
-    def _request_with_retry(self, method, url, **kwargs):
```
cdnninja, line 52: This section deals with a server disconnect issue where server closes the connection from time to time. Ideally should keep it.

cdnninja (changes requested):

hyundai_kia_connect_api/KiaUvoApiCA.py
alphalneo: Do you release without in order to test the Hyundai problem or you prefer that I will add that session back?

alphalneo: I did it, it seems to run correctly for KIA

cdnninja:
Thoughts on this? https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/1252#issuecomment-3192427739

alphalneo:
> Thoughts on this? [Hyundai-Kia-Connect/kia_uvo#1252 (comment)](https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/1252#issuecomment-3192427739)

This is a bad news :(

robinlee09201:
What's going on? Cloudscraper25 works for me all the time. Why do we switch away from it?

cdnninja:
It wouldn't install for some people - so tried another approach.  Without it being on pypi seems to be a challenge to get this working for all users.  I am thinking we revert to cloudscraper for now, with hopes of finding something that works for all.

PR #896, opened 2025-09-22
fix: Remove external dependencies (curlify, pytz, dateutil), fix Canadian timezone detection

Lekensteyn:
Summary:
* Remove external dependencies (curlify, pytz, dateutil) to reduce security risks of supply chain attacks.
* Improve Canadian timezone detection, updates #187 by @torbjornaxelsson.

Details:

* Remove external pytz dependency

    pytz notes that "Projects using Python 3.9 or later should be using the
    support now included as part of the standard library". Indeed, we can
    replace `pytz.utc` by `datetime.timezone.utc`. Its alias `datetime.UTC`
    requires Python 3.11, so we stick to the former for now.

    Automated import deletions and API rewrites with:

        sed -e '/import pytz/,+1{/^\(import pytz\)\?$/d}' \
            -e 's/pytz\.utc/dt.timezone.utc/' \
            $(git grep -l pytz.utc) -i
        ruff format

* Fix Canada/Newfoundland time zone support

    Canadian users can configure a timezone in their car to one of the
    Canadian timezones, it does not have to support arbitary others. It also
    turns out that Newfoundland is an odd timezone with half an hour diff.

    Refactor the code to check for one of the known timezones instead of
    inventing a dummy "VEHICLETIME" one. This allows dropping a dependency
    on tzoffset from the external dateutil library.

    Drop unused `self.vehicle_timezone`, removed since commit c885a81f2ae6.

* Remove external dateutil dependency

    Use zoneinfo for time zone information, added in Python 3.9. This relies
    on the system time zone database whereas dateutil ships one by itself.

    Automated API and import rewrites with:

        sed -e 's/tz.gettz/ZoneInfo/' \
            -e 's/from dateutil import tz/from zoneinfo import ZoneInfo/' \
            $(git grep -l tz.gettz) -i

* Remove unused curlify dependency, remove requirements_dev.txt

    Removed curlify, a comment to this got removed in commit b861e2d164d2.
    Add requests dependency, this was indirectly pulled in by curlify, but
    it is actually a direct dependency.

    Remove requirements_dev.txt, unused since commit bc735d9754ad.


Lekensteyn:
I removed the geopy change to reduce this PR size and truly make it optional in a new PR (#910). Now the PR is mostly focused on reducing dependencies and modernizing time handling.

cdnninja:
Great work!

cdnninja:
Looking closer at this, I see logic flaws on the canadian side. It should be parsing and using the timezone provided by the API.  Was this AI generated?

Lekensteyn:
I can assure you, this was fully written by hand. The only limitation is that I have no Canadian car, and had to make guesses based on publicly available sources:

* Subset of timezones in http://webmanual.kia.com/PREM_GEN5/AVNT/RJ/USA/English/system007.html
* Available timezones based on local tzdata:
```
$ (cd /usr/share/zoneinfo/ && ls -1d Canada/*) | while read -r tz; do TZ="$tz" date -d"TZ=\"UTC\" 2025-03-01 12:00" +%z\ %Y-%m-%d\ %H:%M\ %Z\ $tz; done | sort -V | uniq -cw26
      1 -0330 2025-03-01 08:30 NST Canada/Newfoundland
      1 -0400 2025-03-01 08:00 AST Canada/Atlantic
      1 -0500 2025-03-01 07:00 EST Canada/Eastern
      2 -0600 2025-03-01 06:00 CST Canada/Central
      2 -0700 2025-03-01 05:00 MST Canada/Mountain
      1 -0800 2025-03-01 04:00 PST Canada/Pacific
```
* The Canadian API response which I also linked in the test suite comments: https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/488#issuecomment-1352038594. For future reference, the relevant subset:
```json
{
  "responseHeader": {
    "responseCode": 0,
    "responseDesc": "Success"
  },
  "result": {
    "transaction": {
      "apiCode": "RMT-E0301",
      "apiStartDate": "20221214192418",
      "apiEndDate": "20221214192418",
      "apiResult": "C",
      "apiStatusCode": "200"
    },
    "vehicle": {
      "lastStatusDate": "20221214112426",
```

There is no timezone information that response. There is an `apiStartDate` that could potentially be used instead of the current date, but since I cannot test such a change, I preserved the existing previous logic of comparing the current time against `lastStatusDate`.

I did change `self.data_timezone` in `last_updated_at = parse_datetime` to `dt.timezone.utc` since that is essentially the same for `KiaUvoApiCA`, but more explicit.

I also confirmed that Python datetime math respect the timezone, that's why I can use `ref_date = dt.datetime.now()` without forcing UTC.

My primary goal was to remove unnecessary dependencies while trying my best to avoid regressions. Could you point out the suspected logic flaw? By "API", do you mean the Python API (e.g. `self.data_timezone`), or the cloud API?

cdnninja:
This was the issue I am seeing with the code: 

https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1316

By API I meant hyundai Kia.  I stand corrected - looks like this region doesn't provide the tz in the response.    Sorry for the AI comment - I find I am getting pulls with that sort of approach and makes it hard to know.

Lekensteyn:
I mistakenly assumed that `now()` includes a timezone, fixed that in #908. The tests did not catch this since I already included a timezone there, oops.

Do any of the regions include the timezone in the API? All region implementations have a hard-coded timezone, for example:
https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/55cfd910254ce6615959c4bfa1206c09abdbe0e3/hyundai_kia_connect_api/KiaUvoApiEU.py#L77-L78

In `KiaUvoApiCA.py`, `data_timezone` is not overridden and defaults to UTC:
https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/55cfd910254ce6615959c4bfa1206c09abdbe0e3/hyundai_kia_connect_api/ApiImpl.py#L72-L73

There is special code to autodetect the timezone for `last_updated_at`, but `_update_vehicle_properties_location` uses `self.data_timezone` (UTC, not the autodetected timezone)! Is this inconsistency expected?

cdnninja:
The states uses something called "offset" as a value.  Only one off the top of my head.

PR #903, opened 2025-09-26
fix(EU): ev_charging_power for EU BEVs

wabu:
97e100b did introduce a bug, as the ev_charging_power will not be populated when the vehicle currently is not charging, e.g. power is 0.


wabu:
This should fix Hyundai-Kia-Connect/kia_uvo#1321

PR #905, opened 2025-09-26
fix: if logic for charging power

robotfishe:
This PR fixes an issue with the ev_charging_power failing to update instead of showing as zero, addressing recent updates to issue 1305.


cdnninja:
I merged the other PR when on my phone so I think this is one is needed now?

PR #908, opened 2025-09-28
fix: restore explicit timezone for Canadian responses

Lekensteyn:
Fixes a regression introduced with a previous Canadian timezone fix, `datetime.now()` requires an explicit timezone parameter to prevent:

      File "hyundai_kia_connect_api/utils.py", line 107, in detect_timezone_for_date
        delta = (ref_date - date.replace(tzinfo=tz)).total_seconds()
                 ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
    TypeError: can't subtract offset-naive and offset-aware datetimes

Link: https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1316#issuecomment-3341841390
Fixes: 753076d62607 (#896)


cdnninja:
Thank you!

PR #920, opened 2025-10-15
feat: Hyundai refresh token login

Maaxion:
Change `KiaUvoApiEU.py` so it also allows refresh token based login for Hyyundai

PR #922, opened 2025-10-16
fix: Fixes login for Kia vehicles that got broken by #920

Triple-S:
Hi,

this is a small fix to restore the exact behaviour of the function `_get_access_token` for Kia and Genesis vehicles as it was before #920. Fixes Hyundai-Kia-Connect/kia_uvo#1347

Best regards,
Triple-S


fuatakgun (approved):

fuatakgun:
Thanks for this

PR #879, opened 2025-09-05
fix(EU):  Genesis EU authentication #514, #638 

martinpelant:
This is rebased https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/638
Tested with my GV60 and Hyundai Kona (UK)

PR #935, opened 2025-10-21
fix(EU): connection with Genesis (EU)

juodikis:
This is rebased https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/879

Basically, now it uses same approach as KIA (by providing refresh_token instead of password).

Tested using simple:

```python
vm = VehicleManager(region=1, brand=3, username="email@gmail.com", password="refresh_token", pin="")
vm.check_and_refresh_token()
print(vm.vehicles)
``` 

Still haven't figured out best way to extract Genesis token.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -40,9 +40,6 @@ from .const import (
     TEMPERATURE_UNITS,
     VALET_MODE_ACTION,
 )
-from .exceptions import (
-    AuthenticationError,
-)
 from .utils import (
     get_child_value,
     get_hex_temp_into_index,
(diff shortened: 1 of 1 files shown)


RustyDust:
@juodikis
> Still haven't figured out best way to extract Genesis token.

I'm currently "sort of" maintaining the Hyundai/Kia integration for [evcc](https://github.com/evcc-io/evcc) and would like to extend that to Genesis as well. So I'd also be interested in ways to fetch the refresh_token for Genesis and thus my question: Got any pointers for me, preferably a modified script like the ones for [Kia](https://gist.githubusercontent.com/fuatakgun/fa4ef1e1d48b8dca2d22133d4d028dc9#gistfile1.txt) or [Hyundai](https://gist.github.com/Maaxion/22a38ba8fb06937da18482ddf35171ac#file-gistfile1-txt) which I unified [here](https://gist.github.com/RustyDust/e2a7be978affd85fb5ef5a345f31f67a).

And, of course, I'd be glad if you could test the evcc integration once I got a PR ready, since I obviously don't own a Genesis and have no way to test ;)

AndreasLat:
Any news in this topic? A have a Genesis GV70 and would like to support any integration.

PR #937, opened 2025-10-21
fix: bluelink command json output

### LINKED ISSUE #936: `bluelink` command error: TypeError: Object of type TripInfo is not JSON serializable
I'm about to submit a PR to fix this:

```shell
bluelink --region USA --brand Hyundai --username <cut> --password <cut> info
```

```
...
Debug
  File "/Users/patrick/Code/python/hyundai_kia_connect_api/.venv/bin/bluelink", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/patrick/Code/python/hyundai_kia_connect_api/hyundai_kia_connect_api/bluelink.py", line 453, in main
    return args.func(vm, args)
           ~~~~~~~~~^^^^^^^^^^
  File "/Users/patrick/Code/python/hyundai_kia_connect_api/hyundai_kia_connect_api/bluelink.py", line 372, in cmd_info
    print_vehicle(vehicle)
    ~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/patrick/Code/python/hyundai_kia_connect_api/hyundai_kia_connect_api/bluelink.py", line 196, in print_vehicle
    print(textwrap.indent(json.dumps(vehicle.data, indent=2, sort_keys=True), " "))
                          ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/.local/share/uv/python/cpython-3.13.3-macos-aarch64-none/lib/python3.13/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ~~~~~~^^^^^
  File "/Users/patrick/.local/share/uv/python/cpython-3.13.3-macos-aarch64-none/lib/python3.13/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/patrick/.local/share/uv/python/cpython-3.13.3-macos-aarch64-none/lib/python3.13/json/encoder.py", line 261, in iterencode
    return _iterencode(o, 0)
  File "/Users/patrick/.local/share/uv/python/cpython-3.13.3-macos-aarch64-none/lib/python3.13/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
                    f'is not JSON serializable')
TypeError: Object of type TripInfo is not JSON serializable
```

plasticrake:
`TypeError: Object of type TripInfo is not JSON serializable`

Fixes #936

### DIFF
--- a/hyundai_kia_connect_api/bluelink.py
+++ b/hyundai_kia_connect_api/bluelink.py
@@ -3,6 +3,7 @@
 """Connects to the Bluelink API and query the vehicle."""
 
 import argparse
+import dataclasses
 import datetime
 import json
 import logging
@@ -12,6 +13,13 @@ import textwrap
 
 import hyundai_kia_connect_api
 from hyundai_kia_connect_api import const
+
+
+class EnhancedJSONEncoder(json.JSONEncoder):
+    def default(self, o):
+        if dataclasses.is_dataclass(o):
+            return dataclasses.asdict(o)
+        return super().default(o)
 
 
 def print_vehicle(vehicle):
@@ -193,7 +201,12 @@ def print_vehicle(vehicle):
     print("  month_trip_info:", vehicle.month_trip_info)
     print("  day_trip_info:", vehicle.day_trip_info)
     print("Debug")
-    print(textwrap.indent(json.dumps(vehicle.data, indent=2, sort_keys=True), " "))
+    print(
+        textwrap.indent(
+            json.dumps(vehicle.data, indent=2, sort_keys=True, cls=EnhancedJSONEncoder),
+            " ",
+        )
+    )
 
 
 def vehicle_to_dict(vehicle):


cdnninja:
Thoughts on doing this up front when we store the value vs on display?   As a side note I am interested why you are accessing the raw data vs values on the vehicle side.  Are some values missing?

plasticrake:
Hi @cdnninja, I'm not understanding what you mean by "accessing the raw data vs values on the vehicle side". The `bluelink` command line tool fails for me because it is trying to output as JSON but it cannot serialize it. I'm not changing how anything is accessed. It is already using `json.dumps`, I'm just fixing it to serialize dataclasses which it what `TripInfo` is.

cdnninja:
I mean adjusting it here so all downstream items benefit: 

https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/1efd431c9f9966d702e13fc1d38e397ca2411d90/hyundai_kia_connect_api/KiaUvoApiEU.py#L768C24-L768C29

As for accessing vehicle object.  Accessing items via vehicle.fuel_level or vehicle.daily_stats. 

This bluelinky file is somewhat of a test file.

mmerickel:
Ran into this issue and can confirm the PR fixes it. To be clear the `bluelink info` cli command is just broken right now because of the Debug section trying to json.dumps the `vehicle.data` attribute which has `TripInfo` objects, instead of just primitive types. It seems reasonable to just add the custom encoding here since other places in the code probably prefer to have structured `TripInfo` objects.

plasticrake:
@cdnninja I'm not very familiar with this codebase or where this package is used elsewhere, but I figured changing it there would be a breaking change as it is changing the API. This was the safer non-breaking option.

Issue #939, opened 2025-10-27
Vehicles list remains empty forever after initially failing to fetch


jstammi:
1. create `VehicleManager`
2. call `check_and_refresh_token` and assume it actually failing for fetching the vehicles list, login before succeeded (e.g. requests limit exceeded)
3. call same method again before token expires (3600s actually in the below linked case) and assume the limit was lifted meanwhile (e.g. at midnight)
4. call any method dealing with the vehicles list, and nothing will happen - the vehicles list is empty. And this will never change again until a new `VehicleManager` instance is used (or one manually removes the token) :-(

(seen in the logs of https://github.com/ZuinigeRijder/hyundai_kia_connect_monitor/issues/89)

PR #940, opened 2025-10-27
fix: ensure vehicles list is initialized after token check

jstammi:
Until having once fetched successfully the list of vehicles, re-try to do so on each token validation/renewal again.
Else the `check_and_refresh_token` raises an exception the same way again as on (partial) `initialize`.

Note: as the login may be expensive afaik (logins limit in us?), I rejected the idea to dismiss the token already having retrieved before and stay with a partially initialized instance until next `check_and_refresh_token` call

#939


ZuinigeRijder (commented):
I do not see any problems with these changes, it makes sure that vehicles is initialized. @cdnninja can you merge this, or should I do it?

PR #945, opened 2025-11-03
fix(EU): warn on incorrect password usage

Lekensteyn:


Lekensteyn:
Sending the cleartext password to the oauth2 token API will never work, a proper check like this should save users some time. Ideally we can add a link to the wiki with better instructions.

I am currently using a refresh token extracted using mitmproxy. That, and the selenium scripts are definitely not the most user friendly approaches. Hopefully I can find an alternative that only requires a browser.

cdnninja:
Thanks for doing this!

PR #957, opened 2025-11-25
fix(CA): Enhance API request headers in set_charge_limits

JB12134:
Added additional headers for API request in set_charge_limits method. This fixed https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1393 for me, but someone else should test, if possible. I checked that the other functions still worked as well.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -785,7 +785,9 @@ class KiaUvoApiCA(ApiImpl):
                 },
                 "pin": token.pin,
             }
-        _LOGGER.debug(f"{DOMAIN} - Planned start_climate payload {self._mask_sensitive_data(payload)}")
+        _LOGGER.debug(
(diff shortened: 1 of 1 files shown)


hyundai_kia_connect_api/KiaUvoApiCA.py
```diff
@@ -940,9 +944,9 @@ def set_charge_limits(
             "pin": token.pin,
         }
 
+        _LOGGER.debug(f"{DOMAIN} - Planned set_charge_limits payload {payload}")
```
cdnninja: I am thinking we need to remove this or mask it. Since it will provide PINs in log files and users typically don't remove items like that.

cdnninja (changes requested):
Thank you for doing this!

JB12134:
No worries! That's fair, I think it's exposed in the climate section too.

JB12134:
Added a function to mask so we could keep in the debugging logs, but I don't know if it is best to keep it where it is, I can remove that and the logging with PIN if you prefer. This worked well in my tests.

PR #958, opened 2025-11-28
feat: Add light sensors and other entities from logs to CA and EU

JB12134:
Adding light status sensors for CA and EU, not sure if they exist anywhere else. There will be a PR on kia_uvo as well to match these changes. I don't know how to do both at once, but let me know if there's a way. It does seem to work in my homeassistant. 

https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1394


cdnninja:
This is correct, mapping doesn't always align between regions so lots of duplication right now.

PR #962, opened 2025-12-01
fix(BR): Use hardcoded device_id instead of random UUID

eduardobittencourt:
## Summary
- Brazilian API rejects random UUIDs for `device_id`, returning `400 Bad Request` with "Invalid deviceId" error
- Changed to use the registered `ccsp_device_id` instead of generating a new UUID on each login

## Test plan
- [x] Tested login flow with real account
- [x] Verified vehicle listing works
- [x] Verified vehicle status retrieval works
- [x] Verified location retrieval works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,3 @@
+KIA_US_USERNAME=your_username_here
+KIA_US_PASSWORD=your_password_here
+KIA_US_PIN=your_pin_here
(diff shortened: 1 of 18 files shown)


cdnninja:
Could you resolve the conflict? I will then merge.

PR #967, opened 2025-12-09
fix: reuse persisted refresh token for EU/CA flows

renatkh:
## Summary
- regression: 2.47.1 stores only refresh_token/device_id for EU/CA entries, but VehicleManager kept forwarding `None` to regional login() calls, so tokens expired after a restart
- fix: VehicleManager now reuses the best available secret (password → token.refresh_token → token.password) and regression tests cover both initialize() and refresh paths

## Testing
- pytest tests/vehicle_manager_test.py   # requires pytest in env

### DIFF
--- a/docs/HASSIO_TESTING.md
+++ b/docs/HASSIO_TESTING.md
@@ -62,4 +62,3 @@ Verify that the reported version is `99.99.99` and the location points to the si
 4. Leave the integration running for at least an hour to confirm the token refresh continues to work with the new API build.
 
 Please report whether the test passes and which region/brand you exercised so we can ensure non-US regions remain healthy.
-


renatkh:
@cdnninja this may fix the issue

renatkh:
@cdnninja do not merge this PR. I bumped the version in setup.py to 99.99.99 for testing.

cdnninja:
I am thinking we actually roll back the OTP work.    Start again with a more universal approach.

renatkh:
@cdnninja Let's see how the testing goes. If users still have issues, we will roll it back.

fuatakgun:
I am also inclined to;
1. Rollback to safe point 
2. Release a test branch with these changes

cdnninja:
Closing as role back has been done.   I believe these changes were tested unsuccessfully.  Approaching this here: #972

Issue #972, opened 2025-12-14
OTP Thoughts


cdnninja:
@renatkh  If interested to take another stab at OTP let me know.  Here is what I am thinking from what I read in the code. 

1. Need to implement pass through / storage of device ID
2. Need to support passing token in.   
3. Need to support a login response that informs caller to trigger OTP.   This should be done in vehicle manager.  Universally so that any region can offer it.  

I think we start at the top and work through each one at a time.  That way we limit changes.   

Let me know if interested!

cdnninja:
After reading more.   The first isn't needed.  DeviceID is part of token.   I dropped a PR that will solve what may have been the core bug on setup, plus made PIN follow the standard flow.   I will leave open for review for a bit then merge.   After that I think token needs some clean up since it doesn't need to be passed into the API login command.

renatkh:
@cdnninja sure, I just came back from a business trip and plan to take a look at it today. A tricky part is to deploy changes in both repos in a compatible way. We will also need someone from EU to test things.

cdnninja:
@renatkh  I think the key is smaller bite size chunks.    I should have taken a closer read through how it was done.    We can poke for testers but hard to get them.    Smaller chunks though will make it easier to spot the issues when reported.  

For flow I am thinking vehicle manager may need three methods that in turn pass on through the APIs. Central dataclass needs to be build out in a universal way so always the same between regions. 

1. login --> Returns token or OTPData depending if OTP response was required. 
2. send_OTP which the selection is sent to and OTP is sent.  Not sure if anything needs to be returned beyond true if no errors. 
3. complete_otp_login -- returns a functional token. 


If a token is "restored" on reboot it would be provided to the library when setup.  In which case check_and_refresh won't trigger a login if it is valid.  

Thoughts?

cdnninja:
@renatkh  any thoughts on above? I have gone ahead and merged those first two fixes.   Going to release it to users to confirm other regions are good.  Next up is cleanup of the kia USA file and stubbing out the above assuming agreed to.

cdnninja:
#976  is a base for how I am thinking OTP could work.  I believe it is a none breaking change.   I don't love the login returning Token or OTPDetails.  Not sure the best way to avoid a is instance check as I believe those are expensive.

renatkh:
@cdnninja I can test it. What behavior do you expect? Should it go all the way through otp process and login in hassio?

[...]

PR #973, opened 2025-12-14
fix: Pin correction, vehicle refresh

cdnninja:
This correction passes pin up front vs after the fact.   It also corrects a situation where if we start passing tokens the vehicle list isn't initiated. 

#972


cdnninja:
@renatkh  once reviewed I will merge and release this one.  Same goes for the removing token one, that will need an additional PR to clean up the USA kia library file though.

PR #983, opened 2025-12-29
fix: AC/DC charge limit settings not working on newer Gen3 vehicles

killerbee84:
This fixes the AC and DC charge limit settings on Gen3 vehicles like the IONIQ 9.
Currently, changing the limits inside HA does not work.


hyundai_kia_connect_api/Vehicle.py
```diff
@@ -80,6 +80,9 @@ class Vehicle:
     # Not part of the API, enabled in our library for scanning.
     enabled: bool = True
 
+    # Generation type (e.g. "G3")
+    gen_type: str = None
```
cdnninja: Lets use generation two fields up for this.

cdnninja (changes requested):
Looks like pre-commits are failing.

killerbee84:
I can't fix the pre-commit issue. The API path is "v2/ev/reserv/socset" and pre-commit is complaining about the "reserv".

cdnninja:
I think you can exempt it from code spell.   If we merge as is it will fail forever so needs fixing. I also updated the branch, test snapshots need to be updated.

cdnninja:
Some of this was actually my recent PR by pytest snapshot update isn't working...ugh.

cdnninja:
> I think you can exempt it from code spell. If we merge as is it will fail forever so needs fixing. I also updated the branch, test snapshots need to be updated.

I wanted to follow up if you could fix codespell?   If interested in having this merged still.

blka:
I looked at this PR and see a few issues that need resolving before it can merge:

1. **Type conflict on `generation`** — USA uses `generation: int` (values 2, 3 from `vehicleGeneration`), this PR assigns `genType` (string "G3") to the same field. `typing.Union[int, str]` is a workaround, not a clean solution. Better options:
   - Add a separate `gen_type: str` field for CA/AU string-based generation identifiers
   - Or map "G3" → 3 to keep `int`, but we'd need to know the full set of values (G1? G2?)

2. **No CA response data with `genType`** — There are no test fixtures or sample responses confirming that `genType` actually exists in the CA API vehicle list response. Before changing the type, we need a real API response showing this field.

3. **Codespell** — `reserv` in the URL path `v2/ev/reserv/socset` triggers codespell. Add `reserv` to `--ignore-words-list` in `.pre-commit-config.yaml` alongside the existing `fro,hass,fatc`.

4. **Hardcoded Referer** — `set_charge_limits` has `Referer: https://kiaconnect.ca/remote/` hardcoded, but `KiaUvoApiCA` serves Kia, Hyundai, and Genesis with different `BASE_URL`s. Should use `f"https://{self.BASE_URL}/remote/"`.

5. **Snapshot tests** — Adding `genType` to `get_vehicles` will change Vehicle init and require `pytest --snapshot-update`.

Happy to pick this up and implement it properly if someone can provide a CA API vehicle list response confirming the `genType` field.

cdnninja:
@killerbee84  any update on this?

PR #1004, opened 2026-01-13
fix(EU): fix correct location key

stevietv:
fixes #1003 

sorry, i misinterpreted the key structure, this corrects it according to: 

```json
{'retCode': 'S', 'resCode': '0000', 'resMsg': {'gpsDetail': {'coord': {'lat': xx.xxxxx, 'lon': yy.yyyyy, 'alt': 0, 'type': 0}, 'head': 0, 'speed': {'value': 0, 'unit': 0}, 'accuracy': {'hdop': 0, 'pdop': 0}, 'time': '20260111172656'}, 'drvDistance': {'rangeByFuel': {'evModeRange': {'value': 165, 'unit': 1}, 'totalAvailableRange': {'value': 165, 'unit': 1}}, 'type': 2}}, 'msgId': '...msgid...'}
```

wanted to forcepush the pr #1003 but you were quick!

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -813,7 +813,7 @@ class KiaUvoApiEU(ApiImplType1):
             ).json()
             _LOGGER.debug(f"{DOMAIN} - _get_location response: {response}")
             _check_response_for_errors(response)
-            return response["resMsg"]["gpsDetail"]["coord"]
+            return response["resMsg"]["gpsDetail"]
         except Exception as e:
             _LOGGER.error(f"{DOMAIN} - _get_location failed: {e}", exc_info=True)
             return None


cdnninja:
To confirm this would work for just the one structure? The one pr implies two may exist?

stevietv:
looking into this more:

the update vehicle method expects the location in the state to be a dictionary containg a coord dict, so infact we shouldn't be returning the contents of the coord dict, but the gpsDetail dict:

https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/6d91c9448f8b75ee7ceda08a20a2ced1ac4bb1b8/hyundai_kia_connect_api/KiaUvoApiEU.py#L768-L776

the _get_location method should therefore result in saving the gpsDetail dictionary into the vehicleLocation state, so that coord is part of that.

sorry for the confusion here., this additonal commit should fix it

cdnninja:
Thanks for the work on this!

PR #1024, opened 2026-01-31
fix: parse targetSOC from force refresh response for USA Kia API

kingpanther13:
Root cause: The cmm/gvi (cached state) endpoint does NOT include targetSOC data for some vehicles (confirmed with 2020 Kia Niro EV). The rems/rvs (force refresh) endpoint DOES include it. Previously, the force refresh response was discarded without parsing targetSOC, so ev_charge_limits_ac/dc were always None for affected vehicles.

Changes:
- force_refresh_vehicle_state now captures the rems/rvs response and parses targetSOC from it via _update_charge_limits_from_force_refresh
- _update_vehicle_properties no longer overwrites existing charge limit values with None when the cached response lacks targetSOC
- Added unit tests using real API response data from a 2020 Kia Niro EV

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiIN.py
+++ b/hyundai_kia_connect_api/KiaUvoApiIN.py
@@ -356,6 +356,39 @@ class KiaUvoApiIN(ApiImplType1):
                 DISTANCE_UNITS[get_child_value(state, "dte.unit")],
             )
 
+        vehicle.washer_fluid_warning_is_on = get_child_value(state, "washerFluidStatus")
+        vehicle.accessory_on = bool(get_child_value(state, "acc"))
+        vehicle.ign3 = bool(get_child_value(state, "ign3"))
+        vehicle.transmission_condition = bool(get_child_value(state, "transCond"))
+        vehicle.sleep_mode_check = bool(get_child_value(state, "sleepModeCheck"))
+        vehicle.headlamp_status = bool(
+            get_child_value(state, "lampWireStatus.headLamp.headLampStatus")
+        )
+        vehicle.headlamp_left_low = bool(
+            get_child_value(state, "lampWireStatus.headLamp.leftLowLamp")
+        )
+        vehicle.headlamp_right_low = bool(
+            get_child_value(state, "lampWireStatus.headLamp.rightLowLamp")
+        )
+        vehicle.stop_lamp_left = bool(
+            get_child_value(state, "lampWireStatus.stopLamp.leftLamp")
+        )
+        vehicle.stop_lamp_right = bool(
+            get_child_value(state, "lampWireStatus.stopLamp.rightLamp")
+        )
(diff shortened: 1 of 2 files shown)


tests/us_target_soc_test.py
```diff
@@ -0,0 +1,242 @@
+"""Test targetSOC parsing for USA Kia API.
```
cdnninja: Love seeing this!  One thing I would love to start seeing is a folder with json responses that we can parameterize to the tests to run through variations of cars in builds.  Just a thought for future if interested. :)

hyundai_kia_connect_api/KiaUvoApiUSA.py
```diff
@@ -770,6 +791,46 @@ def _get_forced_vehicle_state(self, token: Token, vehicle: Vehicle) -> dict:
         response_body = response.json()
         return response_body
 
+    def _update_charge_limits_from_force_refresh(
```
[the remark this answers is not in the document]
kingpanther13: To be up front, I use Claude code for pretty much everything ...I asked it to figure out why I don't see what my charge limit is set to and it turned into all of this. Per Claude Code: No, the force refresh (rems/rvs) response structure is NOT the same as the cached (cmm/gvi) response. They can't go through the same _update_vehicle_properties method without losing data.

The cached response includes vehicleConfig (odometer, maintenance), location (GPS), activeDTC, and service data - none of which exist in the force refresh response. The force refresh only returns the vehicleStatusRpt block.

If we wrapped the force refresh payload and fed it through _update_vehicle_properties, it would populate vehicleStatus fields correctly but would null out odometer, location, service distances, and DTC data (since get_child_value returns None for missing paths, and those values get unconditionally assigned).

The vehicleStatus block itself is structurally identical between the two responses - same fields, same nesting. The force refresh just has extra fields the cached one lacks (targetSOC, v2lStatus, v2xStatus, gasModeRange).

That's why the current approach does:

update_vehicle_with_cached_state first → populates everything from the full cmm/gvi response
_update_charge_limits_from_force_refresh second → extracts just targetSOC that cmm/gvi omits
The alternative would be to refactor _update_vehicle_properties to accept the force refresh structure too (e.g., by making it path-prefix-aware, or by splitting it into sub-methods), but that's a larger refactor. The comment in the existing code (# Force update needs work to return the correct data for processing) suggests this was already a known issue.

tests/us_target_soc_test.py
kingpanther13: Thanks! That's a great idea - having a folder of real JSON responses from different vehicles/model years would make it much easier to catch regressions and test edge cases. Personally I can only provide a 2020 Kia Niro EV data to help, but I'll be Happy to help set that up in a follow-up PR

kingpanther13: I got it started, it's PR #1025, currently it only has the kia USA stuff but I have Claude code churning up all the other options.

PR #1026, opened 2026-01-31
feat:Enhanced Bluelink for Hyundai cars - India

S-cripts:
feat: Added the following features for Hyundai cars, region India. I have tested in Hyundai EV vehicle only

1. Added vehicle status like washer fluid status, head and tail lamp status, signal indicator status, ignition state
2. Added EV specific properties - Battery SoC, charging status, estimated charging time, driving range (EV and total)
3. Hyundai cars (at least EV) does not provide details trip details like total time, top and avg. speed, in the "/tripinfo" API. Instead, it's provided using "/tripinfo/detail" API. Have handled it without affecting existing API impl


cdnninja:
Looks good to me!   Can I merge?

S-cripts:
> Looks good to me! Can I merge?

Yes pls

Issue #1023, opened 2026-01-30
Kia Connect in Canada now requires MFA code sent by email


Tichael:
As I was working on #1019 and was ready to test what I have, Kia just added MFA to Kia Connect in Canada... This breaks the client from doing any requests to the API. The MFA sends an email to the user with a 6 digits code, and you can check a box to not ask again for 90 days.

I know there are other regions that have MFA, but I don't know how they work and how this project work with those. My guess is it was added in Canada in the last two days, because I last worked on my feature on Wednesday and everything was working fine.

cdnninja:
I'm happy to help answer questions and walk you through it.  If you are willing to sniff and build the PR for MFA?

Overall library part is straightforward if we know what calls it makes.

ppruneau:
I don't know if it can help, but here are the logs I got :

```
Logger: custom_components.kia_uvo.coordinator
Source: helpers/update_coordinator.py:416
integration: Hyundai / Kia Connect ([documentation](https://github.com/Hyundai-Kia-Connect/kia_uvo), [issues](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues))
First occurred: 7:27:47 AM (199 occurrences)
Last logged: 11:46:46 AM

Unexpected error fetching kia_uvo data
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 416, in _async_refresh
    self.data = await self._async_update_data()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/kia_uvo/coordinator.py", line 115, in _async_update_data
    await self.async_check_and_refresh_token()
  File "/config/custom_components/kia_uvo/coordinator.py", line 177, in async_check_and_refresh_token
    await self.hass.async_add_executor_job(
        self.vehicle_manager.check_and_refresh_token
    )
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.13/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 213, in check_and_refresh_token
    result = self.api.refresh_access_token(
        self.token,
    )
  File "/usr/local/lib/python3.13/site-packages/hyundai_kia_connect_api/ApiImpl.py", line 335, in refresh_access_token
    return self.login(
[...]

PR #1033, opened 2026-02-11
feat: Implement OTP codes for Canada

DavidBilodeau1:
## 📖 Description

A couple of weeks ago (Around january 14th), Kia has implemented OTP codes in Canada that need to be verified every 90 days. This stopped the integration from working correctly, as the login API endpoint now returned 7110 Failure.

This Pull request implements the OTP validation process in the API. Another pull request will be required in the kia_uvo integration to support the OTP process. 

## 👷 Work Done

* Investigation of the OTP process in Canada with the help of Tichael (will be credited in comments) and @cdnninja 
* Code started by @cdnninja 
* Implementation of missing methods such as "send_otp" "validate_otp_and_complete_login"

## 📓 Reference

https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/1023

## 🎉 Result

The kia_uvo integration now correctly supports OTP codes for Canada

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -346,7 +346,7 @@ class KiaUvoApiCA(ApiImpl):
             "otpEmail": otp_request.email,
             "mfaApiCode": "0107",
             "otpValidationKey": otp_validation_key,
-            "mfaYn": "N",
+            "mfaYn": "Y",
         }
 
         genmfatkn_response = self.sessions.post(


DavidBilodeau1:
@cdnninja @Tichael here is the PR as previously discussed

hyundai_kia_connect_api/KiaUvoApiCA.py
```diff
@@ -171,32 +181,190 @@ def login(
...
+            username=username,
+            password=password,
+            access_token=access_token,
+            refresh_token=refresh_token,
+            valid_until=valid_until,
+            device_id=device_id,
+            pin=pin,
+        )
+
+    def send_otp(self, otp_request: OTPRequest, notify_type: OTP_NOTIFY_TYPE) -> str:
```
cdnninja: This definition conflicts with the inherited return type.  https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/9ac511b9883e082763c2fb8dd17f00fe0a7c6c39/hyundai_kia_connect_api/ApiImpl.py#L105

Can we return None and it still work?

hyundai_kia_connect_api/KiaUvoApiCA.py
```diff
@@ -171,32 +181,190 @@ def login(
...
+            "mfaApiCode": "0107",
+            "otpValidationKey": otp_validation_key,
+            "mfaYn": "N",
+        }
+
+        genmfatkn_response = self.sessions.post(
+            genmfatkn_url, json=genmfatkn_data, headers=genmfatkn_headers
+        )
+        _LOGGER.debug(
+            f"{DOMAIN} - Generate MFA Token Response {genmfatkn_response.text}"
```
cdnninja: Ideally we don't log tokens, users don't remove them when they dump logs to us.  Not sure if this exists in other spots.

cdnninja (changes requested):
Great work! A few comments.

DavidBilodeau1:
Accidental closing because of a wrong revert

hyundai_kia_connect_api/KiaUvoApiCA.py
cdnninja: Where is this used though?  Since kia_uvo doesn't consume it.    Reason I wonder if it can be removed is its being set on line 305 in the object.  If this return type truly needs to be updated in kia_uvo we need to update other regions and the implementation so trying to avoid that.

PR #1071, opened 2026-03-27
fix(CA): Skip Canadian timezone detection for fresh UTC timestamps 

Juror2372:
## Skip Canadian timezone detection for fresh UTC timestamps

> ⚡ This change was vibe coded.

### Problem

After a forced vehicle state refresh, the API returns `lastStatusDate` as a UTC timestamp. The previous code would unconditionally pass this to `detect_timezone_for_date()`, which tries to reinterpret the timestamp as a Canadian local time. Since the delta between a fresh UTC timestamp and the current UTC time is near zero, the timezone guessing would always fail — producing a noisy warning log even when everything was working correctly.

### Solution

Add a guard check before attempting Canadian timezone detection: if the absolute delta between `lastStatusDate` and the current UTC time is less than 20 minutes, the timestamp is assumed to already be in UTC and the guessing step is skipped silently.

This avoids false-positive warnings on fresh forced-refresh responses while preserving the existing timezone detection logic for cached/stale timestamps where guessing is meaningful.

### Changes

- `force_refresh_vehicle_state`: compute `raw_delta_seconds` upfront and skip `detect_timezone_for_date()` when `abs(raw_delta_seconds) < 20 * 60`
- Log a debug message (instead of a warning) when the skip occurs, explaining the reason
- Warning log for failed timezone detection now uses the pre-computed `raw_delta_seconds` variable for consistency

### Testing

- Forced refresh with a live vehicle should no longer produce a spurious timezone warning in logs
- Cached state updates with stale timestamps should continue to have timezone detection applied as before

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -851,14 +851,15 @@ class KiaUvoApiCA(ApiImpl):
 
     def get_location(self, token: Token, vehicle: Vehicle) -> dict:
         url = self.API_URL + "fndmcr"
-        headers = self.API_HEADERS
+        headers = self.API_HEADERS.copy()
         headers["accessToken"] = token.access_token
         headers["vehicleId"] = vehicle.id
+        headers["from"] = "SPA"
+        headers["Referer"] = f"https://{self.BASE_URL}/remote/"
         try:
             headers["pAuth"] = self._get_pin_token(token, vehicle)
-
             response = self.sessions.post(
-                url, headers=headers, data=json.dumps({"pin": token.pin})
+                url, headers=headers, json={"pin": token.pin}
             )
             response = response.json()
             _LOGGER.debug(f"{DOMAIN} - Get Vehicle Location {response}")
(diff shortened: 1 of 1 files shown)


cdnninja:
Do no parts of the API response clarify the time zone directly?

Juror2372:
not according to claude code:

The CA API returns lastStatusDate as a bare YYYYMMDDHHMMSS string (e.g. "20240801093000") with no accompanying timezone field — no timeZone, offset, utcOffset, or similar. The timestamp silently reflects whatever timezone the car's entertainment system is set to, which could be any of the six Canadian zones. The detection approach (comparing the timestamp against current UTC with each candidate timezone) is the only mechanism available.

The fix doesn't change the guessing logic at all — it just adds a short-circuit: when the delta is < 20 minutes, the timestamp is clearly "just now" in UTC (a fresh forced-refresh response), so there's nothing useful to guess and no warning should fire.

Lekensteyn:
Do you have logs with the API response for the cached and uncached cases? There was previously a report in https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/931#issuecomment-2353590987 for a car in the EU region where cached results contain the time in UTC (workaround #643).

Can you try to pass it through `jq` to pretty-print the JSON and run a `diff` between the cached and uncached responses? I wonder if there is anything we can use from the API responses to avoid stacking hacks on hacks.

PR #1076, opened 2026-04-02
feat(NZ): add browser-based OAuth auth for New Zealand

thesquib:
The NZ Kia Connect signin endpoint (/api/v1/user/signin) uses TLS fingerprinting to reject non-browser HTTP clients. Python's requests library cannot authenticate NZ users via the standard password flow.

Changes:
- KiaUvoApiAU.login(): raises AuthenticationError immediately for NZ with a message directing callers to use the browser-based flow
- KiaUvoApiAU.get_authorize_url(): returns the OAuth authorize URL for the user to open in a browser
- KiaUvoApiAU.login_with_auth_code(): exchanges a browser-obtained auth code for tokens, bypassing the fingerprinted signin endpoint; stores the raw OAuth refresh token so renewal works without re-authentication
- KiaUvoApiAU.refresh_access_token(): overridden for NZ to use grant_type=refresh_token with the stored token rather than calling login() again
- ApiImpl: adds get_authorize_url() and login_with_auth_code() stubs raising NotImplementedError for regions that don't use browser auth
- VehicleManager: adds get_authorize_url() and login_with_auth_code() pass-throughs
- _get_authorization_code_with_redirect_url(): adds mobileNum field to match what the NZ web app sends

Tested and confirmed working on a real NZ Kia Connect account.

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -335,3 +335,29 @@ class ApiImpl:
         return self.login(
             username=token.username, password=token.password, pin=token.pin
         )
+
+    def get_authorize_url(self) -> str:
+        """Return the OAuth authorize URL for the user to open in a browser.
+
+        Only implemented for regions that require browser-based authentication
+        (currently New Zealand). Raises NotImplementedError for all others.
+        """
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not support browser-based authentication"
+        )
+
+    def login_with_auth_code(
+        self,
+        auth_code: str,
+        username: str,
+        password: str,
+        pin: str | None = None,
+    ) -> Token:
+        """Exchange a browser-obtained OAuth auth code for tokens.
+
+        Only implemented for regions that require browser-based authentication
+        (currently New Zealand). Raises NotImplementedError for all others.
+        """
+        raise NotImplementedError(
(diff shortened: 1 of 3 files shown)


hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -335,3 +335,29 @@ def refresh_access_token(self, token: Token) -> Token | OTPRequest:
...
+        """Return the OAuth authorize URL for the user to open in a browser.
+
+        Only implemented for regions that require browser-based authentication
+        (currently New Zealand). Raises NotImplementedError for all others.
+        """
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not support browser-based authentication"
+        )
+
+    def login_with_auth_code(
```
cdnninja, line 349: I am wondering if you found a better way to do the exact same thing EU has currently.  Right now a script is used to get through a captcha on that one and grab the token and pass it in.

cdnninja:
Any chance the flow over in https://github.com/TMA84/bluelink-refresh-token also works for new zealand?

cdnninja:
A new eu flow was just put in place. Could you check if the similar endpoints work for this?

thesquib:
Thanks I'll take a look as soon as I can (at both your comments above)

cdnninja:
@thesquib  Any thoughts on this?   Should I close the PR?

PR #1077, opened 2026-04-05
security: remove plaintext password and PIN from Token serialization

AidanDalyAus:
## Security Fix

Addresses [GHSA-49vg-vvjg-5mr9](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/security/advisories/GHSA-49vg-vvjg-5mr9)

### Problem

The `Token` dataclass serializes the user's plaintext Hyundai/Kia password and vehicle PIN via `to_dict()`. This exposes credentials through Home Assistant backups, debug logs, and diagnostics exports — affecting 10,000+ installations.

### Fix

Minimal change — 3 modifications to Token.py:

1. `password` and `pin` fields get `field(default=None, repr=False)` to hide from repr output
2. `to_dict()` calls `data.pop("password")` and `data.pop("pin")` before returning
3. `from_dict()` no longer loads password or pin from serialized state

### What's preserved

- All 8 original fields remain on the dataclass
- `token.password` and `token.pin` are fully accessible at runtime for login and session refresh
- `device_id`, `valid_until`, `stamp` unchanged
- Constructor signature unchanged
- Backward compatible with existing serialized data

### Tests

34 verification tests passed covering constructor compatibility, serialization security, round-trip data integrity, repr safety, and existing code patterns.

### DIFF
--- a/hyundai_kia_connect_api/Token.py
+++ b/hyundai_kia_connect_api/Token.py
@@ -3,26 +3,38 @@
 # pylint:disable=invalid-name
 
 import datetime as dt
-from dataclasses import dataclass, asdict
+from dataclasses import dataclass, asdict, field
 
 
 @dataclass
 class Token:
-    """Token"""
+    """Token
+
+    Security note: password and pin are excluded from to_dict() serialization
+    to prevent credential leakage through HA backups, debug logs, and diagnostics.
+    """
 
     username: str = None
-    password: str = None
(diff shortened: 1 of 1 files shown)


cdnninja:
This would break home assistant.  If the pin changes it can't be set anymore.

cdnninja:
Closing due to no response.

PR #1080, opened 2026-04-07
feat(usa): read charge targets via dedicated /evc/gts endpoint

kingpanther13:
## Summary

The `cmm/gvi` and `rems/rvs` endpoints no longer return `targetSOC` for some USA vehicles (confirmed on 2020 Kia Niro EV — the field is simply absent from `evStatus` in both responses). The Kia Connect mobile app still displays charge limits correctly because it uses a dedicated `/evc/gts` endpoint that the Python library has never called.

## The endpoint

`GET /apigw/v1/evc/gts` — returns the current AC and DC charge target percentages:

```json
{
  "status": {"statusCode": 0, "errorMessage": "Success with response body"},
  "payload": {
    "targetSOClist": [
      {"plugType": 0, "targetSOClevel": 100},
      {"plugType": 1, "targetSOClevel": 80}
    ]
  }
}
```

This was discovered by decompiling the Kia Connect APK (`com.myuvo.link`) and finding mock response files for `/evc/gts`. Tested live against `api.owners.kia.com` — returns correct values reliably, both with and without a preceding force refresh.

## Changes

- **`KiaUvoApiUSA._get_charge_targets()`**: New method that calls `GET /evc/gts` and sets `vehicle.ev_charge_limits_ac` and `vehicle.ev_charge_limits_dc`
- **`update_vehicle_with_cached_state()`**: Now calls `_get_charge_targets()` after parsing the cached response, so charge limits are available on every update cycle
- **Force refresh path**: `_get_charge_targets()` is called via `update_vehicle_with_cached_state()`, plus the existing `_update_charge_limits_from_force_refresh()` fallback is preserved for vehicles/regions where `rems/rvs` still returns targetSOC
- **7 new tests**: success, overwrite stale values, error response, empty list, zero values ignored, network exception, bool rejection

## Context

- This is a follow-up to #1059 and #1024 which added targetSOC parsing from `rems/rvs` as a workaround
- The `rems/rvs` workaround stopped working when Kia's server stopped including `targetSOC` in the response
- The `/evc/gts` endpoint is the proper way to read charge targets — it's what the official Kia Connect app uses
- No integration-side changes needed — the integration already reads `vehicle.ev_charge_limits_ac/dc`

## Other /evc/ endpoints found in the APK teardown

For reference, these were also found but not implemented in this PR:
- `/evc/gv2ls` — Get V2L status (returns "not supported" on 2020 Niro EV)
- `/evc/gcs` — Get charging stations (needs location params)
- `/evc/gcsb` — Get charging station booking
- `/evc/gcsi` — Get charging station info
- `/evc/gch` — Get charging history

## Test plan

- [x] 25 tests pass (18 existing + 7 new)
- [x] Live tested against `api.owners.kia.com` with 2020 Kia Niro EV — returns DC=100, AC=80 correctly
- [ ] Needs testing on other Kia/Hyundai USA EVs to confirm broad compatibility

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiUSA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiUSA.py
@@ -416,7 +416,12 @@ class KiaUvoApiUSA(ApiImpl):
     def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> None:
         state = self._get_cached_vehicle_state(token, vehicle)
         self._update_vehicle_properties(vehicle, state)
-        self._get_charge_targets(token, vehicle)
+        # Only EV/PHEV vehicles have charge targets; skip the /evc/gts call
+        # for ICE vehicles. ev_battery_percentage is set by
+        # _update_vehicle_properties when the cached response includes
+        # evStatus.batteryStatus, which is EV/PHEV-only.
+        if vehicle.ev_battery_percentage is not None:
+            self._get_charge_targets(token, vehicle)
 
     def force_refresh_vehicle_state(self, token: Token, vehicle: Vehicle) -> None:
         state = self._get_forced_vehicle_state(token, vehicle)
@@ -863,6 +868,15 @@ class KiaUvoApiUSA(ApiImpl):
         The cmm/gvi and rems/rvs endpoints do not return targetSOC for some
         vehicles (e.g. 2020 Kia Niro EV).  The /evc/gts endpoint reliably
         returns the current AC and DC charge target percentages.
+
+        Note: On a freshly authenticated session, the first call to this
+        endpoint may return targetSOClevel=0 for both plug types until the
+        server populates the cache (a force refresh or a few minutes of
+        session activity is usually enough). We treat 0 as "no data" and
+        preserve any previously cached values rather than overwriting with
+        a bogus 0.  Valid charge limits are 50-100 per the chargeFeature
+        metadata (minTargetSOC=50, maxTargetSOC=100), so 0 is never a
+        legitimate value.
         """
         url = self.API_URL + "evc/gts"
         try:
@@ -887,6 +901,13 @@ class KiaUvoApiUSA(ApiImpl):
                     continue
                 level = int(level)
                 if level <= 0:
+                    # Skip zero/negative values - typically returned on fresh
+                    # sessions before server-side cache populates. Preserve
+                    # any existing cached values instead.
+                    _LOGGER.debug(
+                        f"{DOMAIN} - /evc/gts returned level={level} for "
+                        f"plugType={plug_type}, preserving cached value"
+                    )
                     continue
                 if plug_type == 1:
                     vehicle.ev_charge_limits_ac = level
--- a/hyundai_kia_connect_api/KiaUvoApiUSA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiUSA.py
@@ -876,9 +876,7 @@ class KiaUvoApiUSA(ApiImpl):
                     f"{response_json['status']['errorMessage']}"
                 )
                 return
-            target_soc_list = response_json.get("payload", {}).get(
-                "targetSOClist"
-            )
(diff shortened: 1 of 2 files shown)


kingpanther13:
@cdnninja I spent way more hours figuring this out than I'd care to admit haha, but it is working great right now. I'm testing it now on my live HA instance and it works perfectly, no changes needed to the integration other than supporting the proper API version. I'm going to make the PR on the integration side for making auto -PRs whenever the API is updated.

I went thru a lot of steps trying to ping the API directly, then by trying to sniff traffic, then trying to emulate an android device and nothing worked... Then I decompiled the apk I found a bunch of non-encrypted API points in their demo files and that's how I tracked this down.

hyundai_kia_connect_api/KiaUvoApiUSA.py
```diff
@@ -416,6 +416,7 @@ def refresh_vehicles(
     def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> None:
         state = self._get_cached_vehicle_state(token, vehicle)
         self._update_vehicle_properties(vehicle, state)
+        self._get_charge_targets(token, vehicle)
```
cdnninja: Lets only call this for EV car type.

cdnninja (changes requested):

hyundai_kia_connect_api/KiaUvoApiUSA.py
kingpanther13: Done in 2fb0f94. `_get_charge_targets` is now only called when `vehicle.ev_battery_percentage is not None`, matching the same EV check used on the integration side in kia_uvo#1525 for entity creation.

hyundai_kia_connect_api/KiaUvoApiUSA.py
```diff
@@ -416,6 +416,12 @@ def refresh_vehicles(
     def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> None:
         state = self._get_cached_vehicle_state(token, vehicle)
         self._update_vehicle_properties(vehicle, state)
+        # Only EV/PHEV vehicles have charge targets; skip the /evc/gts call
+        # for ICE vehicles. ev_battery_percentage is set by
+        # _update_vehicle_properties when the cached response includes
+        # evStatus.batteryStatus, which is EV/PHEV-only.
+        if vehicle.ev_battery_percentage is not None:
```
cdnninja, line 423: I know in the one spot done this way but thinking checking engine type is preferred?  Thoughts?

kingpanther13: Good point — I looked into it and `KiaUvoApiUSA` does not currently populate `engine_type` anywhere (unlike `HyundaiBlueLinkApiUSA` which sets it during `get_vehicles` from the `evStatus` field). So `vehicle.engine_type` is always `None` for Kia USA vehicles today, which is why I used the `ev_battery_percentage` check (same approach taken in kia_uvo#1525 for entity creation — precisely because `engine_type` was unreliable here).

Two options I see:

**A.** Keep `ev_battery_percentage is not None` for now (smallest diff, unblocks this PR) and open a separate PR to populate `engine_type` properly in `KiaUvoApiUSA.get_vehicles` / `_update_vehicle_properties`, then switch the check across `KiaUvoApiUSA` in a follow-up.

**B.** Expand this PR to also populate `engine_type` in `KiaUvoApiUSA` (either from `ownr/gvl` response `fuelType` — would need to confirm the integer mapping — or from presence of `evStatus` in the cached response), and use `vehicle.engine_type == ENGINE_TYPES.EV` here.

I lean toward **A** to keep this PR focused, but happy to do **B** if you prefer. Which would you like?

cdnninja: Could you drop a PR for B? I can merge this as in the meantime.    Other items depend on engine_type so that may solve other problems and really is the preferred route.

PR #1082, opened 2026-04-11
fix: remove debug logging of tokens and cookies

Lekensteyn:
Access tokens are typically short-lived. Refresh tokens are more
dangerous to leak, they are like passwords that are valid for a long
time. Remove any logging that could possibly leak these.


cdnninja:
Great thank you!

PR #1085, opened 2026-04-13
feat(usa): populate engine_type and use it to gate /evc/gts

kingpanther13:
## Summary

Follow-up to #1080 per @cdnninja's review suggestion. `KiaUvoApiUSA` was the
only region-specific implementation that never populated `Vehicle.engine_type`,
so #1080 had to gate the new `/evc/gts` call on `ev_battery_percentage is not
None` as a proxy. This PR populates `engine_type` properly and switches the
gate to use it.

## Changes

1. **`get_vehicles()`** now maps `ownr/gvl`'s `fuelType` field. Only
   `fuelType=4` is confirmed from a live account (2020 Niro EV), so that's the
   only value mapped to `ENGINE_TYPES.EV`; anything else is left as `None` and
   refined in step 2.
2. **`_update_vehicle_properties()`** refines `engine_type` from the cached
   `cmm/gvi` state when it is still `None`:
   - `evStatus` present + `drvDistance.rangeByFuel.gasModeRange` → `PHEV`
   - `evStatus` present, no `gasModeRange` → `EV`
   - no `evStatus` → `ICE`
3. **Gate in `update_vehicle_with_cached_state()`** switched from
   `ev_battery_percentage is not None` to
   `engine_type in (ENGINE_TYPES.EV, ENGINE_TYPES.PHEV)`.

## Testing caveat

> ⚠️ **I only have access to a 2020 Kia Niro EV account**, so I cannot
> verify the ICE, PHEV, or HEV code paths against real API responses.
> Those paths are exercised by unit tests based on the expected cached
> response shape, but the `fuelType` integer values for non-EV Kia USA
> vehicles are unknown to me — that's why only `fuelType=4` is mapped
> explicitly and everything else falls through to the cached-state
> inference.

If anyone with an ICE / PHEV / HEV Kia USA account sees issues I'm happy
to adjust.

## Tests

9 new tests in `tests/us_target_soc_test.py`:

- `test_engine_type_from_fuel_type_ev` — `fuelType=4` → `EV`
- `test_engine_type_from_fuel_type_unknown` — other values → `None`
- `test_update_vehicle_properties_infers_ev_from_ev_status`
- `test_update_vehicle_properties_infers_ice_without_ev_status`
- `test_update_vehicle_properties_infers_phev_from_gas_mode_range`
- `test_update_vehicle_properties_preserves_preset_engine_type`
- `test_update_cached_state_calls_evc_gts_for_phev_vehicle`
- (plus the two existing gate tests rewritten to use `engine_type`)

Full suite: **96 passed, 8 skipped**, `ruff check` clean.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiUSA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiUSA.py
@@ -21,6 +21,7 @@ from .Vehicle import Vehicle
 from .const import (
     DISTANCE_UNITS,
     DOMAIN,
+    ENGINE_TYPES,
     LOGIN_TOKEN_LIFETIME,
     ORDER_STATUS,
     TEMPERATURE_UNITS,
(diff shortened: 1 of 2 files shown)


cdnninja:
Thank you!   Moving the kia USA api in the right direction!

kingpanther13:
Glad to help! Let me know if I need to do any followup PRs or anything regarding it, once we can confirm some of the fueltype ID numbers I can get rid of the fallback stuff that is currently set up.

cdnninja:
Sounds good.  Support with region specific items is always appreciated as considering I don't have these cars difficult to really confirm anything.

PR #1102, opened 2026-04-21
fix: After forced refresh, also call refresh from server cache

jablo:
Tested on EU Hyundai Ioniq 5 (2025) and Kia EV3 (2025).

This fixes the underlying issue seen in Hyundai-Kia-Connect/kia_uvo#1617

The response to force_refresh does not include the SOC of my cars, and it makes the call fail and the SOC entity unavailable.

This fix makes a refresh from cache call after the force refresh call to get the correct values.

Related to:
- https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/1629

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -296,6 +296,9 @@ class KiaUvoApiEU(ApiImplType1):
                 )
             else:
                 self._update_vehicle_drive_info(vehicle, state)
+        # Forced update response is incomplete (SOC levels aren't included). Instead update
+        # from cached state.
+        self.update_vehicle_with_cached_state(token, vehicle)
 
     def _update_vehicle_properties(self, vehicle: Vehicle, state: dict) -> None:
         if get_child_value(state, "vehicleStatus.time"):


cdnninja:
Reason I haven't merged this I think this occurs in the kia_uvo repo in certain spots.    So I think we need coordinated effort.  Going to comment on the issue report now on some other thoughts too.

cdnninja:
Looked more at that issue - seems some cars no longer include the data in the payload.   They used to.   So we need to find a way to detect this.  The current PR still drops all the values before it pulls cached which isn't ideal.

hyundai_kia_connect_api/KiaUvoApiEU.py
cdnninja, line 277: This and location should have an if statement.  Thinking its based on car generation / type?  In some cases the data should just get used, in others cached update will be needed.

cdnninja (changes requested):

hyundai_kia_connect_api/KiaUvoApiEU.py
jablo: I am quite challenged at the moment, it seems that whenever I have the time to address these things, neither my KIA app nor my Hyundai App works. Both currently half-update some things (eg. location, but not state of air-condition systems), and it changes constantly. Yesterday my Hyundai App reported "server issues" for several hours. 

Until the apps stabilize I cannot really trust whatever my Home Assistant is doing.

You **may** want to look at my fork, which I had been running successfully for over a week until the native app issues started intermittently. But then again, it's work-in-progress and I should rebase and rework the pull requests anyway.

A very annoying situation

jablo:
Superseded by other PRs

PR #1118, opened 2026-05-03
feat(Europe):  Username/password support

cdnninja:
Draft PR to add support for this work.  https://github.com/TMA84/bluelink-refresh-token/issues/22

Credit to [bluelink-refresh-token](https://github.com/TMA84/bluelink-refresh-token) for the login logic.


Lekensteyn:
I am somewhat concerned about this PR:
* It adds a dependency on external project for critical authentication functionality. This introduces supply chain security risks, something that I have been trying to reduce in previous PRs.
* It removes the ability to use a refresh token, forcing storage of a password again. Probably not that big of a deal though given that there is currently no way to revoke a refresh token anyway. The vendor will probably not be very pleased with the amount of refresh tokens that are needlessly generated by each user, and might start locking things down to reduce the load.

The linked project currently has a lot of cruft, probably due to heavy use of AI. That makes code review harder. Looking for the actual core functionality, there are also questionable practices such as logging the password length: https://github.com/TMA84/bluelink-refresh-token/blob/d71bb181cc9791422d883a1303a5716dadc3f632/standalone/web.py#L1061

I am quite surprised this works as-is. In captures from November 2025, the `https://idpconnect-eu.hyundai.com/auth/account/signin` endpoint contained a `recaptchaResponse` field in the client request, I thought it was mandatory. Wouldn't be surprised if this breaks over time.

cdnninja:
Fair on the dependency.   My goal is to get back to typical Home Assistant setup where it works natively.   HA stores passwords for hundreds if not thousands of integrations, that isn't new.    Right now many users are using the linked repo as other methods of getting tokens don't seem to work.   

The other option is taking the "headless_login" method and integrating here natively, with giving credit to them for finding this.

Lekensteyn:
@blka started integrating the auth module based on this PR, see #1121. Shall we close this PR in favor of that one?

cdnninja:
Planning to use https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1123

PR #871, opened 2025-08-24
fix: Patched the login process for Kia in Europe.

marvinwankersteen:
Patched the login process for Kia in Europe.

PR #1123, opened 2026-05-06
feat(Europe): Integrate headless login into KiaUvoApiEU, remove curl_cffi

blka:
## Summary

Integrates the headless EU login (username + password) directly into `KiaUvoApiEU` as `_login_with_password()`, eliminating the separate `headless_login.py` module and the `curl_cffi` binary dependency.

Key changes:
- **Removed `curl_cffi` dependency** — the IDPConnect auth servers accept standard `requests` with the Android User-Agent header (`_CCS_APP_AOS` suffix). No TLS fingerprinting is needed. Verified live with Hyundai EU credentials.
- **Removed `headless_login.py`** — the RSA password encryption flow is now a method on `KiaUvoApiEU`, reusing existing class attributes (`CCSP_SERVICE_ID`, `CCS_SERVICE_SECRET`, `LOGIN_FORM_HOST`, `USER_API_URL`, etc.)
- **`pycryptodome` is now a required dependency** — added to `requirements.txt`. It's needed for RSA password encryption in the IDPConnect OAuth flow.
- **Reuses existing class attributes** — no duplicate OAuth client IDs or secrets; the method reads `self.CCSP_SERVICE_ID`, `self.CCS_SERVICE_SECRET`, etc.
- **Brand-specific redirect URIs** extracted into `self._oauth_redirect_uri` in `__init__`, eliminating duplication.
- **`refresh_access_token` override** — uses the stored refresh token to get a new access token via a single POST to `/oauth2/token` (~200ms), instead of re-doing the full OAuth2 login flow (~3s). Falls back to full login if the refresh token is expired.
- **Removed dead code** — deleted unused `_get_authorization_code_with_redirect_url`, `_get_authorization_code_with_form`, `_get_refresh_token`, and `LOGIN_FORM_URL` that were superseded by the headless login.
- **Improved error differentiation** — user-caused errors (wrong credentials) now include "Check username and password"; API/internal errors include "This may indicate a Hyundai API change" to help distinguish configuration issues from upstream changes.

## Addresses review feedback

- **Lekensteyn**: "I really hope we are not creating a new refresh token every time" → Added `refresh_access_token` override that uses the stored refresh token via `_get_access_token`, falling back to full login only if the refresh token is expired.
- **Lekensteyn**: "We set these redirect URLs in `__init__` and now duplicate them here again?" → Extracted redirect URI into `self._oauth_redirect_uri` in `__init__`, eliminating the duplication.
- **Lekensteyn**: "`_get_authorization_code_with_redirect_url` shares a great deal of code with this newly added code" → Removed the dead code (old browser-based login methods).
- **Lekensteyn**: "`CCS_APP_` substring is the core requirement to prevent a 400 Bad Request" → Added inline comment explaining this.
- **Lekensteyn**: "Need to clearly differentiate between user errors and internal errors" → Improved error messages with user/API error distinction.
- **cdnninja**: "Is this a page they need to accept somewhere? If so lets make this a new exception" → Already handled via `ConsentRequiredError` exception.
- **juodikis**: Genesis EU support — added correct `CCS_SERVICE_SECRET` and redirect URI.

## Live verification

Both login flows tested successfully with real Hyundai EU (Santa Fe) credentials:
- ✅ Password login via `_login_with_password()` — returns valid access_token and refresh_token
- ✅ Refresh token login via `_get_access_token()` — unchanged, works as before
- ✅ Token refresh via `refresh_access_token()` — uses stored refresh_token, no full re-login

## Test plan

- [x] `python3 -m pytest tests/test_headless_login.py -v` — 12 tests pass
- [x] `ruff check` + `ruff format` — clean
- [x] Live test with Hyundai EU credentials (both password and refresh_token flows)
- [ ] Genesis EU — needs a Genesis EU user to test (@juodikis?)


juodikis:
Only two minor changes and it works perfectly.

`hyundai_kia_connect_api/KiaUvoApiEU.py`:

Change on line 126:
```self.CCS_SERVICE_SECRET: str = "FKDdlef2ffdleFEweELFKERiLER2FED21sDdwdgQz6hFESE3"```

Add at line 244
```elif BRANDS[self.brand] == BRAND_GENESIS:```
```    redirect_uri = "https://accounts-eu.genesis.com/realms/eugenesisidm/ga-api/redirect2"```

**Comment by blka** [comment]
@juodikis Thanks for the review! Both changes implemented:

1. **Genesis `CCS_SERVICE_SECRET`** — replaced the `"secret"` placeholder with the actual value (`FKDdlef2ffdleFEweELFKERiLER2FED21sDdwdgQz6hFESE3`), which matches what's encoded in `BASIC_AUTHORIZATION` (base64 decode confirms: `3020afa2-...:FKDdlef2ffdleFEweELFKERiLER2FED21sDdwdgQz6hFESE3`).

2. **Genesis `redirect_uri`** — added explicit `BRAND_GENESIS` branch that uses `https://accounts-eu.genesis.com/realms/eugenesisidm/ga-api/redirect2`, instead of the generic PORT-443 path.

File: hyundai_kia_connect_api/KiaUvoApiEU.py [hunk 1]
```diff
@@ -181,26 +184,33 @@ def login(
...
+        # Determine if password is a refresh_token or plaintext credentials
+        is_refresh_token = bool(re.match(r"^[A-Z0-9]{48}$", password))
+
+        if is_refresh_token:
+            # Existing flow: use the 48-char refresh_token directly
+            refresh_token = password
+            _, access_token, _, expires_in = self._get_access_token(
+                stamp, refresh_token
+            )
+        elif self.brand in (1, 2, 3):  # Kia=1, Hyundai=2, Genesis=3
```
cdnninja: Could we switch this to using the ENUM?

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -181,26 +184,33 @@ def login(
...
+            # Headless login for Kia/Hyundai/Genesis EU: username + plaintext password
+            access_token, refresh_token, expires_in = self._login_with_password(
+                username, password
+            )
+        else:
             raise AuthenticationError(
-                "Passwords are no longer supported, provide a refresh_token instead"
+                "Username/password login is only supported for "
+                "Kia, Hyundai, and Genesis (EU). Provide a refresh_token instead."
```
cdnninja: Doesn't this support all 3?

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -215,6 +225,136 @@ def login(
...
+        code_list = parse_qs(urlparse(location).query).get("code")
+        if not code_list:
+            if "error" in location.lower():
+                error_desc = parse_qs(urlparse(location).query).get(
+                    "error_description", ["unknown"]
+                )[0]
+                raise AuthenticationError(f"Signin rejected: {error_desc}")
+            if "/web/v1/user/authorization" in location:
+                raise AuthenticationError(
+                    "Signin succeeded but requires a consent page "
```
cdnninja: Is this a page they need to accept somewhere?  If so lets make this a new exception so we can handle this well in HA with a translated message.

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -215,6 +225,136 @@ def login(
...
+                    "(SPA redirect). Try using a refresh token instead."
+                )
+            if "authorize" in location:
+                raise AuthenticationError(
+                    "Signin failed — redirected back to login page. "
```
cdnninja: I don't think we are redirecting anywhere.

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -215,6 +225,136 @@ def login(
             pin=pin,
         )
 
+    def _login_with_password(self, username, password):
```
cdnninja: This should define the return types, and data types.

setup.py
```diff
@@ -26,13 +26,17 @@
     author_email="fuatakgun@gmail.com",
     python_requires=">=3.10",
     classifiers=[
-        "Development Status :: 2 - Pre-Alpha",
+        "Development Status :: 4 - Beta",
```
cdnninja, line 29: This could probably move to stable but not related to this PR.

cdnninja (changes requested):

cdnninja:
Anyone for final feedback before I merge?

fuatakgun:
> Anyone for final feedback before I merge? 

Looks good. Let's see, thanks

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -215,6 +221,141 @@ def login(
...
+        elif BRANDS[self.brand] == BRAND_GENESIS:
+            redirect_uri = (
+                "https://accounts-eu.genesis.com/realms/eugenesisidm/ga-api/redirect2"
+            )
+        elif self.PORT == 443:
+            redirect_uri = f"https://{self.BASE_DOMAIN}/api/v1/user/oauth2/redirect"
+        else:
+            redirect_uri = self.USER_API_URL + "oauth2/redirect"
+
+        mobile_ua = USER_AGENT_MOZILLA + "_CCS_APP_AOS"
```
Lekensteyn, line 216: Last year I confirmed that this `CCS_APP_` substring is the core requirement to prevent an "400 Bad Request" error through the redirect flow of `https://idpconnect-eu.hyundai.com/auth/api/v2/user/oauth2/authorize?...`. Might be worth a comment.

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -215,6 +221,141 @@ def login(
...
+        if BRANDS[self.brand] == BRAND_HYUNDAI:
+            redirect_uri = self.USER_API_URL + "oauth2/token"
+        elif BRANDS[self.brand] == BRAND_GENESIS:
+            redirect_uri = (
+                "https://accounts-eu.genesis.com/realms/eugenesisidm/ga-api/redirect2"
+            )
+        elif self.PORT == 443:
+            redirect_uri = f"https://{self.BASE_DOMAIN}/api/v1/user/oauth2/redirect"
+        else:
+            redirect_uri = self.USER_API_URL + "oauth2/redirect"
```
Lekensteyn: We set these redirect URLs in `KiaUvoAPIEU.__init__` and now duplicate them here again? Can't we reuse those?

Don't add redundant comments like "All three EU brands" are supported. And it seems rather unlikely that we need two different redirect URL flows for Kia? Based on the `__init__` code, the port is always 8080 as I would expect for the idpconnect API.

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -215,6 +221,141 @@ def login(
...
+                    "once to accept the terms, then use the refresh token."
+                )
+            if "authorize" in location:
+                raise AuthenticationError(
+                    "Signin failed — returned to login page. "
+                    "Check username and password."
+                )
+            raise AuthenticationError(
+                f"No authorization code in redirect: {location[:250]}"
+            )
```
Lekensteyn, line 296: Were you able to reproduce these flows and capture these flows? What are the full URLs observed during auth errors?

Need to clearly differentiate between user errors (password incorrect) and internal errors (Hyundai changing their APIs and the code not handling it correctly).

Lekensteyn (commented):
@cdnninja the code seems functional, so I would understand if you squash and merge the code as-is.

It unfortunately makes the code even harder to maintain since it choses to add another special case without considering/fixing the existing code.

We already have a `_get_authorization_code_with_redirect_url`, but its caller was removed in 61a5c3b635ea187a6db4fbbd55b9562e1d7351c0 by @juodikis. That function shares a great deal of code with this newly added code.

Design wise, I really hope we are not creating a new refresh token every time. Ideally we can persist the refresh token, and only use the username/password to get a new refresh token if it somehow expired or became invalid.

blka:
## Addressing review feedback

Pushed 5 commits addressing all review feedback from @Lekensteyn and @cdnninja:

### 1. Refresh token persistence (Lekensteyn's main concern)
> "I really hope we are not creating a new refresh token every time."

Added `refresh_access_token` override in `KiaUvoApiEU` that uses the stored `token.refresh_token` via `_get_access_token` — a single POST to `/oauth2/token` (~200ms) instead of re-doing the full OAuth2 login flow with device registration and cookie fetching (~3s). Falls back to full login only if the refresh token itself is expired or invalid.

This was confirmed by HA debug logs showing "Refresh token expired" → "Get Device ID request" → "Get cookies request" every 30 minutes. With this fix, token refresh no longer triggers full re-authentication.

### 2. Redirect URI duplication
> "We set these redirect URLs in `__init__` and now duplicate them here again?"

Extracted the brand-specific redirect URI into `self._oauth_redirect_uri` in `__init__`, alongside other brand configuration. `_login_with_password` now reads from this attribute instead of computing it locally.

### 3. Dead code cleanup
> "`_get_authorization_code_with_redirect_url` shares a great deal of code with this newly added code."

Removed three unused methods (`_get_authorization_code_with_redirect_url`, `_get_authorization_code_with_form`, `_get_refresh_token`) and the `LOGIN_FORM_URL` attribute that were superseded by the headless login flow. Also removed the now-unused `BeautifulSoup` import.

### 4. CCS_APP_ comment
> "`CCS_APP_` substring is the core requirement to prevent a 400 Bad Request"

Added inline comment explaining that the `_CCS_APP_AOS` suffix in the User-Agent is mandatory for the IDPConnect authorize endpoint.

### 5. Error differentiation
> "Need to clearly differentiate between user errors and internal errors"

Improved error messages: user-caused errors (wrong credentials, consent required) now include "Check username and password" guidance; API/internal errors (unexpected HTTP status, redirect structure) include "This may indicate a Hyundai API change" to help distinguish configuration issues from upstream API changes.

---

Regarding @cdnninja's question about consent/exception handling — the `ConsentRequiredError` exception is already raised when the user is redirected to the `/web/v1/user/authorization` page, which happens when an account needs to accept terms before first use.

blka:
@cdnninja @Lekensteyn @juodikis @fuatakgun — pinging everyone who commented or reviewed this PR.

I've addressed cdnninja's review feedback (rebased, enum usage, exception handling). Could you take another look and let me know if there's anything else needed?

Also — what's the typical review turnaround time for this repo? Just trying to set expectations on when these changes might land. Thanks!

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -139,68 +142,46 @@ def __init__(self, region: int, brand: int, language: str) -> None:
...
-                + auth_client_id
-                + "&scope=openid%20profile%20email%20phone&response_type=code&hkid_session_reset=true&redirect_uri="  # noqa
-                + self.USER_API_URL
-                + "integration/redirect/login&ui_locales="
-                + self.LANGUAGE
-                + "&state=$service_id:$user_id"
+            self._oauth_redirect_uri: str = (
+                "https://accounts-eu.genesis.com/realms/eugenesisidm/ga-api/redirect2"
+            )
+        elif self.PORT == 443:
```
cdnninja: Is this how we are grabbing Kia?

blka:
@cdnninja Re: "Is this how we are grabbing Kia?" (line 152)

The `state=ccsp` in `LOGIN_FORM_URL` is from the old Kia EU login flow (PR #871 by marvinwankersteen, Aug 2025). Our new headless login for Kia uses the Keycloak endpoint (`/auth/realms/eukiaidm/protocol/openid-connect/auth`) via `_login_with_password`, not the `state=ccsp` path.

The `LOGIN_FORM_URL` attribute and `_get_authorization_code_with_redirect_url` method (which uses `state=ccsp`) are dead code — they're defined but never called. I missed cleaning these up in the "remove dead code" commit. Want me to remove them in a follow-up commit on this PR?

cdnninja:
Yes please.    I noticed Genesis and Hyundai have a if statement but not Kia.

blka:
@cdnninja Correction to my previous comment — I was wrong about Kia using Keycloak.

Our `_login_with_password` uses the same IDPConnect flow for all three EU brands (Kia, Hyundai, Genesis):
1. `idpconnect-eu.{brand}.com/auth/api/v2/user/oauth2/authorize` — with `state=ccsp`
2. `auth/api/v1/accounts/certs` — RSA public key
3. `auth/account/signin` — encrypted password

So `state=ccsp` is still very much in use — it's part of the active headless login flow. The `LOGIN_FORM_URL` attribute in `__init__` is what's dead code, not the `state=ccsp` parameter itself.

The `LOGIN_FORM_URL` attribute was used by the old browser-based login flow (`_get_authorization_code_with_redirect_url` / `_get_authorization_code_with_form`). Those methods are never called anymore since our PR replaces them with `_login_with_password`. Want me to clean up that dead code?

blka:
@cdnninja Done — pushed `ddc1f4b`. Added explicit `BRAND_KIA` branch for `_oauth_redirect_uri` and removed the fallback `else` / `PORT == 443` logic. Now all three brands have their own explicit branch.

The dead code (`LOGIN_FORM_URL`, `_get_authorization_code_with_redirect_url`, `_get_authorization_code_with_form`) was already removed in an earlier commit on this branch.

PR #1128, opened 2026-05-12
feat: add supports_window_control capability flag

blka:
## Summary

- Add `supports_window_control: bool` class attribute to API implementations so the Home Assistant integration can check whether window control is supported before creating entities
- `ApiImpl` base class defaults to `False`; `ApiImplType1` (EU/CN/AU/IN) and `HyundaiBlueLinkApiBR` override to `True`
- USA and CA API classes inherit `False` from `ApiImpl`

## Problem

Issue #1655 in the kia_uvo integration: window cover entities appear for USA/CA vehicles because those APIs populate `front_left_window_is_open` etc. from vehicle status data (read-only), but `set_windows_state` raises `NotImplementedError` when users try to control them.

## Solution

A capability flag that the integration can check before creating window entities. Regions that implement `set_windows_state` (EU, CN, AU, IN, BR) set it to `True`; regions that don't (USA, CA) inherit `False`.

## Companion PR

Hyundai-Kia-Connect/kia_uvo#xxx — uses this flag to skip cover entity creation for unsupported regions

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -89,6 +89,7 @@ class ApiImpl:
     temperature_range = None
     previous_latitude: float = None
     previous_longitude: float = None
+    supports_window_control: bool = False
 
     def __init__(self) -> None:
         """Initialize."""
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -156,7 +156,7 @@ class ApiImplType1(ApiImpl):
                             "This appears to be corrupted data from Hyundai's API."
                         )
                         dte["ICE"] = None
-        except KeyError, TypeError, AttributeError:
+        except (KeyError, TypeError, AttributeError):  # fmt: skip
             # If the structure doesn't exist or is malformed, silently continue
             # This is defensive programming in case the API structure changes
             pass
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -113,6 +113,8 @@ def _check_response_for_errors(response: dict) -> None:
 class ApiImplType1(ApiImpl):
     """ApiImplType1"""
 
+    supports_window_control: bool = True
+
     def __init__(self) -> None:
         """Initialize."""
 
@@ -154,7 +156,7 @@ class ApiImplType1(ApiImpl):
                             "This appears to be corrupted data from Hyundai's API."
                         )
                         dte["ICE"] = None
-        except (KeyError, TypeError, AttributeError):
+        except KeyError, TypeError, AttributeError:
             # If the structure doesn't exist or is malformed, silently continue
             # This is defensive programming in case the API structure changes
             pass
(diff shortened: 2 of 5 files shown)


cdnninja:
Should this be on the vehicle? As I assume not all cars support it within a region.

blka:
Good point. Ideally this would be per-vehicle, but the API doesn't expose a "supports window control" capability per vehicle.

The problem this solves: USA/CA vehicles populate `front_left_window_is_open` etc. from vehicle status (read-only), so cover entities get created even though `set_windows_state` raises `NotImplementedError` for those regions. A per-vehicle flag would be more precise, but we'd need the API to tell us which vehicles support window control — and it doesn't.

The current check works as a safety net:
- **Region flag `False`** (USA, CA) → no window entities at all (correct, since `set_windows_state` is not implemented)
- **Region flag `True`** (EU, CN, AU, IN, BR) → entities appear only if the vehicle reports window state (the existing `getattr` check in cover.py already handles per-vehicle variation)

So in practice, per-vehicle filtering already happens via `getattr(vehicle, description.key, None) is not None` — only vehicles that report window state get entities. The region flag just prevents creating entities that would throw on use.

If in the future the API adds a per-vehicle capability indicator for window control, we can add that check too.

blka:
I looked into whether the API exposes per-vehicle capability data for window control — it doesn't.

For EU/CN/AU/IN (ApiImplType1), we have `ccuCCS2ProtocolSupport` per-vehicle, but that's specific to the CCS2 data protocol, not window control capability. The `set_windows_state` method exists unconditionally for all vehicles in these regions.

For USA/CA, there's no equivalent per-vehicle capability flag at all. The USA API also has `vehicleFeature: "0"` (disabled) in its request body for `cmm/gvi` — meaning it doesn't even ask the server for feature data.

So the region-level flag works as a safety net:
- **Region `True`** (EU, CN, AU, IN, BR): entities appear only if the vehicle reports window state (`getattr` check in cover.py already filters per-vehicle)
- **Region `False`** (USA, CA): no window entities at all, since `set_windows_state` raises `NotImplementedError`

One idea: we could try setting `vehicleFeature: "1"` in the USA API request to see if the server returns per-vehicle feature/capability data — but that's a separate investigation beyond this PR.

@cdnninja — do you think the region-level approach is acceptable for now, or would you prefer we investigate `vehicleFeature` first?

blka:
@cdnninja — following up on your question about per-vehicle vs region-level. As discussed in the previous comment, the API doesn't expose per-vehicle window control capability — the region-level flag is the best we can do right now, with per-vehicle filtering already working via the existing `getattr` check.

Is there anything else you'd like changed, or is this approach acceptable? Also curious about typical review turnaround times here. Thanks!

cdnninja:
I suspect will need to move to vehicle later.  Even though yes api doesn't state it we know not all years and models support it.   So by having on the vehicle we can manage that later.  Even if for now is the same for the api.

blka:
@cdnninja Good point — pushed `1220bce`. Moved `supports_window_control` to the `Vehicle` class:

- Added `supports_window_control: bool = None` to `Vehicle`
- Set it from `api.supports_window_control` in `VehicleManager.initialize_vehicles()`
- The HA integration now checks `vehicle.supports_window_control` instead of `api.supports_window_control`

For now the value is the same for all vehicles within a region, but this makes it possible to override per-vehicle later (e.g. based on year/model).

The `supports_window_control` flag on `ApiImpl` is still used as the source of truth for what each region API supports — it's just read at vehicle init time instead of being checked directly at entity creation time.

PR #1139, opened 2026-05-18
fix: add check_action_status to HyundaiBlueLinkApiUSA

blka:
## Summary
- Fixes #1061 — Hyundai USA BlueLink API does not wait for response completion after sending commands
- Adds `check_action_status` method that polls `ac/v2/rmt/getRunningStatus` endpoint (based on [reference implementation](https://github.com/andyfase/egmp-bluelink-scriptable/blob/07d49ee19fc29cd3e1dd7f17e52526943c9fe900/src/lib/bluelink-regions/usa.ts#L272-L321))
- Updates all command methods to return the transaction ID from response headers, enabling callers to poll for completion

## Changes
1. **`check_action_status` method**: Polls `ac/v2/rmt/getRunningStatus` with the transaction ID from command response headers. Supports synchronous (polling loop with timeout) and non-synchronous (single check) modes. Returns proper `ORDER_STATUS` enum values.

2. **`_get_transaction_id` helper**: Extracts transaction ID from response headers. Checks `tmsTid` (from reference TypeScript implementation), `transactionId` (from commented code), and `Xid` (from Kia USA API) as fallbacks. Logs warning if no transaction ID found.

3. **Command methods updated**: `lock_action`, `start_climate`, `stop_climate`, `start_charge`, `stop_charge`, `set_charge_limits` now return the transaction ID string instead of `None`, aligning with KiaUvoApiUSA which returns `response.headers["Xid"]`.

## Test plan
- [x] 10 new unit tests for `check_action_status` and `_get_transaction_id`
- [x] All 162 existing tests pass
- [x] Existing USA command tests updated with proper headers mocking
- [ ] Verify transaction ID header name against real Hyundai USA API (currently checking `tmsTid`, `transactionId`, `Xid` as fallbacks)

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
@@ -18,6 +18,7 @@ from .const import (
     DISTANCE_UNITS,
     DOMAIN,
     ENGINE_TYPES,
+    ORDER_STATUS,
     SEAT_STATUS,
     TEMPERATURE_UNITS,
     VEHICLE_LOCK_ACTION,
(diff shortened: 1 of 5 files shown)


cdnninja:
Thank you for this!

PR #1147, opened 2026-05-25
fix: correct last_updated_at time being one hour off in summer

### LINKED ISSUE #1096: EU update timestamp is 1 hour behind

Lekensteyn:
The last_updated_at times for Kona EU (CCS2) were off by one in summer. A previous fix for AU (#770) broke summer time transitions.

API response for Mon, 25 May 2026 01:17:34 +0200 (CEST):

    "lastUpdateTime": "1779664654458",
    "state": {
      "Vehicle": {
        "Date": "20260524231733.000",
        "Offset": "1",
    }

API response for Sat, 21 Mar 2026 17:02:07 +0100 (CET):

    "lastUpdateTime": "1774108927875",
    "state": {
      "Vehicle": {
        "Date": "20260321160206.000",
        "Offset": "1",
    }

Fixes #1096

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -5,7 +5,6 @@ import requests
 import logging
 import math
 from typing import Optional
-from datetime import timedelta, timezone
 
 from time import sleep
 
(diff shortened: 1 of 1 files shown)


blka:
Confirmed — if no users from WET/EET countries (Portugal, Finland, Greece, etc.), the hardcoded `Europe/Berlin` + this fix is sufficient. Closing #1159 since per-country timezone detection is unnecessary without real affected users.

Issue #1153, opened 2026-05-26
EU: check_action_status stays PENDING even after command completes on vehicle


blka:
## Problem

After sending a vehicle command (lock, climate, etc.), `check_action_status` polls `GET /api/v1/spa/notifications/{vehicleId}/records` to track completion. In ~31% of cases (4/13 in testing), the endpoint finds the action record but `result` stays `None` (mapped to `ORDER_STATUS.PENDING`) for the entire polling window — even though the vehicle has already executed the command.

This is a **different issue from #1143**: #1143 describes empty `resMsg` (device_id mismatch preventing record lookup). Here, the record IS found (`recordId` matches), but the server never updates `result` from `None` to `"success"`.

## Evidence

20-minute stress test on 2026-05-26 (EU Hyundai SANTA FE):
- 13 lock commands with action status polling (10 polls × 5s = 50s window)
- 9/13: PENDING → SUCCESS (typically poll 1-2, within 10s)
- **4/13: PENDING for all 10 polls (50s), never reached SUCCESS** — despite the car actually locking

Example from test output:
```
[12:05:54] Lock OK 1392ms: actionId=7b712440...
[12:05:54]   Polling action status for 7b712440...
[12:05:54]   Poll 1/10: PENDING 277ms
[12:05:59]   Poll 2/10: PENDING 282ms
[12:06:05]   Poll 3/10: PENDING 328ms
...
[12:06:42]   Poll 10/10: PENDING 310ms
[12:06:47]   Gave up after 10 polls
```

The lock command succeeded — the car locked. But `notifications/records` never reported it.

## Impact on Home Assistant

- `VehicleManager.check_action_status()` with `synchronous=True` and `timeout=120` will poll for up to 2 minutes, blocking the executor thread
- If the server never reports SUCCESS, the method returns `ORDER_STATUS.TIMEOUT` after the polling window expires
- The HA coordinator treats this as a failed action, even though the vehicle actually performed it
- User sees "action failed" or "timeout" in the UI despite the car having responded correctly

## Root cause hypothesis

The `notifications/records` endpoint returns push notification delivery records. When the server fails to deliver the push (because our GCM registration is fake), it may not update the `result` field for that action record. This means:

- If push delivery **succeeds** (e.g., the real myHyundai app is also registered): `result` is set to `"success"`
- If push delivery **fails** (our fake GCM registration): `result` stays `None` → PENDING forever

This would explain why:
[...]

Issue #1152, opened 2026-05-26
Token.device_id race condition — no thread safety for concurrent HA calls


blka:
## Problem

`Token.device_id` is a plain string attribute mutated by `KiaUvoApiEU._get_device_id()` and `@_retry_on_device_id_error`. In Home Assistant, `DataUpdateCoordinator` may call the library from multiple executor threads concurrently (e.g., one polling `update_vehicle_with_cached_state` while another handles a user-initiated `lock_action`). This creates a race condition:

1. Thread A reads `token.device_id` for an API call
2. Thread B's `_retry_on_device_id_error` handler overwrites `token.device_id` mid-request
3. Thread A's subsequent `check_action_status` call uses the wrong device_id → empty resMsg or 4002 error

## Evidence

The `@_retry_on_device_id_error` decorator (added in PR pending) writes `token.device_id = self._get_device_id(stamp)` on DeviceIDError. If two HA calls hit DeviceIDError simultaneously, both will call `_get_device_id()` and write to `token.device_id` — the second write overwrites the first, but both threads use whatever device_id is on the token at their respective read points.

## Proposed fix

Add a `threading.Lock` to `Token` for device_id mutations:

```python
class Token:
    def __init__(self, ...):
        self._device_id_lock = threading.Lock()
        self._device_id = None

    @property
    def device_id(self):
        return self._device_id

    @device_id.setter
    def device_id(self, value):
        with self._device_id_lock:
            self._device_id = value
```

This ensures that:
- Writes to device_id are atomic — no torn reads
- Concurrent `_retry_on_device_id_error` handlers don't corrupt each other's device_id
- Reads always get a complete, valid device_id

## Alternatives considered

1. **Lock around entire API call** — too coarse-grained, serializes all API calls. The device_id is the only shared mutable state; locking just that is sufficient.
2. **Copy token per call** — would break the current pattern where `_retry_on_device_id_error` mutates the token in-place. Would require rewriting all retry logic.
3. **Immutable token** — cleanest design but requires significant refactoring of all API methods that currently mutate `token.device_id` in-place.

## Testing

- [ ] Unit test: concurrent threads writing `token.device_id` should not corrupt state
- [ ] Unit test: reads always return a valid, complete device_id string
- [ ] Integration test: two simultaneous lock commands don't interfere with each other's device_id

cdnninja:
Is this solved since you put in a PR to only allow one call at a time?

[...]

Issue #1151, opened 2026-05-26
All HTTP requests missing timeout parameter (139 calls across 10 files)


blka:
## Problem

Every `requests.get/post/put` call in the library is missing the `timeout` parameter. This means any HTTP request can hang indefinitely if the server stops responding without closing the TCP connection (CLOSE_WAIT state). The library currently suppresses this with `# pylint:disable=missing-timeout` at the top of most API files.

**Impact:** Home Assistant integrations using this library become unresponsive when the EU/US API server leaves connections in CLOSE_WAIT. The HA `DataUpdateCoordinator` calls `asyncio.to_thread()` which blocks the executor thread, preventing other integrations from updating.

**Evidence:** During integration testing on 2026-05-26, the EU API server (`prd.eu-ccapi.hyundai.com`) left a TCP connection in CLOSE_WAIT state, causing the test process to hang indefinitely until manually killed. Adding a `timeout=120` monkey-patch to `requests.Session.request` resolved the hang.

## Current state of Session usage

The library has **inconsistent HTTP transport** across regions:

| Region | Class | Parent | Session | Adapter | Created |
|--------|-------|--------|---------|---------|---------|
| EU | `KiaUvoApiEU` | `ApiImplType1` | None (bare `requests.*`) | None | N/A |
| AU | `KiaUvoApiAU` | `ApiImplType1` | None (bare `requests.*`) | None | N/A |
| IN | `KiaUvoApiIN` | `ApiImplType1` | None (bare `requests.*`) | None | N/A |
| CN | `KiaUvoApiCN` | `ApiImplType1` | None (bare `requests.*`) | None | N/A |
| CA | `KiaUvoApiCA` | `ApiImpl` | `RetrySession(requests.Session)` | None | Lazy (property) |
| USA Kia | `KiaUvoApiUSA` | `ApiImpl` | `requests.Session` | `KiaSSLAdapter` on `https://` | Lazy (property) |
| USA Hyundai | `HyundaiBlueLinkApiUSA` | `ApiImpl` | `requests.Session` | `cipherAdapter` on origin | Eager (`__init__`) |
| BR Hyundai | `HyundaiBlueLinkApiBR` | `ApiImpl` | `requests.Session` | None | Eager (`__init__`) |

**Problems:**
1. EU/AU/IN/CN use bare `requests.get/post` — no connection pooling, no timeout, new TLS handshake per call, no retry
[...]

Issue #1143, opened 2026-05-19
EU: check_action_status returns empty resMsg due to stale device_id


blka:
## Problem

After sending a vehicle command on EU, `check_action_status` polls `GET /api/v1/spa/notifications/{vehicleId}/records` to track completion. This endpoint frequently returns `resMsg: []` (empty list) because the `device_id` used in the request headers has been invalidated by the server.

This was previously reported in #424 and #564, but both were closed without a proper fix.

## Root cause

The server invalidates `device_id` after it fails to deliver push notifications to it. Our push notification registration (`notifications/register`) uses a hardcoded `pushType: "GCM"` with a static `pushRegId` hash that isn't a real Firebase Cloud Messaging token. The server tries to send a push update, fails because the registration is invalid, and invalidates the device_id.

## Current workaround

Every command method in `ApiImplType1.py` calls `token.device_id = self._get_device_id(self._get_stamp())` after the POST. This means when `check_action_status` runs immediately after a command, `token.device_id` is already fresh. This works for the normal sequential flow (command → poll), but has gaps:

1. **`check_action_status` does not call `_get_device_id` before polling** — it trusts whatever `token.device_id` contains. If called without a prior command, or after the device_id has been invalidated again, polling returns empty results.
2. **No retry logic for `DeviceIDError` (resCode 4002)** — if the server rejects the device_id during polling, it crashes rather than refreshing and retrying.
3. **During long synchronous polling** (the loop in `check_action_status` with `forced_wait=True`), the device_id could become stale mid-loop but is never refreshed.

## Proposed fix

Add device_id refresh + retry inside `check_action_status`:

1. Before each polling attempt, check if the response contains `resCode: "4002"` (DeviceIDError) or returns empty `resMsg: []` when an action is expected to be pending
2. If so, refresh `token.device_id = self._get_device_id(self._get_stamp())` and retry the GET request
3. Limit retries to avoid infinite loops

This is a client-side fix that makes the polling more robust regardless of whether the push notification registration is valid.

A more complete fix would involve proper push notification registration with valid FCM credentials, but that requires significantly more infrastructure (a backend service to receive FCM pushes).

## Related

- #424 — Invalid device id after start_charge() or stop_charge()
- #564 — "No action found with ID..." when calling check_action_status (Kia/EU)
- PR #1120 — Closed (test coverage for polling, DuplicateRequestError doesn't occur on polling endpoint)

blka:
[...]

PR #1156, opened 2026-05-26
fix: replace proactive device_id rotation with retry-on-error decorator

### LINKED ISSUE #1155: EU: Replace proactive device_id rotation with retry-on-error decorator
## Problem

EU server invalidates `device_id` when push delivery to the fake GCM token fails, causing `DeviceIDError` (resCode 4002). The current master workaround proactively rotates `device_id` after every control command (15 rotation points across `ApiImplType1.py` and `KiaUvoApiEU.py`), creating race conditions with `check_action_status` polling and concurrent API calls.

See #1143 for the original problem description and root cause analysis.

## Solution

Replace all 15 proactive rotation points with a single `@_retry_on_device_id_error` decorator on API methods:

### What changes

1. **New decorator**: `_retry_on_device_id_error` in `ApiImplType1.py` catches `DeviceIDError`, re-registers `device_id` with a `threading.Lock`, and retries the call once.

2. **Decorator applied to 17 methods** (replacing 15 proactive rotation lines):
   - `get_vehicles` (new — no proactive rotation on master)
   - `start_charge`, `stop_charge`, `set_charging_current`, `set_charge_limits`, `set_vehicle_to_load_discharge_limit`
   - `lock_action`
   - `schedule_charging_and_climate` (new — no proactive rotation on master)
   - `start_climate`, `stop_climate`
   - `start_hazard_lights`, `start_hazard_lights_and_horn`
   - `set_windows_state`, `set_navigation`
   - `charge_port_action`, `valet_mode_action` (KiaUvoApiEU overrides)
   - `update_vehicle_with_cached_state`, `force_refresh_vehicle_state` (KiaUvoApiEU overrides)

3. **All 15 `token.device_id = self._get_device_id(self._get_stamp())` lines removed** — proactive rotation eliminated.

4. **`check_action_status` does NOT get the decorator** — it returns `ORDER_STATUS` enum, not a response that could raise `DeviceIDError` directly. The polling endpoint uses a different error path (empty `resMsg`).

### What stays the same

- `_get_device_id()` method remains — used by the decorator to re-register
- `check_action_status` logic unchanged (separate fix in #1153)
- USA/CA/BR/AU/IN/CN regions unaffected — they don't use `notifications/register`

### Why this approach over #1143's proposal

#1143 proposed adding retry logic inside `check_action_status` only. The decorator approach is broader and more robust:

| Aspect | #1143 proposal | Decorator approach |
|--------|---------------|-------------------|
| Scope | `check_action_status` only | All API methods |
| Proactive rotation | Still in place | Removed entirely |
| Race conditions | Still possible | Eliminated (lock + no proactive rotation) |
| Read-only calls | Unprotected | `get_vehicles`, `update_vehicle_with_cached_state` now protected |

## Out of scope (separate issues)

- #1151 — HTTP timeout for all `requests` calls
- #1152 — Thread-safe `Token.device_id` (complements this PR — the decorator's lock only covers re-registration)
- #1153 — `check_action_status` PENDING-without-SUCCESS fix
- kia_uvo#1683 — Coordinator auto-recover from `DeviceIDError`

## Testing

Unit tests in `tests/test_device_id_retry.py` cover:
- Decorator retries once on `DeviceIDError`
- Decorator doesn't retry on other exceptions
- Decorator returns normally on success
- Lock prevents concurrent re-registration

Integration tests in `tests/integration/` for live validation.

blka:
## Summary

Replaces all 15 proactive `device_id` rotation points with a single `@_retry_on_device_id_error` decorator that catches `DeviceIDError` (resCode 4002), re-registers `device_id` with thread-safe locking, and retries the call once.

Closes #1155
Closes kia_uvo#1683

### What changed

- **New decorator** `_retry_on_device_id_error` in `ApiImplType1.py`: catches `DeviceIDError`, re-registers `device_id` under `threading.Lock`, retries once
- **Applied to 17 API methods** (was 15 proactive rotation points): `get_vehicles`, `start_charge`, `stop_charge`, `set_charging_current`, `set_charge_limits`, `set_vehicle_to_load_discharge_limit`, `lock_action`, `schedule_charging_and_climate`, `start_climate`, `stop_climate`, `start_hazard_lights`, `start_hazard_lights_and_horn`, `set_windows_state`, `set_navigation`, `charge_port_action`, `valet_mode_action`, `update_vehicle_with_cached_state`, `force_refresh_vehicle_state`
- **Removed all 15** `token.device_id = self._get_device_id(self._get_stamp())` proactive rotation lines from `ApiImplType1.py` (13) and `KiaUvoApiEU.py` (2)
- **`check_action_status` unchanged** — separate fix in #1153
- **New end-to-end validation test** `tests/integration/test_device_id_retry_validation.py` — corrupts `token.device_id` and asserts the decorator recovers against the live EU server (see Validation below)

### Why decorator over per-command rotation

| Aspect | Before (proactive rotation) | After (retry decorator) |
|--------|---------------------------|------------------------|
| When device_id refreshes | After every command, regardless of need | Only on actual `DeviceIDError` |
| Race condition risk | High — `check_action_status` polls with stale ID during rotation | Low — ID stays stable until error, then locked re-registration |
| Read-only call protection | None | `get_vehicles`, `update_vehicle_with_cached_state` now protected |
| Thread safety | None | `threading.Lock` on re-registration |

### Root cause background

EU server invalidates `device_id` when push delivery fails (we register with fake GCM token). Proactive rotation was a workaround that created race conditions during `check_action_status` polling. See #1143 and #1155 for full analysis.

### Why this closes kia_uvo#1683

@timsweb reported in [kia_uvo#1683 (comment)](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1683#issuecomment-4728645908) that when the EU server invalidates `device_id` (resCode 4002), the kia_uvo coordinator enters a **setup_retry loop**: a fresh coordinator is created per retry and reloads the stale `CONF_TOKEN` from config, so the same 4002 repeats indefinitely. He applied a coordinator-side workaround (clear the config token on `DeviceIDError`).

This PR fixes it at the source instead: the decorator re-registers a fresh `device_id` and retries **within the same API call**, so `DeviceIDError` never escapes the API layer — no `UpdateFailed` → `ConfigEntryNotReady` → setup_retry loop, and no need for a kia_uvo-side config-token-clear. The decorator is the right layer (fixes all consumers, no re-login/OTP, no HA-config coupling); the kia_uvo coordinator change proposed in #1683 becomes unnecessary.

### Testing

- **Unit tests**: `tests/test_device_id_retry.py` — 4 tests covering retry-on-error, no-retry-on-success, no-retry-on-other-exceptions, lock behavior. Plus `tests/test_token_thread_safety.py`.
- **Integration tests**: `tests/integration/` — live EU API validation (requires `.env` credentials, excluded from CI via `--ignore=tests/integration`)
- **Stress test**: `tests/integration/device_id_stress_test.py` — 20-min monitoring + action polling
- **Live validation**: `tests/integration/test_device_id_retry_validation.py` — corrupts `token.device_id`, calls a decorated EU state read, asserts recovery. Validated end-to-end against the live EU server: corrupted device_id → `resCode 4002` → decorator re-registers via `notifications/register` → retry returns `resCode 0000` with full vehicle state.

### Out of scope (separate issues)

- #1151 — HTTP timeout for all `requests` calls
- #1152 — Thread-safe `Token.device_id` property
- #1153 — `check_action_status` PENDING-without-SUCCESS fix
- #1190 — decorator coverage gaps (`check_action_status`, non-EU base `ApiImpl` state reads, CA/CN stamp-compatible path)

## Test plan

- [x] All unit tests pass; CI `test py3.12` green
- [x] ruff lint + format clean
- [x] Integration test with real EU credentials — `test_device_id_retry_validation.py` PASSED against the live EU server (corrupted device_id → 4002 → decorator recovers → retry succeeds)
- [ ] Home Assistant integration test — verify coordinator works without `DeviceIDError`

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -1,8 +1,10 @@
 """ApiImplType1.py"""
 
 import datetime as dt
+import functools
 import logging
 import math
+import threading
 from typing import Optional
 
 import time
@@ -47,6 +49,29 @@ from .exceptions import (
 USER_AGENT_OK_HTTP: str = "okhttp/3.12.0"
 
 _LOGGER = logging.getLogger(__name__)
+
+
+def _retry_on_device_id_error(func):
+    """On DeviceIDError, re-register device_id and retry once.
+
+    EU server invalidates device_id when push delivery fails. Instead of
+    proactively rotating device_id after every control command (which causes
+    race conditions), we retry only when the error actually occurs.
+    """
+    _device_id_lock = threading.Lock()
+
+    @functools.wraps(func)
+    def wrapper(self, token, *args, **kwargs):
+        try:
+            return func(self, token, *args, **kwargs)
+        except DeviceIDError:
+            with _device_id_lock:
+                _LOGGER.debug(f"{DOMAIN} - DeviceIDError, re-registering device_id")
+                stamp = self._get_stamp()
+                token.device_id = self._get_device_id(stamp)
+            return func(self, token, *args, **kwargs)
+
+    return wrapper
 
 
 def _check_response_for_errors(response: dict) -> None:
@@ -164,6 +189,7 @@ class ApiImplType1(ApiImpl):
             # This is defensive programming in case the API structure changes
             pass
 
+    @_retry_on_device_id_error
     def get_vehicles(self, token: Token) -> list[Vehicle]:
         url = self.SPA_API_URL + "vehicles"
         response = self.session.get(
@@ -647,6 +673,7 @@ class ApiImplType1(ApiImpl):
 
         vehicle.data = state
 
+    @_retry_on_device_id_error
     def start_charge(self, token: Token, vehicle: Vehicle) -> str:
         if not vehicle.ccu_ccs2_protocol_support:
             url = self.SPA_API_URL + "vehicles/" + vehicle.id + "/control/charge"
@@ -668,9 +695,9 @@ class ApiImplType1(ApiImpl):
         response = self.session.post(url, json=payload, headers=headers).json()
         _LOGGER.debug(f"{DOMAIN} - Start Charge Action Response: {response}")
         _check_response_for_errors(response)
-        token.device_id = self._get_device_id(self._get_stamp())
         return response["msgId"]
 
+    @_retry_on_device_id_error
     def stop_charge(self, token: Token, vehicle: Vehicle) -> str:
         if not vehicle.ccu_ccs2_protocol_support:
             url = self.SPA_API_URL + "vehicles/" + vehicle.id + "/control/charge"
@@ -692,9 +719,9 @@ class ApiImplType1(ApiImpl):
         response = self.session.post(url, json=payload, headers=headers).json()
         _LOGGER.debug(f"{DOMAIN} - Stop Charge Action Response: {response}")
         _check_response_for_errors(response)
-        token.device_id = self._get_device_id(self._get_stamp())
         return response["msgId"]
 
+    @_retry_on_device_id_error
     def set_charging_current(self, token: Token, vehicle: Vehicle, level: int) -> str:
         if not vehicle.ccu_ccs2_protocol_support:
             raise UnsupportedControlError(
@@ -714,9 +741,9 @@ class ApiImplType1(ApiImpl):
         ).json()
         _LOGGER.debug(f"{DOMAIN} - Set Charging Current Response: {response}")
         _check_response_for_errors(response)
-        token.device_id = self._get_device_id(self._get_stamp())
         return response["msgId"]
 
+    @_retry_on_device_id_error
     def set_charge_limits(
         self, token: Token, vehicle: Vehicle, ac: int, dc: int
     ) -> str:
@@ -744,9 +771,9 @@ class ApiImplType1(ApiImpl):
         ).json()
         _LOGGER.debug(f"{DOMAIN} - Set Charge Limits Response: {response}")
         _check_response_for_errors(response)
-        token.device_id = self._get_device_id(self._get_stamp())
         return response["msgId"]
 
+    @_retry_on_device_id_error
     def set_vehicle_to_load_discharge_limit(
         self, token: Token, vehicle: Vehicle, limit: int
     ) -> str:
@@ -764,9 +791,9 @@ class ApiImplType1(ApiImpl):
         ).json()
         _LOGGER.debug(f"{DOMAIN} - Set v2l limit Response: {response}")
         _check_response_for_errors(response)
-        token.device_id = self._get_device_id(self._get_stamp())
         return response["msgId"]
 
+    @_retry_on_device_id_error
     def lock_action(
         self, token: Token, vehicle: Vehicle, action: VEHICLE_LOCK_ACTION
     ) -> str:
@@ -789,7 +816,6 @@ class ApiImplType1(ApiImpl):
         response = self.session.post(url, json=payload, headers=headers).json()
         _LOGGER.debug(f"{DOMAIN} - Lock Action Response: {response}")
         _check_response_for_errors(response)
-        token.device_id = self._get_device_id(self._get_stamp())
         return response["msgId"]
 
     def check_action_status(
@@ -852,6 +878,7 @@ class ApiImplType1(ApiImpl):
             # Old code: raise APIError(f"No action found with ID {action_id}")
             return ORDER_STATUS.UNKNOWN
 
+    @_retry_on_device_id_error
     def schedule_charging_and_climate(
         self,
         token: Token,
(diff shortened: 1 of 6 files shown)


cdnninja:
Is this the right way to fix it vs figuring out the registration side?

blka:
The registration side can't be fixed from this library. Here's why:

The server invalidates `device_id` when push delivery fails. We register with a fake GCM/APNS push token — there's no real device behind it, so push always fails, and the server expires the session.

To fix registration, we'd need a real push token. Getting one requires either a real mobile app (with the right package name and signing certificate) or a Firebase Cloud Messaging backend that listens for push notifications. This library is a Python API client — it can't obtain real push tokens from the OS, and even if it could, those tokens are bound to the Hyundai/Kia app identity and signing cert.

So the options are:
1. **Decorator retry** — treat `DeviceIDError` as recoverable, re-register with thread-safe locking, retry once. This is what #1156 does.
2. **Fix registration** — requires infrastructure outside this library's scope.
3. **Proactive rotation** — current status quo, race-condition-prone during `check_action_status` polling.

The decorator is the only feasible fix within this library.

Have you considered any other approach? You've been around this project longer than I have — curious if there's something I'm missing.

cdnninja:
Could we not register?   How does the native app handle it with push disabled?  This is also assuming push is the issue, we don't know that for fact just a theory of mine.

blka:
We can't skip registration — the server returns `deviceId` from the `notifications/register` endpoint, and every subsequent API call requires it in the headers. Without registration, we have no `deviceId` and can't make any API calls at all.

Regarding push being the root cause: you're right, it's a theory not proven fact. The native app still registers for push even when notifications are disabled at the OS level — FCM/GCM token is obtained regardless, the OS just doesn't display the notifications. So from the server's perspective, the push token is valid, the device is "active", and `deviceId` stays alive.

Our library registers with a fake `pushRegId` (random hex). If the server tries to deliver a push and gets a delivery failure, that could trigger `deviceId` invalidation — but we don't know for certain. It could also be session timeout, token rotation, or something else entirely.

The only way to confirm would be mitmproxy capture of the native app's behavior — watch what happens to `deviceId` over time with the real app. I'm working on that for the CCI API investigation.

In the meantime, this PR's decorator approach is the pragmatic fix: handle `DeviceIDError` gracefully when it happens, regardless of root cause. The CCI API (new EU endpoint) eliminates device_id entirely with a different auth flow, so long-term this becomes moot for EU.

Should I simplify the PR to just the decorator without the extra scope, or keep as-is?

blka:
@cdnninja could you take a look? The branch was force-updated — rebased onto current master (it was 19 commits behind) and cleaned up (the previous version carried local-only noise commits; it's now 3 focused commits on master). No logic changed beyond the rebase + a new validation test.

**Why the update:** @timsweb reported in [kia_uvo#1683 (comment)](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1683#issuecomment-4728645908) that when the EU server invalidates `device_id` (resCode 4002), the kia_uvo coordinator enters a **setup_retry loop** — a fresh coordinator is spun up per retry and reloads the stale `CONF_TOKEN` from config, so the same 4002 repeats forever. He applied a coordinator-side workaround (clear the config token on `DeviceIDError`).

This PR closes kia_uvo#1683 at the source instead: the `_retry_on_device_id_error` decorator re-registers a fresh `device_id` and retries **within the same API call**, so `DeviceIDError` never escapes the API layer → no `UpdateFailed` → no `ConfigEntryNotReady` → no setup_retry loop, and no need for a kia_uvo-side config-token-clear. The decorator is the right layer (fixes all consumers, no re-login/OTP, no HA-config coupling); the kia_uvo coordinator change proposed in #1683 becomes unnecessary.

**Validation added:** `tests/integration/test_device_id_retry_validation.py` corrupts `token.device_id`, calls a decorated EU state read, and asserts recovery against the live EU server. Ran it with real EU credentials: corrupted device_id → `resCode 4002` → decorator logs "DeviceIDError, re-registering device_id" → fresh `notifications/register` → retry returns `resCode 0000` with full vehicle state. This is the end-to-end proof the unit tests' mocked `_get_device_id` can't provide, and it closes the gap from the earlier stress test that couldn't reproduce `DeviceIDError`.

**Coverage gaps** (decorator not yet on `check_action_status`, non-EU base `ApiImpl` state reads, CA/CN) are tracked in #1190, scoped out of this PR to keep it focused.

CI is green (Analyze, CodeQL, main, pre-commit, test py3.12). Ready for review.

cdnninja:
We have too many changes being requested, multiple with retry logic that will become hard to support.   I would like to see less architectural changes happening at the same time.  I appreciate the keen interest but want to make sure we are careful.  Are you able to message me on discord?

blka:
no worries,  I think I am done for now. I have working EU CCI with MQTTy and need to focus on it. 
If you are able, please check and merge one-by-one I will be producing any needed changes after the reviews. Even if you don’t want to merge anything right now – that’s fine – I think a lot of new code would provide a good foundation for solving most of the issues that are currently holding the library back.

PR #1158, opened 2026-05-26
feat: add UserAccount dataclass and EU user profile endpoint

blka:
## Summary

- Add `UserAccount` dataclass (user_id, email, name, mobile_number, language, country, status, sign_up_date, pin_date)
- Add `KiaUvoApiEU.get_user_profile()` fetching `GET /api/v1/user/profile` with graceful degradation on failure
- Add `KiaUvoApiEU._map_user_account()` static method for EU response mapping
- Add `ApiImpl.get_user_profile()` no-op base returning `None` (non-EU regions unaffected)
- Wire `VehicleManager.account` in `initialize_vehicles()` so user account data auto-populates on login
- 4 unit tests + sanitized EU user profile fixture

**Depends on:** #1157 (VehicleProfile + capability properties)

## Test plan

- [x] `pytest tests/` — 201 tests pass (8 snapshots)
- [x] `UserAccount` dataclass instantiation and field access
- [x] `_map_user_account()` mapping from real API fixture
- [x] Missing fields default to `None`
- [x] Non-EU regions return `None` from base `get_user_profile()`

## Deferred to separate PR

- Per-country timezone detection (`data_timezone` per-vehicle)


hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -140,6 +140,13 @@ class ApiImpl:
     def __init__(self) -> None:
         """Initialize."""
 
+    def _fetch_vehicle_profiles(self, token: Token, vehicles: list[Vehicle]) -> None:
```
cdnninja: Internal methods don't need to be exposed in the impl as they wouldn't be exposed.

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -152,6 +154,144 @@ def __init__(self, region: int, brand: int, language: str) -> None:
                 "https://accounts-eu.genesis.com/realms/eugenesisidm/ga-api/redirect2"
             )
 
+    @staticmethod
+    def _str(v):
```
cdnninja: This looks like a utility.

hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -152,6 +154,144 @@ def __init__(self, region: int, brand: int, language: str) -> None:
...
+        """Coerce int-or-string option values to strings for capability comparisons."""
+        if v is None:
+            return None
+        return str(v)
+
+    def _map_vehicle_profile(self, profile_data: dict) -> VehicleProfile:
```
cdnninja, line 158: I think all of this could move up to impltype1.

hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -127,9 +128,11 @@ def initialize_vehicles(self):
                 "Vehicles already initialized, this will re-initialize and cause data loss mapping errors"
             )
         vehicles = self.api.get_vehicles(self.token)
+        self.api._fetch_vehicle_profiles(self.token, vehicles)
         for vehicle in vehicles:
             vehicle.supports_window_control = self.api.supports_window_control
             self.vehicles[vehicle.id] = vehicle
+        self.account = self.api.get_user_profile(self.token)
```
cdnninja, line 135: I think this is creating a new object which means update will never flow through to HA as memory locations change.

cdnninja (changes requested):
A few initial thoughts.

hyundai_kia_connect_api/KiaUvoApiEU.py
blka: true, but used only in one place. will move.

hyundai_kia_connect_api/KiaUvoApiEU.py
blka: don't quite. EU resposne is specified for the region - fields names like `ecuOtaUpdateSupport, drvSeatLoc, strgWhlHeatOption, seatHeaterVent` are EU-specific camelCase. Other regions (CA, AU) use different field names in their APIs. If CA/AU receive endpoint profiles, they will need their own mapping. Moving the EU mapping to Type1 would imply a level of sharing that does not exist.
any suggestion?

hyundai_kia_connect_api/VehicleManager.py
blka: true. 
the user profile is fetched only once in `initialize_vehicles()`. There is no update cycle for the account. Data such as email, first name and country do not change between update cycles, so there is no need to refresh it. This is static data, not dynamic like the vehicle state. will implement update_account() method in VehicleManager which will update in-place.
Or perhaps you have a different take on this solution?

hyundai_kia_connect_api/ApiImpl.py
blka: Two options:
- Move profile fetching into `KiaUvoApiEU.get_vehicles()` by overriding it and calling `super().get_vehicles()` first, then `self._fetch_vehicle_profiles()`. This way `_fetch_vehicle_profiles` stays a truly private method of the EU class with no base-class stub, and VehicleManager doesn't need to know about profiles at all. Other regions without profile endpoints are unaffected.
 
-  Rename to `fetch_vehicle_profiles` (drop the underscore) and keep the base-class no-op. VehicleManager calls it explicitly, but now it's a public API method that every region must stub.

My option is the first one, but you might prefer to do it differently?

cdnninja: I think since this API could be used by others we use 2.   Each region doesn't need to stub it, just the parent class.

cdnninja: My reason thinking on 2 is this library is used by some that don't need this data.

hyundai_kia_connect_api/VehicleManager.py
cdnninja: Oops yes missed that. Should be fine.

hyundai_kia_connect_api/KiaUvoApiEU.py
cdnninja: CA doesn't use that class.    Are you sure those values don't exist for AU? So far almost everything for AU aligns to EU which is why we have Type1 to share the code.

Lekensteyn (commented):
This class was originally added for extracting timezone information. That turns out not necessary, the related PR was closed.

Do we still need this UserAccount and profiles functionality? If there are no benefits to kia_uvo or other users, let's not add a lot of extra code that makes maintenance harder?

One clear disadvantage of this PR, it adds extra remote API calls in `initialize_vehicles`. For applications that are only interested in downloading trip info, it only slows down the process and creates extra load on the server.

blka:
Closing without merge: UserAccount has no consumer in kia_uvo or elsewhere (per @Lekensteyn's review — no end-user benefit, adds extra API calls). The capability work (VehicleProfile + 13 capability properties) continues in #1157; kia_uvo will gain support for the new capability-gated entities in a near-future follow-up PR. Thanks for the review.

PR #1159, opened 2026-05-26
feat: per-country timezone detection for EU API

blka:
## Summary

- Convert `data_timezone` from class attribute to instance attribute (default `Europe/Berlin`) in `KiaUvoApiEU`, so each instance can have its own timezone
- Add `EU_COUNTRY_TIMEZONES` mapping (31 EU countries → IANA timezone names) in `const.py`
- Add `KiaUvoApiEU._detect_user_timezone()` — fetches `GET /api/v1/user/language`, extracts country code, sets `self.data_timezone` to the correct `ZoneInfo`
- Add `ApiImpl._detect_user_timezone()` no-op base (other regions unaffected)
- Wire `_detect_user_timezone()` into `VehicleManager.initialize_vehicles()` after user profile fetch
- Graceful degradation: network failure or unknown country keeps `Europe/Berlin` default
- 10 unit tests covering IANA validation, country→timezone mapping, mock API responses, and failure fallback

**Depends on:** #1157 (VehicleProfile), #1158 (UserAccount)

## Problem

EU API users in Portugal, Finland, Greece, Ireland, etc. get wrong timestamps because `data_timezone` is hardcoded to `Europe/Berlin`. DST transitions produce incorrect times for 28 out of 31 EU countries.

## Test plan

- [x] `pytest tests/` — 211 tests pass (8 snapshots)
- [x] All IANA timezone names in `EU_COUNTRY_TIMEZONES` are valid `ZoneInfo` entries
- [x] Poland → Europe/Warsaw, Portugal → Europe/Lisbon, Finland → Europe/Helsinki
- [x] Unknown country falls back to Europe/Berlin
- [x] Network failure keeps default Europe/Berlin

## Deferred to separate PRs

- Per-vehicle timezone override (using `vehicle.profile.country`)


Lekensteyn:
This PR seems way overkill with lots of code added for unused fields. Is there any evidence of actual affected users, or is this just a guess?

For example, there are currently heuristics that help detect the right timezone for Canadian users. Maybe we can use the same, or use the "offset" API response. But only if there is actually a report of a real affected user.

Adding a lot of code that turns out to be unnecessary just adds to the maintenance burden.

cdnninja:
Agreed I am not sure this is the right approach.  Right now offset is provided.   Is this actually a DST issue rather than a re-write?

blka:
Closing — per maintainer feedback, per-country timezone detection is overkill without evidence of affected users. PR #1147 correctly handles DST for CET countries, which covers the known user base.

PR #1160, opened 2026-05-27
feat: add ApiImplSession with timeout and connection pooling

### LINKED ISSUE #1151: All HTTP requests missing timeout parameter (139 calls across 10 files)

blka:
## Summary

Replaces bare `requests.get/post/put` calls across all 8 region implementations with a shared `ApiImplSession(requests.Session)` that provides timeout defaults, urllib3 retry, and TLS connection pooling.

Closes #1151

### What changed

**New class `ApiImplSession` in `ApiImpl.py`:**
- Default timeout `(10s connect, 30s read)`, overridable per region via `self.session.HTTP_READ_TIMEOUT = 45`
- `urllib3.util.retry.Retry` on 502/503/504 (2 retries, backoff=1)
- `requests.exceptions.Timeout` → `RequestTimeoutError` mapping
- Connection pooling via `requests.Session` (reuses TLS connections)

**All regions migrated:**
| Region | Before | After |
|--------|--------|-------|
| EU/AU/IN/CN | Bare `requests.get/post` | `self.session.get/post` (ApiImplSession) |
| CA | Custom `RetrySession(requests.Session)` | `ApiImplSession` (urllib3 Retry replaces manual retry) |
| USA Kia | `requests.Session` + `KiaSSLAdapter` | `ApiImplSession` + `KiaSSLAdapter` |
| USA Hyundai | `requests.Session` + `cipherAdapter` | `ApiImplSession` + `cipherAdapter` |
| BR | Plain `requests.Session` | `ApiImplSession` |

**Other changes:**
- `self.session` attribute name unified (was `sessions`, `session`, `_session`)
- `KiaUvoApiCA.RetrySession` class removed (35 lines)
- All `# pylint:disable=missing-timeout` removed
- Nominatim geocoding: explicit `timeout=(5, 15)` (third-party, fail fast)
- Temporary `requests.Session()` in auth flows → `ApiImplSession()`

### Why Session-based instead of per-call timeout

Per [maintainer feedback](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/1151#issuecomment-XXXXXXXX): instead of adding `timeout=` to 139 individual calls, configure timeout at the session level. This also enables connection reuse (the mobile apps reuse TLS connections) and provides a single place for retry/timeout configuration.

### Testing

- 9 new unit tests in `tests/test_api_impl_session.py`: default timeout, per-call override, region override, Timeout→RequestTimeoutError, retry adapter
- All 176 existing tests pass
- ruff lint + format clean
- Pre-commit hooks pass

### Out of scope

- Per-region timeout customization (infrastructure is there, no regions override yet)
- CA RetrySession behavior difference: urllib3 Retry retries on 502/503/504, old RetrySession retried on ConnectionError. If ConnectionError retry is needed, add to `status_forcelist` or use `Retry.raise_on_status=False`

### DIFF
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -1,6 +1,6 @@
 """ApiImpl.py"""
 
-# pylint:disable=unnecessary-pass,missing-class-docstring,invalid-name,missing-function-docstring,wildcard-import,unused-wildcard-import,unused-argument,missing-timeout,logging-fstring-interpolation
+# pylint:disable=unnecessary-pass,missing-class-docstring,invalid-name,missing-function-docstring,wildcard-import,unused-wildcard-import,unused-argument,logging-fstring-interpolation
 import datetime as dt
 import logging
 from dataclasses import dataclass
@@ -130,6 +130,33 @@ class POIInfo:
         }
 
 
+class ApiImplSession(requests.Session):
+    """Shared HTTP session with default timeout and connection pooling.
+
+    All regions should use this session (or a subclass) for HTTP calls.
+    Override class attributes per region in __init__ if needed.
+
+    Retry is intentionally NOT configured here. Pre-PR behavior retried only
+    CA connection errors (#857, login error 104), and not all requests are
+    safe to replay (e.g. non-idempotent control POSTs). If connection-reset
+    issues reappear, re-add a login-only connection retry. See PR #1160.
+    """
+
+    HTTP_CONNECT_TIMEOUT = 10
+    HTTP_READ_TIMEOUT = 30
+
+    def request(self, method, url, **kwargs):
+        kwargs.setdefault(
+            "timeout", (self.HTTP_CONNECT_TIMEOUT, self.HTTP_READ_TIMEOUT)
+        )
+        try:
+            return super().request(method, url, **kwargs)
+        except requests.exceptions.Timeout as exc:
+            from .exceptions import RequestTimeoutError
+
+            raise RequestTimeoutError(str(exc)) from exc
+
+
 class ApiImpl:
     data_timezone = dt.timezone.utc
     temperature_range = None
@@ -231,7 +258,7 @@ class ApiImpl:
                     + email_parameter
                 )
                 headers = {"user-agent": "curl/7.81.0"}
-                response = requests.get(url, headers=headers)
+                response = requests.get(url, headers=headers, timeout=(5, 15))
                 try:
                     response = response.json()
                 except JSONDecodeError:
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -1,7 +1,6 @@
 """ApiImplType1.py"""
 
 import datetime as dt
-import requests
 import logging
 import math
 from typing import Optional
@@ -11,6 +10,7 @@ from time import sleep
 
 from .ApiImpl import (
     ApiImpl,
+    ApiImplSession,
     ScheduleChargingClimateRequestOptions,
     ClimateRequestOptions,
     WindowRequestOptions,
@@ -119,6 +119,7 @@ class ApiImplType1(ApiImpl):
 
     def __init__(self) -> None:
         """Initialize."""
+        self.session = ApiImplSession()
 
     def _sanitize_ice_value(self, state: dict) -> None:
         """
@@ -165,7 +166,7 @@ class ApiImplType1(ApiImpl):
 
     def get_vehicles(self, token: Token) -> list[Vehicle]:
         url = self.SPA_API_URL + "vehicles"
-        response = requests.get(
+        response = self.session.get(
             url,
             headers=self._get_authenticated_headers(token),
         ).json()
@@ -664,7 +665,7 @@ class ApiImplType1(ApiImpl):
             headers = self._get_control_headers(token, vehicle)
 
         _LOGGER.debug(f"{DOMAIN} - Start Charge Action Request: {payload}")
-        response = requests.post(url, json=payload, headers=headers).json()
+        response = self.session.post(url, json=payload, headers=headers).json()
         _LOGGER.debug(f"{DOMAIN} - Start Charge Action Response: {response}")
         _check_response_for_errors(response)
         token.device_id = self._get_device_id(self._get_stamp())
@@ -688,7 +689,7 @@ class ApiImplType1(ApiImpl):
             headers = self._get_control_headers(token, vehicle)
 
         _LOGGER.debug(f"{DOMAIN} - Stop Charge Action Request: {payload}")
-        response = requests.post(url, json=payload, headers=headers).json()
+        response = self.session.post(url, json=payload, headers=headers).json()
         _LOGGER.debug(f"{DOMAIN} - Stop Charge Action Response: {response}")
         _check_response_for_errors(response)
         token.device_id = self._get_device_id(self._get_stamp())
@@ -704,7 +705,7 @@ class ApiImplType1(ApiImpl):
         )
 
         body = {"chargingCurrent": level}
-        response = requests.post(
+        response = self.session.post(
             url,
             json=body,
             headers=self._get_authenticated_headers(
@@ -734,7 +735,7 @@ class ApiImplType1(ApiImpl):
             ]
         }
         _LOGGER.debug(f"{DOMAIN} - Set Charge Limits Body: {body}")
-        response = requests.post(
+        response = self.session.post(
             url,
             json=body,
             headers=self._get_authenticated_headers(
@@ -754,7 +755,7 @@ class ApiImplType1(ApiImpl):
         )
 
         body = {"dischargingLimit": int(limit)}
-        response = requests.post(
+        response = self.session.post(
             url,
             json=body,
             headers=self._get_authenticated_headers(
@@ -785,7 +786,7 @@ class ApiImplType1(ApiImpl):
 
         _LOGGER.debug(f"{DOMAIN} - Lock Action Request: {payload}")
 
-        response = requests.post(url, json=payload, headers=headers).json()
+        response = self.session.post(url, json=payload, headers=headers).json()
         _LOGGER.debug(f"{DOMAIN} - Lock Action Response: {response}")
         _check_response_for_errors(response)
         token.device_id = self._get_device_id(self._get_stamp())
@@ -823,7 +824,7 @@ class ApiImplType1(ApiImpl):
             return ORDER_STATUS.TIMEOUT
 
         else:
-            response = requests.get(
+            response = self.session.get(
                 url,
                 headers=self._get_authenticated_headers(
                     token, vehicle.ccu_ccs2_protocol_support
@@ -957,7 +958,7 @@ class ApiImplType1(ApiImpl):
         }
 
         _LOGGER.debug(f"{DOMAIN} - Schedule Charging and Climate Request: {payload}")
-        response = requests.post(
+        response = self.session.post(
             url, json=payload, headers=self._get_control_headers(token, vehicle)
         ).json()
         _LOGGER.debug(f"{DOMAIN} - Schedule Charging and Climate Response: {response}")
@@ -1018,7 +1019,7 @@ class ApiImplType1(ApiImpl):
                 "unit": "C",
             }
             _LOGGER.debug(f"{DOMAIN} - Start Climate Action Request: {payload}")
-            response = requests.post(
+            response = self.session.post(
                 url,
                 json=payload,
                 headers=self._get_authenticated_headers(
@@ -1050,7 +1051,7 @@ class ApiImplType1(ApiImpl):
                 "windshieldFrontDefogState": options.defrost,
             }
             _LOGGER.debug(f"{DOMAIN} - Start Climate Action Request: {payload}")
-            response = requests.post(
+            response = self.session.post(
                 url,
                 json=payload,
                 headers=self._get_control_headers(token, vehicle),
@@ -1074,7 +1075,7 @@ class ApiImplType1(ApiImpl):
                 "unit": "C",
             }
             _LOGGER.debug(f"{DOMAIN} - Stop Climate Action Request: {payload}")
-            response = requests.post(
+            response = self.session.post(
                 url,
                 json=payload,
                 headers=self._get_authenticated_headers(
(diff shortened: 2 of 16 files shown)


blka:
Note: PR #1156 (DeviceIDError retry decorator) modifies overlapping files (`ApiImplType1.py`, `KiaUvoApiEU.py`). I recommend merging this session PR first — it's the HTTP transport foundation that all API calls go through. After merge, I'll rebase #1156 on top to add the decorator methods on the session-based code.

The changes are in different locations (this PR changes HTTP call sites, #1156 adds decorators on `def` lines), so rebasing should be straightforward.

cdnninja:
@Lekensteyn are you able to provide thoughts on this?

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -130,6 +132,41 @@ def to_dict(self) -> dict:
...
+    HTTP_READ_TIMEOUT = 30
+    MAX_RETRIES = 2
+    RETRY_BACKOFF = 1
+
+    def __init__(self):
+        super().__init__()
+        retry = Retry(
+            total=self.MAX_RETRIES,
+            backoff_factor=self.RETRY_BACKOFF,
+            status_forcelist=[502, 503, 504],
```
Lekensteyn: Why would you backoff on these 5xx server errors? The current implementation before this PR doesn't retry HTTP errors. It only retries connection errors.

According to #857 by @alphalneo where the connection retry error was added, it was done due to errors during login only.

Let's be careful and remove retries by default, even during login. Not all requests are safe to replay.

If @alphalneo or others report connection reset issues again, we can re-add it for logins only. And perhaps debug via a packet capture + SSLKEYLOGFILE to get an idea where the error happens.

Lekensteyn (changes requested):
The `requests` -> `self.session` changes conversion looks straightforward. I haven't been able to test this yet.

@blka Please do not add unrelated changes to this PR. The new integration test change doesn't belong here. Should there be any breakage, the squashed commit can be cleanly reverted if it doesn't have unrelated changes.

Consider rebasing on current master instead of merging master into this PR.

blka:
Thanks @Lekensteyn — all three points addressed.

**Retry (5xx):** removed. `ApiImplSession` now provides only default timeout + connection pooling + `Timeout`→`RequestTimeoutError`. No `urllib3 Retry`, no `status_forcelist`. Rationale as you noted: pre-PR behavior retried only CA `ConnectionError` (#857, login error 104), and 5xx replay is unsafe for non-idempotent control POSTs. If connection-reset reports return, I will re-add a login-only connection retry (and debug via packet capture + SSLKEYLOGFILE as you suggested).

**CA:** left untouched on purpose — `KiaUvoApiCA` keeps its existing `RetrySession` (ConnectionError retry from #857) and `self.sessions` attribute. I did not want to fold the CA retry-safety question into this PR. Adding timeout to CA's session is deferred to a follow-up.

**Unrelated changes:** integration test infrastructure removed from this PR.

**Rebase:** rebased on current master — the two merge commits are gone, it is a single commit on top of master now.

Diff is now 16 files: `ApiImplSession` + migration of 7 regions (EU/AU/IN/CN/USA Kia/USA Hyundai/BR) + tests. 200 tests pass, ruff clean.

Lekensteyn (approved):
I reviewed the patch locally, the changes are targeted and look good to me:
```sh
git fetch origin pull/1160/head
git show -I '.* (self\.sessions?|requests)\.(put|post|get)\((url, headers=headers)?|api\.sessions?[ .(].*' --color-words  -w -p --stat FETCH_HEAD
```

I tested it out with Hyundai (EU), and other than lack of connection reuse (explained in https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/issues/1151#issuecomment-4759276902), it doesn't seem to break anything.

The main benefit of this PR is a shared timeout configuration, there is no connection pooling for at least the EU region due to server configuration.

blka:
@Lekensteyn Following up on your observation that there's no connection pooling for EU — I ran a live probe of every regional endpoint to map this out. Method: 2 sequential GETs over a single `requests.Session` per host, observing the response `Connection` header and urllib3's `Resetting dropped connection` log (the definitive signal that the server closed the socket and the connection could not be reused).

**Hosts that send `Connection: close` (no reuse — full new TCP+TLS per request):**
- EU CCAPI: `prd.eu-ccapi.kia.com:8080`, `prd.eu-ccapi.hyundai.com:8080`, `prd-eu-ccapi.genesis.com:443`
- CN CCAPI: `prd.cn-ccapi.kia.com`, `prd.cn-ccapi.hyundai.com`
- BR: `br-ccapi.hyundai.com.br`
- USA Hyundai: `api.telematics.hyundaiusa.com`

**Hosts that send `keep-alive` (reuse works — 2nd request reuses the socket):**
- EU SSO/Keycloak: `idpconnect-eu.hyundai.com`, `eu-account.hyundai.com`
- EU new CCI gateway: `cci-api-eu.hyundai.com`, `gspa-ccs-eu.hyundai.com` (the replacement for the legacy CCAPI gateway — these do keep connections alive)
- AU: `au-apigw.ccs.kia.com.au:8082`, `au-apigw.ccs.hyundai.com.au:8080`
- IN: `prd.in-ccapi.{hyundai,kia}.connected-car.io:8080`
- USA Kia: `api.owners.kia.com`
- CA: `kiaconnect.ca`, `mybluelink.ca`, `genesisconnect.ca`

So your finding holds for the **legacy EU CCAPI gateway** (`prd.eu-ccapi.*:8080`) — and the same `close` behavior extends to CN, BR, and USA Hyundai, all of which run the same legacy Hyundai edge without a CDN. The Cloudflare-fronted regions (CA, USA Kia), the newer gateways (AU apigw, IN connected-car.io), and the new EU CCI gateway all keep connections alive. So the lack of pooling on the EU legacy path is a property of that specific legacy gateway, not EU-wide — the EU replacement (CCI) does reuse.

I also verified the auth status does not change the `Connection` header — on the keep-alive hosts, 401 / 403 / 500 responses all returned `keep-alive` with the connection reused on the 2nd request. The keep-alive policy is set at the edge, before auth is evaluated, so the unauthenticated probe is authoritative for this question.

Net: the `ApiImplSession` approach (plain `requests.Session` + default timeout, no `HTTPAdapter` retry) is already optimal for both cases — the Session reuses when the server permits and reconnects when the server sends `close`, with no per-region config needed. The shared timeout remains the primary benefit of this PR; connection reuse is a bonus where the server allows it. No change to the PR suggested — just confirming the EU `close` behavior is a property of the legacy CCAPI gateway specifically, not EU-wide.

PR #1124, opened 2026-05-10
fix: add CCS2 protocol support to force_refresh_vehicle_state

blka:
## Summary

Add CCS2 protocol branching to `force_refresh_vehicle_state` across all affected regions (EU, AU, IN, CN).

The legacy `/status` endpoint returns empty metadata (`ServiceNo`, `RetCode` only) for CCS2-capable vehicles. When `force_refresh_vehicle_state` calls this endpoint and parses the response with the legacy parser, all vehicle properties are overwritten with `None` values — causing every entity in Home Assistant to become `unknown`.

This fix mirrors the CCS2 branching logic already present in `update_vehicle_with_cached_state`: for CCS2 vehicles, the `/ccs2/carstatus/latest` endpoint and appropriate CCS2 parser are used instead.

## Root Cause

`force_refresh_vehicle_state` was not protocol-aware. It unconditionally used:
1. The legacy `/status` endpoint (which returns empty data for CCS2 vehicles)
2. The legacy `_update_vehicle_properties` parser (which writes `None` for missing keys)

Meanwhile, `update_vehicle_with_cached_state` correctly branches on `vehicle.ccu_ccs2_protocol_support` and uses the CCS2 endpoint + parser. This asymmetry meant that cached updates worked fine, but any force refresh wiped all vehicle state.

+ CCS2 API was returning temp HVAC as a string "23.0", but HA require  float
### Directly affected (fix applied)
| File | Region | Notes |
|------|--------|-------|
| `KiaUvoApiEU.py` | EU | CCS2 branch + `_update_vehicle_properties_ccs2` + location via `_set_cached_location_park` |
| `KiaUvoApiAU.py` | AU | CCS2 branch + `_update_vehicle_properties_ccs2` + explicit location override |
| `KiaUvoApiIN.py` | IN | CCS2 branch + `_update_vehicle_properties` (IN's own parser) + `_update_vehicle_location` |
| `KiaUvoApiCN.py` | CN | Defensive CCS2 branch (CN doesn't currently populate `ccuCCS2ProtocolSupport`, so this is a no-op today) |
| `ApiImplType1.py` | * | CCS2 API was returning temp HVAC as a string "23.0", but HA require  float |

### Not affected (no changes needed)
| File | Region | Why |
|------|--------|-----|
| `HyundaiBlueLinkApiBR.py` | BR | Already branches on CCS2 in shared `_get_vehicle_state` — correct architecture |
[...]

PR #1161, opened 2026-05-27
fix: preserve last known unit when API returns None (unit-only)

blka:
## Summary

Vehicle `(value, unit)` setters now keep the last known **unit** when the API returns `None` for the unit, instead of overwriting it. Prevents the Home Assistant unit flip-flop (`km` → `""` → `km`) and the recorder error:

```
The unit of sensor.X_total_driving_range (None) cannot be converted to the unit
of previously compiled statistics (km). Generation of long term statistics will
be suppressed...
```

Reported in kia_uvo#1430 and kia_uvo#1725 (the latter has a real recorder log).

## Scope — unit only (per review)

The value itself is **not** preserved: if the API returns `None` for a value, that is reflected as-is (the data is genuinely unavailable). Only the unit is held. This is the scope cdnninja asked for in the earlier round of this PR: *"only unit should hold."*

## Why reopen (I closed this too hastily before)

When I closed this PR earlier I wrote that the flip-flop "may have been a symptom of #1124 rather than a separate issue." That was wrong. kia_uvo#1725 is a CA / Hyundai Ioniq6 (Type1) vehicle — #1124 only touches CCS2 force-refresh, so it cannot have fixed this. The recorder log in #1725 is the unit→`None` flip happening on a non-CCS2 vehicle, ~once a month. That's the evidence I was missing when I agreed to close.

Applied to all 19 `(value, unit)` setters: `total_driving_range`, `next_service_distance`, `last_service_distance`, `odometer`, `outside_temperature`, `air_temperature`, `ev_battery_water_temperature`, `ev_battery_temperature_min/max`, `ev_driving_range`, `ev_estimated_*_charge_duration` (4), `ev_target_range_charge_AC/DC`, `ev_first/second_departure_climate_temperature`, `fuel_driving_range`.

## Test plan

- [x] `tests/test_vehicle_unit_preservation.py` — unit preserved when API returns `None` unit (all 19 setters), unit updates when API returns a new unit, value reflects `None` when API returns `None` value
- [x] `pytest tests/ -q` — 216 passed
- [x] `pre-commit` clean

### DIFF
--- a/hyundai_kia_connect_api/Vehicle.py
+++ b/hyundai_kia_connect_api/Vehicle.py
@@ -392,8 +392,9 @@ class Vehicle:
     @total_driving_range.setter
     def total_driving_range(self, value):
         self._total_driving_range_value = value[0]
-        self._total_driving_range_unit = value[1]
         self._total_driving_range = value[0]
+        if value[1] is not None:
+            self._total_driving_range_unit = value[1]
 
     @property
     def next_service_distance(self):
@@ -402,8 +403,9 @@ class Vehicle:
     @next_service_distance.setter
     def next_service_distance(self, value):
         self._next_service_distance_value = value[0]
-        self._next_service_distance_unit = value[1]
         self._next_service_distance = value[0]
+        if value[1] is not None:
+            self._next_service_distance_unit = value[1]
 
     @property
     def last_service_distance(self):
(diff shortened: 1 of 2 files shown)


cdnninja:
Based on no responses with people with the issue still I wonder if this was caused by the force update issue.    Should we close for now?  

If we do proceed I think only unit should hold.

blka:
Good point. I agree with closing this for now — the original bug reports (#1430, #1334) haven't been reproduced since the CCS2 force refresh fix (#1124) went in, so the unit flip-flop may have been a symptom of that rather than a separate issue.

If we do revisit: I agree only the unit should hold. If the API returns `None` for a value, we should reflect that (the data is genuinely unavailable), not preserve stale data. But a `None` unit causes HA to forget the measurement system, which triggers the flip-flop notification — that's the real bug. So unit-only protection is the right scope.

Closing.

PR #1163, opened 2026-05-29
fix: handle HATA 502 server error gracefully on cached vehicle status

blka:
## Summary

Fixes #1651 — USA Genesis (HyundaiBlueLinkApiUSA) entities become unavailable when the HATA backend returns `errorCode: 502` on cached vehicle status requests.

Since approximately May 6, 2026, the HATA backend for USA Genesis vehicles returns a 502 server error on `remoteVehicleStatus` with `REFRESH: false` (cached). Force refresh (`REFRESH: true`) works fine. This causes all vehicle entities in Home Assistant to become unavailable on every cached poll cycle.

## Root Cause Analysis

There are **two overlapping bugs** introduced by PR #1097 (April 2026, @cdnninja):

### Bug 1: 502 is a server error, not an auth failure

PR #1097 added `{"502": AuthenticationError}` to `_check_response_for_errors`, treating 502 as an authentication failure. This was based on observing 502 errors on the **login endpoint**, where a 502 *may* indicate invalid credentials. But on the **vehicle status endpoint**, 502 is a server-side failure — jdhume's diagnostics showed the exact HATA response:

```json
{
  "errorCode": 502,
  "errorSubCode": "GEN",
  "systemName": "HATA",
  "functionName": "remoteVehicleStatus",
  "errorMessage": "We're sorry, but we could not complete your request. Please try again later."
}
```

This is clearly not an auth error. Mapping it to `AuthenticationError` would trigger token invalidation and re-login attempts, which is the wrong response to a transient server failure.

### Bug 2: int vs string key mismatch (a lucky accident)

The mapping used `"502"` (string) as the dict key, but the HATA API returns `errorCode` as an **int** `502`. In Python, `502 in {"502": ...}` is `False`, so the buggy mapping was **never actually triggered** — the code fell through to the generic `APIError` path instead.

This is the only reason users weren't hit with the worse consequence of Bug 1 (token invalidation loops). It's a lucky type mismatch that accidentally prevented the real damage.

### Why the original code was correct

Before PR #1097, `HyundaiBlueLinkApiUSA._check_response_for_errors` had **no** errorCode-to-exception mapping. It simply raised `APIError` for any error code — the correct behavior. PR #1097 incorrectly added the 502 mapping. For reference, the base class `ApiImplType1._check_response_for_errors` has specific mappings for specific error codes (e.g. `401` → auth, `429` → rate limit), but **502 is not in that mapping** because it's not a client-recoverable error.

## The Fix

Three changes:

### 1. `_check_response_for_errors`: Remove the incorrect mapping
Remove `{"502": AuthenticationError}`. All errorCodes now raise `APIError`. This is the correct behavior — HATA 502 is a server error, not an auth failure. No known errorCode should map to `AuthenticationError` in this API implementation.

### 2. `update_vehicle_with_cached_state`: Gracefully handle cached status failure
Catch `APIError` from `_get_vehicle_status(token, vehicle, False)` and return early with a warning instead of propagating the exception. Cached vehicle status is a best-effort request; when unavailable, stale data is preferable to total entity loss. Force refresh still works when explicitly requested.

**Why not escalate to force refresh on 502?** Force refresh sends a remote command to the car (waking it and drawing from the 12V battery). The default polling interval in Home Assistant is ~2 minutes. If every cached 502 triggered a force refresh, a sustained HATA outage would cause ~720 force refreshes per day vs the normal 1. Stale data for a few minutes is far better than draining the car battery.

### 3. `check_and_force_update_vehicle`: Force refresh when no data exists
When `last_updated_at is None` (first poll after HA restart), use force refresh instead of cached. Cached requests cannot populate empty vehicle state — if the cached endpoint returns 502, the vehicle would remain empty indefinitely. This only causes one extra force refresh at startup.

## Potentially Affected Issues

- kia_uvo #1651 — USA Genesis 502 on cached status (confirmed, this PR fixes it)
- kia_uvo #1208 — Genesis entities unavailable (closed stale, same root cause since July 2025)

**Note on [#1208](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1208#issuecomment-3733640111):** Logs from July 2025 show the identical HATA 502 response on `remoteVehicleStatus`. Users there created template sensor workarounds to show "last known" values when entities went `unavailable`. This PR eliminates the need for that workaround — cached 502 no longer causes entities to go unavailable. Additionally, PR #1161 (preserve last known value/unit when API returns None) prevents individual field values from being overwritten with `None` during partial API responses, which addresses the unit flip-flop issue (km → empty → km) also mentioned in #1208.

## Testing

13 new tests covering:
- `_check_response_for_errors`: 6 tests (int/string errorCodes, 502 never raises AuthenticationError, real HATA response)
- `update_vehicle_with_cached_state` 502 resilience: 4 tests (doesn't propagate, keeps existing data, success still works, auth errors still propagate)
- `check_and_force_update_vehicle`: 3 tests (no data → force refresh, within interval → cached, past interval → force refresh)

All 190 existing tests pass.

---

@cdnninja — could you review the reasoning here? Specifically: (1) the 502 on vehicle status is a server error, not auth, (2) the int/string key mismatch in your PR #1097 accidentally prevented the mapping from triggering, (3) whether there's a login-endpoint scenario where 502 should still map to AuthenticationError. I believe login 502s are already handled by HTTP status code checks rather than body errorCode, but want to make sure I'm not missing a case.

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
@@ -10,7 +10,11 @@ import certifi
 from requests.adapters import HTTPAdapter
 from urllib3.util.ssl_ import create_urllib3_context
 
-from hyundai_kia_connect_api.exceptions import APIError, AuthenticationError
+from hyundai_kia_connect_api.exceptions import (
+    APIError,
+    AuthenticationError,
+    ServiceTemporaryUnavailable,
+)
 
 from .ApiImpl import ApiImpl, ApiImplSession, ClimateRequestOptions
 from .const import (
@@ -42,21 +46,25 @@ def _check_response_for_errors(response: dict) -> None:
     """
     Checks for errors in the API response.
     If an error is found, an exception is raised.
-    Known values:
-    502: AuthenticationError - Incorrect username or password
+
+    The HATA backend returns errorCode as int (e.g. 502) in the
+    response body.  HTTP 502 from HATA is a server-side error
+    ("HATA remoteVehicleStatus service failed"), NOT an
+    authentication failure.  502 maps to ServiceTemporaryUnavailable;
+    other codes raise generic APIError.
 
     :param response: the API's JSON response
     """
     error_code_mapping = {
-        "502": AuthenticationError,
+        "502": ServiceTemporaryUnavailable,
     }
     if "errorCode" in response:
-        if response["errorCode"] in error_code_mapping:
-            raise error_code_mapping[response["errorCode"]](response["errorMessage"])
-        else:
-            raise APIError(
-                f"API Error {response['errorCode']}: {response['errorMessage']}"
-            )
+        system = response.get("systemName", "")
+        function = response.get("functionName", "")
+        suffix = f" [{system}/{function}]" if system or function else ""
+        code = str(response["errorCode"])
+        exc = error_code_mapping.get(code, APIError)
+        raise exc(f"API Error {code}{suffix}: {response['errorMessage']}")
 
 
 def _safe_parse_json(response, action_name: str):
--- a/tests/us_cached_502_test.py
+++ b/tests/us_cached_502_test.py
@@ -1,0 +1,171 @@
+"""Tests for HyundaiBlueLinkApiUSA cached status 502 handling.
+
+When the HATA backend returns 502 on cached vehicle status (REFRESH: false),
+update_vehicle_with_cached_state propagates the APIError so entities go
+unavailable per HA principles (data unknown, not stale). Success and auth
+errors are also covered.
+"""
+
+import datetime as dt
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from hyundai_kia_connect_api.HyundaiBlueLinkApiUSA import HyundaiBlueLinkApiUSA
+from hyundai_kia_connect_api.Token import Token
+from hyundai_kia_connect_api.Vehicle import Vehicle
+from hyundai_kia_connect_api.exceptions import APIError, AuthenticationError
+
+
+def _make_token():
+    return Token(
+        username="test@user.com",
+        password="pass",
+        access_token="at",
+        refresh_token="rt",
+        valid_until=None,
+        device_id="dev",
+        pin="1234",
+    )
+
+
+def _make_vehicle():
+    v = Vehicle()
+    v.id = "test-vehicle-id"
+    v.name = "Test Genesis"
+    v.enabled = True
+    return v
+
+
+def _make_api():
+    return HyundaiBlueLinkApiUSA.__new__(HyundaiBlueLinkApiUSA)
+
+
+class TestCachedStatus502Handling:
+    def test_cached_502_propagates(self):
+        """Cached vehicle status returning 502 should propagate APIError."""
+        api = _make_api()
+        token = _make_token()
+        vehicle = _make_vehicle()
+
+        api._get_vehicle_details = MagicMock(return_value={"odometer": 1000})
+        api._get_vehicle_status = MagicMock(
+            side_effect=APIError("API Error 502: server error")
+        )
+
+        with pytest.raises(APIError, match="502"):
+            api.update_vehicle_with_cached_state(token, vehicle)
+
+    def test_cached_502_leaves_vehicle_data_unchanged(self):
+        """When cached status fails, vehicle data is not updated (exception propagates)."""
+        api = _make_api()
+        token = _make_token()
+        vehicle = _make_vehicle()
+        vehicle._total_driving_range = 100.0
+        vehicle._total_driving_range_value = 100.0
+        vehicle._total_driving_range_unit = "km"
+
+        api._get_vehicle_details = MagicMock(return_value={"odometer": 1000})
+        api._get_vehicle_status = MagicMock(
+            side_effect=APIError("API Error 502: server error")
+        )
+        api._update_vehicle_properties = MagicMock()
+
+        with pytest.raises(APIError):
+            api.update_vehicle_with_cached_state(token, vehicle)
+
+        # _update_vehicle_properties never ran, so vehicle data is unchanged
+        api._update_vehicle_properties.assert_not_called()
+        assert vehicle.total_driving_range == 100.0
+
+    def test_cached_success_still_updates(self):
+        """When cached status succeeds, vehicle is updated normally."""
+        api = _make_api()
+        token = _make_token()
+        vehicle = _make_vehicle()
+
+        api._get_vehicle_details = MagicMock(return_value={"odometer": 1000})
+        api._get_vehicle_status = MagicMock(
+            return_value={"vehicleStatus": {"doorLock": True}}
+        )
+        api._get_ev_trip_details = MagicMock(return_value=None)
+        api._get_vehicle_location = MagicMock(return_value=None)
+        api._update_vehicle_properties = MagicMock()
+
+        api.update_vehicle_with_cached_state(token, vehicle)
+
+        api._update_vehicle_properties.assert_called_once()
+
+    def test_cached_auth_error_does_propagate(self):
+        """AuthenticationError should propagate (not be swallowed)."""
+        api = _make_api()
+        token = _make_token()
+        vehicle = _make_vehicle()
+
+        api._get_vehicle_details = MagicMock(return_value={"odometer": 1000})
+        api._get_vehicle_status = MagicMock(
+            side_effect=AuthenticationError("Auth failed")
+        )
+
+        with pytest.raises(AuthenticationError):
+            api.update_vehicle_with_cached_state(token, vehicle)
+
+
+class TestCheckAndForceUpdateVehicleNoData:
+    """When last_updated_at is None, cached update is used (no forced wake)."""
+
+    def test_no_data_uses_cached(self):
+        from hyundai_kia_connect_api.VehicleManager import VehicleManager
+
+        mgr = VehicleManager.__new__(VehicleManager)
+        vehicle = _make_vehicle()
+        mgr.vehicles = {vehicle.id: vehicle}
+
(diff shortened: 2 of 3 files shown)


cdnninja:
This would make old data show. Not sure we should have some regions act differently on failure.  This also stops the ha retry logic.

hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
```diff
@@ -43,21 +43,17 @@ def _check_response_for_errors(response: dict) -> None:
...
+    The HATA backend returns errorCode as int (e.g. 502) in the
+    response body.  HTTP 502 from HATA is a server-side error
+    ("HATA remoteVehicleStatus service failed"), NOT an
+    authentication failure.  There are no known errorCodes that
+    should map to AuthenticationError.
 
     :param response: the API's JSON response
     """
-    error_code_mapping = {
-        "502": AuthenticationError,
```
cdnninja, line 51: Would we just want to adjust the mapping? Potentially using text like the eu one?

cdnninja:
Please set to ready for review when you have had a chance on the above feedback.

hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
blka: The EU and USA APIs use different response formats, so the mapping approaches are fundamentally different and can't be unified:

- **EU** (`ApiImplType1._check_response_for_errors`): parses `resCode` (string keys like `"4002"`, `"7501"`, `"4004"`) from the `resMsg` envelope. No 502 exists in this mapping because the EU API doesn't return `resCode: 502` — EU server errors surface as HTTP status codes, not response body codes.

- **USA** (`HyundaiBlueLinkApiUSA._check_response_for_errors`): parses `errorCode` (int `502` or string `"502"` depending on endpoint) from the HATA error envelope (`errorCode`, `errorSubCode`, `systemName`, `errorMessage`). This is a completely different structure.

The HATA 502 response on `remoteVehicleStatus` is:
```json
{"errorCode": 502, "errorSubCode": "GEN", "systemName": "HATA", "errorMessage": "We're sorry, but we could not complete your request..."}
```

This is a transient server error, not an auth failure. No known HATA `errorCode` should map to `AuthenticationError` — the login endpoint handles auth failures via HTTP status codes, not `errorCode` in the body.

**The correct fix is to remove 502 from the mapping entirely**, not adjust it. The EU mapping doesn't have a 502 entry either. After this PR, the USA mapping becomes an empty dict (same effect as removing the method and inheriting the no-op from `ApiImpl`), which is correct because HATA error codes are not client-recoverable — they all raise generic `APIError`.

cdnninja: I am not looking for it to be unified rather using a similar approach.  This PR removes the base work that would allow for improved errors/exceptions.     

When you saw HATA error codes are not client recoverable what do you mean by that?   Exceptions allow us to either 1. Retry.  2. Stop retrying and force repair.  3.  Force re-auth.      By removing logic to pickup on error codes aren't we removing functionality?

cdnninja (changes requested):

hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
blka: Fair point — "not client-recoverable" was too strong. What I meant: the original `{"502": AuthenticationError}` was wrong because 502 on `remoteVehicleStatus` is a transient HATA backend failure, not an auth issue. But that doesn't mean the mapping structure itself is wrong — it's a useful hook for differentiating exception types so callers can retry / re-auth / surface appropriately, mirroring how `ApiImplType1` maps EU `resCode` to `DeviceIDError` / `RateLimitingError` / `ServiceTemporaryUnavailable` etc.

I restored the mapping with the one code we have confirmed from a real USA dump (jdhume's Genesis diagnostics): `502 → ServiceTemporaryUnavailable`. `ServiceTemporaryUnavailable` exists in `exceptions.py` exactly for this — transient backend. `update_vehicle_with_cached_state` still catches `APIError` (parent) so graceful-stale is unchanged, but the typed exception gives the coordinator a hook to retry transient backend errors differently from generic failures in the future.

Other HATA errorCodes (e.g. auth-related 401/403/1003/1005 that `KiaUvoApiUSA.request_with_logging` already handles on the Kia side) — I'm leaving unmapped until we have real USA dumps confirming their meaning on the Hyundai Blue Link backend. Speculatively mapping them to `AuthenticationError` risks unwanted re-auth loops if they mean something else here. The mapping dict is the hook; we expand it as dumps come in.

Also kept the `[systemName/functionName]` log suffix so diagnostic dumps distinguish HATA backend failures from other error sources.

hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
```diff
@@ -42,21 +46,31 @@ def _check_response_for_errors(response: dict) -> None:
...
+    can differentiate transient backend failures from other errors and
+    react accordingly (retry, re-auth, surface).  Unknown codes raise
+    generic APIError.  When systemName/functionName are present, they
+    are appended to the error message for diagnostics.
 
     :param response: the API's JSON response
     """
     error_code_mapping = {
-        "502": AuthenticationError,
+        502: ServiceTemporaryUnavailable,
```
cdnninja: Any reason both string and int?

hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
```diff
@@ -42,21 +46,31 @@ def _check_response_for_errors(response: dict) -> None:
     """
     Checks for errors in the API response.
     If an error is found, an exception is raised.
-    Known values:
-    502: AuthenticationError - Incorrect username or password
+
+    The HATA backend returns errorCode as int (e.g. 502) in the
+    response body.  HTTP 502 from HATA is a server-side error
+    ("HATA remoteVehicleStatus service failed"), NOT an
+    authentication failure.
```
cdnninja: This comment seems AI generated and I don't think is helpful, rather bloat in our code.

hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
```diff
@@ -840,7 +854,18 @@ def update_day_trip_info(
...
         state = {}
         state["vehicleDetails"] = self._get_vehicle_details(token, vehicle)
-        state["vehicleStatus"] = self._get_vehicle_status(token, vehicle, False)
+        try:
+            state["vehicleStatus"] = self._get_vehicle_status(token, vehicle, False)
+        except APIError as e:
+            # HATA backend may return 502 on cached vehicle status requests
+            # (REFRESH: false).  This is a server-side failure, not an auth
+            # or client error.  Keep existing vehicle data rather than
+            # propagating the exception and making all entities unavailable.
```
cdnninja: This is a change compared to other regions and general HA principals.  Unavailable shows the data isn't known vs incorrectly trying to show it is current when we know it isn't.

hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -177,7 +177,8 @@ def check_and_force_update_vehicle(
             else:
                 self.update_vehicle_with_cached_state(vehicle_id)
         else:
-            self.update_vehicle_with_cached_state(vehicle_id)
+            # No data yet — force refresh to get initial vehicle state
+            self.force_refresh_vehicle_state(vehicle_id)
```
cdnninja: Why would we start with a force refresh?  Most likely cloud has data so we should start with that. Forced calls consume battery so this now wakes the car on every HA restart.  That seems concerning.

hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
cdnninja: I would also think this should be within the _get_vehicle_status method?  Since that way it is universally handled and not just when used here.

cdnninja (changes requested):

hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
blka: Agreed on both. Removed the graceful-stale try/except from `update_vehicle_with_cached_state` — on HATA 502 the `APIError` (`ServiceTemporaryUnavailable`) now propagates, entities go unavailable per HA principles, and recover on the next successful poll or force-refresh cycle. Since the graceful-stale is gone, the "move it into `_get_vehicle_status`" point is moot — there's nothing to relocate.

hyundai_kia_connect_api/VehicleManager.py
blka: Agreed — reverted. `check_and_force_update_vehicle` now uses `update_vehicle_with_cached_state` when `last_updated_at is None` (original behavior). Cloud cache is likely populated on restart, cached fetch is cheap, and we don't wake the car on every HA restart. If cached fails, entities go unavailable and the next force-refresh cycle (or manual button) recovers.

hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
blka: Normalized to a single string key — `code = str(response["errorCode"])` with `{"502": ServiceTemporaryUnavailable}`. Handles both int (jdhume's dump) and string via normalization, matches the EU pattern (string `resCode` keys), drops the dual-key bloat.

hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
blka: Trimmed. Removed the "Known errorCodes are mapped..." and "When systemName/functionName..." paragraphs — the code is self-evident. Kept the HATA 502 note + concise mapping reference.

PR #1164, opened 2026-06-01
feat: Add fields for high and bifunc headlamp status and add light status readout to new dataformat handler

Triple-S:
Hello,

I noticed that my MY 2023 Hyundai IONIQ 5 was reporting also the status of the high and bifunc headlamps which were not yet implemented in this API. In addition I noticed that my father's MY 2025 Hyundai IONIQ 5 which uses the new dataformat did not show any lamp status although the information were present in the reported data.
This PR adds both functionalities by introducing additional fields for the information that were not yet covered. I introduced them into all regional APIs that already had the low headlamp status although I could only verify the presence of the datafields for the EU version. Maybe someone from one of these other regions can verify that these fields also exist there.

Best regards,
Triple-S

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -312,6 +312,40 @@ class ApiImplType1(ApiImpl):
             get_child_value(state, "Cabin.Seat.Row2.Right.Climate.State")
         )
 
+        vehicle.headlamp_status = get_child_value(
+            state, "Body.Lights.Front.HeadLamp.SystemWarning"
+        )
+        vehicle.headlamp_left_low = get_child_value(
+            state, "Body.Lights.Front.Left.Low.Warning"
+        )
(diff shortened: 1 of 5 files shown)


cdnninja:
Each region may use different names.  So I would think for now please add for regions we know.   You may be able to search the value in the kia_uvo repo to see logs for other regions to see if it exists.

Triple-S:
@cdnninja Yes, that is why I added the new fields only to regional APIs where the fields `lampWireStatus.headLamp.leftLowLamp` and `lampWireStatus.headLamp.rightLowLamp` were already existing before. I think if these two are existing and have the same name it is very reasonable to assume that also the new `...leftHighLamp`, `...rightHighLamp`, `...leftBifuncLamp` and `...rightBifuncLamp` are existing and have the same name. Regional APIs that did not already have `...leftLowLamp` and `...rightLowLamp` were left unchanged. The only exception is the new format API where I had the data from my father's car as reference.

PR #1177, opened 2026-06-16
feat(CA): add Cloudflare __cf_bm cookie for bot management bypass

### LINKED ISSUE #1131: Hyundai Canada login blocked by Cloudflare 403 / 1015 rate limit before JSON response
Hyundai Canada login is blocked by Cloudflare for hyundai-kia-connect-api 4.11.1.

Region: Canada
Brand: Hyundai
Vehicle: 2026 IONIQ 5
Endpoint: https://mybluelink.ca/tods/api/v2/login

Tested from:
- Home Assistant kia_uvo on Raspberry Pi / DietPi
- bluelink CLI on macOS
- bluelink CLI over phone hotspot / different public IP

Results:
- HTTP 403 text/html Cloudflare “Attention Required” page from CLI
- Cloudflare Error 1015 / status 429 rate_limited JSON from Home Assistant after retries
- kia_uvo then fails with KeyError: 'responseHeader' because the response is Cloudflare JSON/HTML, not Hyundai Canada login JSON

This appears to happen before OTP/MFA handling.

blka:
## Problem

Hyundai Canada API is protected by Cloudflare bot management. Requests without the `__cf_bm` cookie get blocked with HTTP 403 / rate limit 1015, as reported in #1131.

## Solution

Fetch the `__cf_bm` cookie from the `/login` page and include it in all authenticated API requests:

- **`_get_cloudflare_cookie()`** — GET `https://{apiHost}/login`, extract `__cf_bm` from `Set-Cookie` header
- **`_ensure_cloudflare_cookie()`** — proactive refresh at 25 min (cookie expires ~30 min)
- **`_add_cloudflare_cookie(headers)`** — inject cookie into request headers
- **`_is_cloudflare_block(response)`** — detect 403 with Cloudflare HTML/keywords
- **`_post_with_cloudflare_retry(url, **kwargs)`** — POST wrapper that catches 403 Cloudflare blocks, refreshes cookie, retries

### Bug fixes found during implementation

7 instances of `headers = self.API_HEADERS` without `.copy()` — this permanently mutates the shared dict when headers are modified downstream. Fixed to `headers = self.API_HEADERS.copy()`.

## Testing

16 unit tests covering:
- Cookie fetch from login page
- Cookie injection in login flow
- Proactive refresh (25 min threshold)
- Cloudflare 403 detection (HTML body, "cloudflare" keyword)
- POST retry on Cloudflare block
- Non-Cloudflare 403 not triggering retry

Closes #1131

cc @paul-e-m @kiluazen — this should resolve the Cloudflare 403 blocks you were seeing. Would appreciate if you could test with your CA accounts.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -137,6 +137,8 @@ class KiaUvoApiCA(ApiImpl):
             "client_secret": "CLISCR01AHSPA",
         }
         self._sessions = None
+        self._cloudflare_cookie: str = ""
+        self._cloudflare_cookie_fetched_at: dt.datetime | None = None
 
     def _get_device_id(self) -> str:
         """Generate a deterministic device ID based on MAC address and hostname.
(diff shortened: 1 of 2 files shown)


cdnninja:
I haven't seen reports of this issue for awhile.  Do you have someone currently impacted?

blka:
Heads-up on something worth checking alongside this PR — CA currently has a second, separate retry mechanism (`RetrySession` from #857) that this PR would end up stacked on top of:

- `RetrySession` (existing, #857): `max_retries=4`, backoff 2/4/8/16s, retries ONLY `ConnectionError` on **POST** — which includes control commands (lock/unlock/remote-start).
- This PR adds `_post_with_cloudflare_retry`: on 403 Cloudflare block → refresh `__cf_bm` → retry once.

Two retry layers, different failure modes (transport reset vs edge/403 block). Stacking both on a single POST = up to 4× ConnectionError retries + 1× cookie retry, which is messy and adds up to ~30s latency on persistent failure.

The bigger concern: `RetrySession` retries POST on `ConnectionError`, but `requests` raises `ConnectionError` for both connect-failure (request never sent — safe to retry) and response-read-failure (request reached the server, response dropped — unsafe to replay for non-idempotent control POSTs). `RetrySession` can't distinguish the two, so it can replay lock/unlock/remote-start commands. This is the same "no replay of non-idempotent POSTs" principle @Lekensteyn applied in #1160 (where 5xx retry was removed for exactly this reason).

Worth checking: was #857's CA login error 104 / ConnectionError actually Cloudflare-block-driven? If yes, this cookie PR fixes the root cause and `RetrySession` becomes dead code — then it could be removed (or trimmed to login-only) here or in a follow-up, which would also let CA migrate to `ApiImplSession` (#1160, the one region currently without a default timeout — closes #1151 for CA). If 104 was a genuine transport reset unrelated to Cloudflare, the cookie doesn't cover it and the retry still has a job.

@cdnninja — given you're not seeing recent reports of the Cloudflare issue, could you check whether the ConnectionErrors that motivated #857 are still occurring in the field? That determines whether `RetrySession` is still earning its keep or whether this cookie PR supersedes it. No change requested in this PR — just flagging the overlap so we can decide whether to consolidate the two retry paths.

cdnninja:
If users aren't reporting it to kia_uvo I doubt it's a current issue.

cdnninja:
If users aren't reporting it to kia_uvo I doubt it's a current issue.

cdnninja:
Can this one be closed?

Issue #190, opened 2022-12-16
CA no seat heating function


404alex:
* Hyundai / Kia Connect version: 1.45.6
* Python version: 3.9
* Operating System: HAOS

### Description

I was trying to get the seat heating work for the start_climate function. 

### What I Did

I changed the KiaUvoApiCA.py file line number 495 to the following and it worked. (Non EV)
Steeling wheel heating: On
Front and back defrost: On
Driver Seat: HI heat, All other off.

Was using burp to capture the API call.

```python
payload = {
            "setting": {
                "airCtrl": int(climate),
                "defrost": defrost,
                "heating1": 4,
                "igniOnDuration": duration,
                "ims": 0,
                "airTemp": {"value": set_temp, "unit": 0, "hvacTempType": 1},
                "seatHeaterVentCMD":{"drvSeatOptCmd":8, "astSeatOptCmd":2, "rlSeatOptCmd":2, "rrSeatOptCmd":2},
            },
            "pin": self.pin,
        }
```

Any chance it could be added to the API? I usually just change the KiaUvoApiCA.py each time when this pip package get updates. 

Thanks!

cdnninja:
Yes we have another issue opened for this.   All the basics exist.  I just need to test it by sending heated seats will it break my car as mine doesn't support that via remote control.   It changes if I can lump it into all payloads or need to do an if statement with unique payload. 

I'll see if I can do that today if not my wife has the Kia this weekend away. 

Other than that this will be quick.

cdnninja:
Looks like you have an EV so it is separate from my cars logic today as start is different for reach. Can you test the latest release and let me know if it works for seats? 

As a note the values are explained in https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/master/hyundai_kia_connect_api/const.py

404alex:
Hi cdnninjia,

No mine is ICE, 2021 Hyundai Elantra, I will test it and let you know, thanks!

cdnninja:
Interesting!   This whole time I thought when "setting" was used it was an EV in the payload as an EV won't start with my payload of a 2019 sorento ICE.  This logic can be seen here: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/f943c17b0fa971e59fb8b22251fd714e788bc5bb/hyundai_kia_connect_api/KiaUvoApiCA.py#L488-L514

[...]

PR #1183, opened 2026-06-16
feat(EU/AU/IN/CN): CCS2 climate seat retry + dynamic drvSeatLoc

blka:
### 1. CCS2 climate seat retry

When CCS2 `start_climate` fails with `retCode=="F"` and `seatClimateInfo` is in the payload, remove seat settings and retry. Some vehicles do not support heated/vented seats remotely — the retry ensures climate still starts without them.

Matches `egmp-bluelink-scriptable` behavior.

### 2. Dynamic drvSeatLoc

`drvSeatLoc` was hardcoded to `"R"` (RHD). Now uses `"L"` for km-based vehicles (most of EU: Germany, France, Spain, Poland, etc.) and `"R"` for mile-based vehicles (UK, AU, IN).

Based on `vehicle.distance_unit`: `DISTANCE_UNITS[1]` (km) → `"L"`, otherwise → `"R"`.

## Files changed

- `ApiImplType1.py` — CCS2 retry logic + drvSeatLoc dynamic
- `tests/test_eu_ccs2_climate_retry.py` — 6 unit tests (4 seat retry + 2 drvSeatLoc)

## Testing

- 186 unit tests pass
- Live tested on EU Hyundai Santa Fe 2026 (CCS2=1, km) — drvSeatLoc="L" verified

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -1011,6 +1011,8 @@ class ApiImplType1(ApiImpl):
                 + vehicle.id
                 + "/ccs2/control/temperature"
             )
+            # drvSeatLoc: "L" for LHD (km), "R" for RHD (miles)
+            drv_seat_loc = "L" if vehicle.distance_unit == DISTANCE_UNITS[1] else "R"
             payload = {
                 "command": "start",
                 "ignitionDuration": options.duration,
@@ -1018,7 +1020,7 @@ class ApiImplType1(ApiImpl):
                 "hvacTempType": 1,
                 "hvacTemp": options.set_temp,
                 "sideRearMirrorHeating": 1,
-                "drvSeatLoc": "R",
+                "drvSeatLoc": drv_seat_loc,
                 "seatClimateInfo": {
                     "drvSeatClimateState": options.front_left_seat,
                     "psgSeatClimateState": options.front_right_seat,
@@ -1034,6 +1036,19 @@ class ApiImplType1(ApiImpl):
                 json=payload,
                 headers=self._get_control_headers(token, vehicle),
             ).json()
+            # CCS2 retry: if seat settings caused a failure, retry without them
+            if response.get("retCode") == "F" and "seatClimateInfo" in payload:
+                _LOGGER.warning(
+                    f"{DOMAIN} - CCS2 climate start with seat settings failed "
+                    f"(resCode={response.get('resCode')}), "
+                    f"retrying without seat settings"
+                )
+                del payload["seatClimateInfo"]
+                response = requests.post(
+                    url,
+                    json=payload,
+                    headers=self._get_control_headers(token, vehicle),
+                ).json()
         _LOGGER.debug(f"{DOMAIN} - Start Climate Action Response: {response}")
         _check_response_for_errors(response)
         token.device_id = self._get_device_id(self._get_stamp())
--- a/tests/test_eu_ccs2_climate_retry.py
+++ b/tests/test_eu_ccs2_climate_retry.py
@@ -1,0 +1,284 @@
+"""Tests for CCS2 start_climate seat retry and drvSeatLoc in ApiImplType1.
+
+CCS2 climate path: if seat settings cause a failure (retCode=="F"),
+retry without seatClimateInfo. drvSeatLoc is dynamic based on
+vehicle.distance_unit (km→"L", mi→"R").
+"""
+
+import pytest
+from unittest.mock import MagicMock, patch
+
+from hyundai_kia_connect_api.ApiImpl import ClimateRequestOptions
+from hyundai_kia_connect_api.KiaUvoApiEU import KiaUvoApiEU
+from hyundai_kia_connect_api.Token import Token
+from hyundai_kia_connect_api.Vehicle import Vehicle
+from hyundai_kia_connect_api.const import DISTANCE_UNITS, ENGINE_TYPES
+from hyundai_kia_connect_api.exceptions import APIError, DuplicateRequestError
+
+
+@pytest.fixture
+def eu_api():
+    return KiaUvoApiEU(region=1, brand=2, language="en")
+
+
+@pytest.fixture
+def ccs2_vehicle():
+    v = Vehicle()
+    v.id = "test-vehicle-id"
+    v.engine_type = ENGINE_TYPES.PHEV
+    v.ccu_ccs2_protocol_support = 1
+    v.distance_unit = DISTANCE_UNITS[1]  # km → LHD
+    return v
+
+
+@pytest.fixture
+def token():
+    t = Token()
+    t.access_token = "Bearer test-access-token"
+    t.pin = "1234"
+    t.device_id = "test-device-id"
+    return t
+
+
+@pytest.fixture
+def climate_options_with_seats():
+    opts = ClimateRequestOptions()
+    opts.climate = True
+    opts.set_temp = 22
+    opts.duration = 10
+    opts.heating = 0
+    opts.defrost = False
+    opts.steering_wheel = 0
+    opts.front_left_seat = 3
+    opts.front_right_seat = 2
+    opts.rear_left_seat = 0
+    opts.rear_right_seat = 0
+    return opts
+
+
+class TestCCS2ClimateSeatRetry:
+    def test_retry_without_seats_on_failure(
+        self, eu_api, token, ccs2_vehicle, climate_options_with_seats
+    ):
+        """Should retry without seat settings when CCS2 returns retCode=F."""
+        call_count = 0
+
+        def mock_post(url, **kwargs):
+            nonlocal call_count
+            call_count += 1
+            resp = MagicMock()
+            if call_count == 1:
+                resp.json.return_value = {
+                    "retCode": "F",
+                    "resCode": "4005",
+                    "resMsg": "Unsupported control",
+                }
+            else:
+                resp.json.return_value = {
+                    "retCode": "S",
+                    "resCode": "0000",
+                    "resMsg": {},
+                    "msgId": "retry-msg-id",
+                }
+            return resp
+
+        with (
+            patch(
+                "hyundai_kia_connect_api.ApiImplType1.requests.post",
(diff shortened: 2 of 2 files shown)


cdnninja:
Do you have a reference for issue 1? I am good with dynamic seatloc but these are in on PR.

blka:
Re: issue references for the CCS2 seat retry + live test results.

## Issue references for CCS2 seat retry

- **kia_uvo https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1007** (strongest) — @badnbusy confirms: _"If you have a new car (2023+) and in the EU then the current integration doesn't work for heated seats. Myself and @cdnninja are working on getting this working."_ EU Hyundai Kona, climate works but seat + steering wheel don't start.
- kia_uvo https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/718 — USA Kia K5 GT heated/vented seats not functional/not sent
- kia_uvo https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/775 — 2024 Kia Sportage "Incorrect request payload format" (errorCode 9001) with heating accessory
- API #190 — CA no seat heating function

## Live test on EU Hyundai Santa Fe 2026 (CCS2=1, km, PL)

I ran the PR branch against my own car. Key findings:

**CCS2 retry logic fires correctly** — log confirms:
```
CCS2 climate start with seat settings failed (resCode=4005), retrying without seat settings
```

**But the retry doesn't help on this vehicle.** Both paths (with and without `seatClimateInfo`) return:
```json
{"retCode": "F", "resCode": "4005", "resMsg": "Unsupported control request - Vehicle is not Supported"}
```

So `ccs2/control/temperature` is rejected entirely for this 2026 Santa Fe — the failure is NOT caused by the seat settings. The cached state shows the car does have seat climate (`Seat.Row1.Driver.Climate.State: 3` = Low Cool) and `Service.ConnectedCar.RemoteControl.Available: 1`, so the feature exists physically. The myHyundai iOS app just gained remote seat ventilation options in a recent update, which means the app is using a **different endpoint** than `ccs2/control/temperature` for this model year (likely the newer CCI API).

Net: the retry is defensive and harmless, but on at least one current CCS2 vehicle it masks a deeper problem — the endpoint itself is no longer accepted. The retry still has value for older CCS2 vehicles where seat settings genuinely cause a `retCode=="F"` (matching the `egmp-bluelink-scriptable` behavior it was modeled on), so I'd keep it — just flagging that it's not a complete fix for 2025+ models.

## drvSeatLoc bug (separate concern — agree it should be a separate PR)

The `drvSeatLoc` change has a bug that crashes `start_climate`:

```python
drv_seat_loc = "L" if vehicle.distance_unit == DISTANCE_UNITS[1] else "R"
```

`Vehicle` has **no `distance_unit` attribute** — that field exists only on `DailyDrivingStats`. On a fresh `update_all_vehicles_with_cached_state()`, `start_climate` raises:

```
AttributeError: 'Vehicle' object has no attribute 'distance_unit'
```

I had to monkeypatch `vehicle.distance_unit` in my live test to get past it. `Vehicle` does have `odometer_unit`, but that's also only populated by `get_driving_info`, not cached state — so it's `None` after a normal poll and the heuristic resolves to `"R"` by default (wrong for EU LHD).

Two problems with the heuristic itself (km→L, miles→R):
- **AU** uses km but is RHD → PR gives "L", should be "R"
- **IN** uses km but is RHD → PR gives "L", should be "R"

Distance unit is not a reliable LHD/RHD proxy. The correct signal is region/country. I'll split drvSeatLoc into its own PR and use a region/country-based mapping (or a `drvSeatLoc`/`steeringWheelLocation` field from the vehicle state if one exists in the CCS2 response).

## Plan

Agreed on splitting. I'll open:
1. **PR (seat retry)** — keep the CCS2 retry logic only, with the issue references above.
2. **PR (drvSeatLoc)** — fix the `AttributeError` and replace the distance-unit heuristic with a region/country-based LHD/RHD mapping.

Want me to keep the seat-retry PR as-is (just the retry), or also add a note that 2025+ vehicles need a different endpoint?

blka:
Closing this PR. Documenting the full investigation here so the findings survive.

## Decision: drop the CCS2 seat retry, keep only a drvSeatLoc fix in a separate PR

Per @cdnninja's two points (issue reference for the retry + split the PR), I went and verified whether the seat-retry case actually exists in the wild. Conclusion: **no user has reported the specific CCS2 seat-retry failure in our tracker, so I'm not adding speculative defensive code.** drvSeatLoc goes in its own PR.

## Issue references (what exists in the trackers)

- **kia_uvo #1007** — strongest, but **died of natural causes**. Timeline:
  - 2025-01-07: opened (EU Hyundai Kona EV 2020, PL)
  - 2025-01-09: @badnbusy: _"If you have a new car (2023+) and in the EU then the current integration doesn't work for heated seats. Myself and @cdnninja are working on getting this working"_
  - **2025-01-12**: @cdnninja opened **PR #705 "fix: EU Seats"** — CCS2 payload with hardcoded `drvSeatLoc:"R"` (RHD, wrong for EU LHD) and all seats hardcoded to level 6 — **closed the same day without merge, no body, no comments**
  - 2025-03-13: #1007 auto-closed by the stale bot (cdnninja never commented on the issue)
  - Note: #1007's car is a 2020 Kona on the **old CCAPI path**, not CCS2 — so it doesn't validate the CCS2 retry either.
- kia_uvo #718 — USA Kia K5 heated/vented seats not sent (USA `heatVentSeat` format, different endpoint)
- kia_uvo #775 — 2024 Kia Sportage "Incorrect request payload format" (USA)
- API #190 — CA no seat heating function (CA)

**No issue in either tracker reports a CCS2 `ccs2/control/temperature` call failing with `retCode:"F"` specifically because of `seatClimateInfo`, where retrying without seats succeeds.**

## Reference implementation: egmp-bluelink-scriptable

The retry was modeled on `egmp-bluelink-scriptable`. I checked its tracker and source:

- **egmp-bluelink-scriptable [#28](https://github.com/andyfase/egmp-bluelink-scriptable/issues/28)** — Kia EV6 (US): climate fails when seat climate is included; that EV6 doesn't support cooled seats at all. Owner @andyfase shipped a retry (removing seat settings on failure) in v1.6.0.
- **egmp `src/lib/bluelink-regions/europe.ts`** does have a retry on the EU CCS2 path, on the **same endpoint we use** (`/api/v2/spa/vehicles/{id}/ccs2/control/temperature`):
  ```ts
  if (this.requestResponseValid(resp.resp, resp.json).valid) { /* poll */ }
  else {
    // Kia/Hyundai US seems pretty particular with seat settings, hence if fail retry without them,
    if (!retryWithNoSeat && climateRequest.seatClimateInfo) {
      delete climateRequest.seatClimateInfo
      return this.climateStartStop(id, climateRequest, true)
    }
  }
  ```
  Trigger: `requestResponseValid` checks **HTTP status** (200/204/302 = valid). So egmp retries on HTTP-level failure, **not** on `retCode:"F"` in a 200 body.

### Why this matters — our retry fired on the wrong thing

Our PR #1183 retried on `response.get("retCode") == "F"`. Korean connected-car APIs return errors as **HTTP 200 with the error in the body** (that's why `_check_response_for_errors` parses the body, not `status_code`). So our trigger fires on `resCode 4005` ("Vehicle is not Supported"), where retry is pure waste — a doomed second POST every time.

I confirmed this on my own car (EU Hyundai Santa Fe 2026, CCS2=1, PL):

```
Start Climate Action Response: {'retCode': 'F', 'resCode': '4005', 'resMsg': 'Unsupported control request - Vehicle is not Supported'}
WARNING - CCS2 climate start with seat settings failed (resCode=4005), retrying without seat settings
[second POST → same 4005]
```

egmp's HTTP-status trigger would **not** retry on this 4005 (HTTP 200 = valid) — so egmp also fails on this vehicle, just without the wasted second call.

## Root cause of the 4005 (separate from seat settings)

The 2026 Santa Fe rejects `ccs2/control/temperature` (CCAPI v2, host `prd.eu-ccapi.hyundai.com`) entirely with 4005 — both with and without `seatClimateInfo`. The car does have seat climate (`Seat.Row1.Driver.Climate.State: 3`, `Service.ConnectedCar.RemoteControl.Available: 1`) and the myHyundai iOS app just gained remote seat ventilation options, so the app uses a **different endpoint** for this model year:

`POST https://gspa-ccs-eu.hyundai.com/gspa/v1/remote/vehicles/{carId}/temperature` (GSPA path, async with `{SID,svcSID,svcTime}` polling). 

The CCAPI v2 `ccs2/control/temperature` path is being deprecated/rejected for 2025+ models. That's a GSPA/CCI-API migration topic, out of scope here.

## drvSeatLoc bug (the part worth keeping)

The drvSeatLoc change crashes `start_climate`:

```python
drv_seat_loc = "L" if vehicle.distance_unit == DISTANCE_UNITS[1] else "R"
```

`Vehicle` has **no `distance_unit` attribute** (that field is on `DailyDrivingStats` only). On a fresh `update_all_vehicles_with_cached_state()`, `start_climate` raises:

```
AttributeError: 'Vehicle' object has no attribute 'distance_unit'
```

`Vehicle` does have `odometer_unit`, but it's also only populated by `get_driving_info`, not cached state, so it's `None` after a normal poll. The km→L / miles→R heuristic is also wrong for **AU and IN** (both km but RHD). egmp handles this correctly: `europe.ts` uses `distanceUnit === 'mi' ? 'R' : 'L'`, while `australia.ts` hardcodes `drvSeatLoc: 'R'` (RHD regardless of unit).

## What I'll do instead

1. **Drop the CCS2 seat retry** — no reported case in our tracker; not adding speculative code. egmp's HTTP-status-based retry is a different trigger and doesn't justify a body-`retCode`-based retry that wastes a call on 4005.
2. **Open a separate PR for drvSeatLoc** — fix the `AttributeError` and use a correct LHD/RHD signal (region-based defaults with AU/IN hardcoded RHD, mirroring egmp), instead of the broken `distance_unit` heuristic.

Closing this PR now.

blka:
Closing in favor of a drvSeatLoc-only PR. Full investigation documented above. Seat retry dropped — no reported case in our tracker.

PR #1184, opened 2026-06-16
feat(EU/AU/IN/CN): CCS2 force refresh polling loop

blka:
## Problem

`_force_refresh_vehicle_state_ccs2()` (EU/AU/IN/CN) calls `GET /ccs2/carstatus` — the **wake** endpoint (no `/latest`) — and reads `response["resMsg"]["state"]["Vehicle"]` synchronously. That endpoint returns an async command envelope `{retCode, resCode, msgId}`, **not** vehicle state, so it raises `KeyError: 'resMsg'` every time Force Refresh is triggered. Reported in kia_uvo #1786 (3.7.0/3.7.1) and again in **kia_uvo #1806** (IONIQ 5, CCS2, kia_uvo 3.8.0 / hyundai_kia_connect_api 4.25.2 — i.e. still present in the current release). The prior cached-only path (`/latest`) returned stale data and never woke the car.

## Solution

Trigger + poll, refactored into base `ApiImplType1` with 3 polymorphic hooks so EU/AU/IN/CN share one loop:

1. **Trigger** — `GET /ccs2/carstatus` (no `/latest`) wakes the vehicle (same request the current code already makes).
2. **Poll** — `GET /ccs2/carstatus/latest` every 10s, max 14 attempts (~140s window) via `_inspect_ccs2_response`. **Early-exit on fresh** — stops the moment `last_updated > trigger_time`.
3. **Freshness** — apply state only when `last_updated > trigger_time` (HA: do not mask stale).
4. **Timeout** — if no fresh data after 14 polls (~140s), keep previous vehicle state untouched + warning log (no stale masking).
5. **Location** — `_post_refresh_ccs2_location` after a successful refresh.

The **logic and timing mirror what the official app already does**: wake the vehicle, then receive fresh status. The lib polls `/latest` as the HTTP equivalent of that status delivery, with the same ~140s upper bound the app allows for a slow/deep-sleep car to report, and a 10s poll cadence. (All requests use the access token — no PIN/control token per poll.)

## Request budget vs current code (acceptability)

Current code (master / 4.25.2): **1 GET** `/ccs2/carstatus` → `KeyError` (no fresh data returned).

This PR: 1 trigger GET (same as today) + N poll GETs on `/latest`:

| Scenario | Extra GETs vs today | Total | Wall-clock |
|---|---|---|---|
| **Typical (car reachable)** | **+2** | 3 (1 trigger + 2 polls) | ~20s, then early-exit |
| **Max (car unreachable / deep-sleep)** | **+14** | 15 (1 trigger + 14 polls) | ~140s, then timeout |

- **Typical +2 requests** is the actual cost of returning fresh data — today the call returns none (it crashes). The loop stops the instant fresh data arrives (early-exit), so a reachable car costs 2 polls and ~20s, full stop.
- **Max +14 requests** only occurs when the car cannot report at all (the case where today's code fails immediately anyway). 14 polls spread over ~140s = **~0.11 req/s** — a gentle cadence far below any server rate limit; on timeout it keeps the previous state (defined, bounded retry) instead of raising `KeyError`.
- The ~140s cap and 10s cadence match the app's own timing for a slow-to-report vehicle, so the upper bound is not a number invented here — it's what the app already tolerates.

### Hooks per region

| Region | `_inspect_ccs2_response` | `_apply_ccs2_state` | `_post_refresh_ccs2_location` |
|---|---|---|---|
| EU | base (`resMsg.state.Vehicle` + `Date` UTC) | base (CCS2 parser) | override (`/location/park`) |
| AU | base (inherits) | base (inherits) | base (`_get_location` + inline) |
| IN | override (`resMsg` root + `state.time`) | override (IN parser) | override (`_update_vehicle_location`) |
| CN | override (`resMsg` root + `status.time`) | override (CN parser) | base (inherits) |

## Files changed

- `ApiImplType1.py` — base `_force_refresh_vehicle_state_ccs2` + 3 hooks
- `KiaUvoApiEU.py` — remove old override, add `_post_refresh_ccs2_location` (park)
- `KiaUvoApiAU.py` — remove old override (inherits base)
- `KiaUvoApiIN.py` — `_inspect` + `_apply` + `_post_refresh` hooks
- `KiaUvoApiCN.py` — `_inspect` + `_apply` hooks
- `tests/test_{eu,au,in}_ccs2_force_refresh.py`, `tests/test_cn_force_refresh_nonccs2.py` — 4 new test files

## Testing

- Unit tests pass for the polling loop (fresh, timeout-keeps-previous-state, budget 14×10s), hooks, EU park-location, IN resMsg-root, AU inheritance, CN hooks + non-CCS2 characterization.
- **Live-validated on EU Hyundai Santa Fe 2026 (CCS2), car reachable**: trigger a state change (lock, 0.4s) → `force_refresh_vehicle_state` → **fresh `Date > trigger_time` detected after 2 polls = 20.7s** (early-exit; the 140s window is unused for the typical reachable case). State applied.
- IN/CN timezone interpretation of `status.time`/`state.time` is pre-existing and documented in the tests; deferred pending a dump confirming UTC vs wall-clock.

Rebased onto current upstream/master (`65b4464`), 1 commit. Fixes kia_uvo #1786 and the newly-filed duplicate kia_uvo #1806.

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -8,6 +8,7 @@ import threading
 
 import time
 from time import sleep
+from typing import Any
 
 from .ApiImpl import (
     ApiImpl,
@@ -797,6 +798,102 @@ class ApiImplType1(ApiImpl):
         # self._sanitize_ice_value(state)
 
         vehicle.data = state
+
+    def _inspect_ccs2_response(
+        self, response: dict[str, Any]
+    ) -> tuple[dict[str, Any] | None, dt.datetime | None]:
+        """Extract (state, last_updated) from a CCS2 /latest response.
+
+        Default impl for EU/AU shape: resMsg.state.Vehicle + Date (UTC).
+        Does NOT call the parser — caller checks freshness before applying.
+        """
+        try:
+            state: dict[str, Any] | None = response["resMsg"]["state"]["Vehicle"]
+        except (KeyError, TypeError):
+            return None, None
+        if state is None:
+            return None, None
+        raw_date = get_child_value(state, "Date")
+        if not raw_date:
+            return state, None
+        try:
+            last = parse_datetime(str(raw_date).split(".")[0], dt.timezone.utc)
+            return state, last.astimezone(self.data_timezone)
+        except (ValueError, TypeError):
+            _LOGGER.warning(f"{DOMAIN} - CCS2 _inspect: cannot parse Date '{raw_date}'")
+            return state, None
+
+    def _apply_ccs2_state(self, vehicle: Vehicle, state: dict[str, Any]) -> None:
+        """Parse + write vehicle state. Called only when data is fresh.
+
+        Default impl: CCS2 parser (EU/AU).
+        """
+        self._update_vehicle_properties_ccs2(vehicle, state)
+
+    def _post_refresh_ccs2_location(self, token: Token, vehicle: Vehicle) -> None:
+        """Refresh vehicle location after a CCS2 force-refresh.
+
+        Default impl: _get_location + inline set (AU/CN shape).
+        """
+        location = self._get_location(token, vehicle)
+        if location and get_child_value(location, "coord.lat"):
+            vehicle.location = (
+                get_child_value(location, "coord.lat"),
+                get_child_value(location, "coord.lon"),
+                parse_datetime(get_child_value(location, "time"), self.data_timezone),
+            )
+
+    def _force_refresh_vehicle_state_ccs2(self, token: Token, vehicle: Vehicle) -> None:
+        """Force-refresh CCS2 vehicle state: trigger, poll until fresh, apply.
+
+        1. GET /ccs2/carstatus (no /latest) to wake the vehicle.
+        2. Poll /ccs2/carstatus/latest every 10s, max 14 attempts (~140s window,
+           matching the app's controlRetrofit readTimeout = backend max
+           wake-to-report), until last_updated > trigger_time.
+        3. Apply state only when fresh (HA: do not mask stale).
+        4. Refresh location.
+        """
+        headers = self._get_authenticated_headers(
+            token, vehicle.ccu_ccs2_protocol_support
+        )
+        trigger_url = self.SPA_API_URL + "vehicles/" + vehicle.id + "/ccs2/carstatus"
+        self.session.get(trigger_url, headers=headers).json()
+        trigger_time = time.time()
+
+        poll_url = trigger_url + "/latest"
+        # Window ~140s = app controlRetrofit readTimeout (backend max wake for a
+        # slow/deep-sleep car). Early-exit on fresh => typical ~20s unchanged;
+        # 140s matters only for slow wakes. 10s interval => 14 calls max
+        # (server-friendly; HA force_refresh has no HA-side timeout cap).
+        max_attempts = 14
+        poll_interval = 10
+        fresh_state: dict[str, Any] | None = None
+
+        for attempt in range(1, max_attempts + 1):
+            sleep(poll_interval)
+            response: dict[str, Any] = self.session.get(
+                poll_url, headers=headers
+            ).json()
+            state, last_updated = self._inspect_ccs2_response(response)
+            if (
+                state is not None
+                and last_updated is not None
+                and last_updated.timestamp() > trigger_time
+            ):
+                fresh_state = state
+                _LOGGER.debug(
+                    f"{DOMAIN} - CCS2 force refresh: fresh after {attempt} polls"
+                )
+                break
+
+        if fresh_state is not None:
+            self._apply_ccs2_state(vehicle, fresh_state)
+            self._post_refresh_ccs2_location(token, vehicle)
+        else:
+            _LOGGER.warning(
+                f"{DOMAIN} - CCS2 force refresh: no fresh data after "
+                f"{max_attempts} polls, keeping previous vehicle state"
+            )
 
     @_retry_on_device_id_error
     def start_charge(self, token: Token, vehicle: Vehicle) -> str:
--- a/hyundai_kia_connect_api/KiaUvoApiAU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiAU.py
@@ -231,28 +231,6 @@ class KiaUvoApiAU(ApiImplType1):
             else:
                 self._update_vehicle_drive_info(vehicle, state)
 
-    def _force_refresh_vehicle_state_ccs2(self, token: Token, vehicle: Vehicle) -> None:
-        url = self.SPA_API_URL + "vehicles/" + vehicle.id + "/ccs2/carstatus"
-        response = self.session.get(
-            url,
-            headers=self._get_authenticated_headers(
-                token, vehicle.ccu_ccs2_protocol_support
-            ),
-        ).json()
-        _LOGGER.debug(
-            f"{DOMAIN} - Force refresh CCS2 vehicle status response: {response}"
-        )
-        _check_response_for_errors(response)
-        state = response["resMsg"]["state"]["Vehicle"]
-        self._update_vehicle_properties_ccs2(vehicle, state)
-        location = self._get_location(token, vehicle)
-        if location and get_child_value(location, "coord.lat"):
-            vehicle.location = (
-                get_child_value(location, "coord.lat"),
-                get_child_value(location, "coord.lon"),
-                parse_datetime(get_child_value(location, "time"), self.data_timezone),
-            )
-
     def _update_vehicle_properties(self, vehicle: Vehicle, state: dict) -> None:
         if get_child_value(state, "status.time"):
             vehicle.last_updated_at = parse_datetime(
--- a/hyundai_kia_connect_api/KiaUvoApiCN.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCN.py
@@ -8,6 +8,7 @@ import math
 import typing as ty
 import uuid
 from time import sleep
+from typing import Any
 from urllib.parse import parse_qs, urlparse
 from zoneinfo import ZoneInfo
 
@@ -269,27 +270,31 @@ class KiaUvoApiCN(ApiImplType1):
             else:
                 self._update_vehicle_drive_info(vehicle, state)
 
-    def _force_refresh_vehicle_state_ccs2(self, token: Token, vehicle: Vehicle) -> None:
-        url = self.SPA_API_URL + "vehicles/" + vehicle.id + "/ccs2/carstatus"
-        response = self.session.get(
-            url,
-            headers=self._get_authenticated_headers(
-                token, vehicle.ccu_ccs2_protocol_support
-            ),
-        ).json()
-        _LOGGER.debug(
-            f"{DOMAIN} - Force refresh CCS2 vehicle status response: {response}"
-        )
-        _check_response_for_errors(response)
-        state = response["resMsg"]
-        self._update_vehicle_properties(vehicle, state)
-        location = self._get_location(token, vehicle)
-        if location and get_child_value(location, "coord.lat"):
-            vehicle.location = (
-                get_child_value(location, "coord.lat"),
-                get_child_value(location, "coord.lon"),
-                parse_datetime(get_child_value(location, "time"), self.data_timezone),
+    def _inspect_ccs2_response(
+        self, response: dict[str, Any]
+    ) -> tuple[dict[str, Any] | None, dt.datetime | None]:
+        """CN CCS2 shape: resMsg root + status.time (non-CCS2 lookalike).
+
+        Does NOT call the parser — caller checks freshness before applying.
+        """
+        state = response.get("resMsg")
+        if not isinstance(state, dict):
+            return None, None
+        raw_time = get_child_value(state, "status.time")
+        if not raw_time:
+            return state, None
+        try:
+            last = parse_datetime(str(raw_time), self.data_timezone)
+            return state, last
+        except (ValueError, TypeError):  # fmt: skip
+            _LOGGER.warning(
+                f"{DOMAIN} - CN CCS2 _inspect: cannot parse status.time '{raw_time}'"
             )
+            return state, None
+
+    def _apply_ccs2_state(self, vehicle: Vehicle, state: dict[str, Any]) -> None:
+        """CN uses its own non-CCS2 parser (status.* path)."""
+        self._update_vehicle_properties(vehicle, state)
 
     def _update_vehicle_properties(self, vehicle: Vehicle, state: dict) -> None:
         if get_child_value(state, "status.time"):
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -444,22 +444,6 @@ class KiaUvoApiEU(ApiImplType1):
             else:
                 self._update_vehicle_drive_info(vehicle, state)
 
-    def _force_refresh_vehicle_state_ccs2(self, token: Token, vehicle: Vehicle) -> None:
-        url = self.SPA_API_URL + "vehicles/" + vehicle.id + "/ccs2/carstatus"
-        response = self.session.get(
-            url,
-            headers=self._get_authenticated_headers(
-                token, vehicle.ccu_ccs2_protocol_support
-            ),
-        ).json()
-        _LOGGER.debug(
-            f"{DOMAIN} - Force refresh CCS2 vehicle status response: {response}"
-        )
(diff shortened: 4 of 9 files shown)


blka:
## Live validation of the trigger/poll approach

Ran live tests on **EU Hyundai Santa Fe 2026 (CCS2=1)** on cellular signal (2026-06-19) to confirm the core assumption: `/ccs2/carstatus` (without `/latest`) wakes the car and polling `/ccs2/carstatus/latest` detects the fresh snapshot via `Date`.

### Endpoint behavior (empirical)

| Endpoint | Method | Wakes car? | Response shape |
|---|---|---|---|
| `/ccs2/carstatus/latest` | GET | **No** | `resMsg` = `{resCode, ServiceNo, RetCode, lastUpdateTime, state.Vehicle…}` (cached snapshot) |
| `/ccs2/carstatus` (no `/latest`) | GET | **Yes** (~20 s) | `{retCode:S, resCode:0000, msgId:…}`, `resMsg=null` |
| `/status` (non-CCS2 force) | GET | Yes (~20 s) | `resMsg` = `{ServiceNo:"RVS-B", SID, svcSID, FncCnt, svcTime…}` |

### Control — `/latest` alone does **not** wake

Polled `/ccs2/carstatus/latest` 3× over 30 s with **no trigger call** → CCU `Date` stayed `20260619123117.000` (unchanged). Confirms the current single-`/latest` read returns stale cache, matching the problem statement.

### Trigger — `/ccs2/carstatus` wakes

Baseline `Date=20260619123117.000` → `GET /ccs2/carstatus` (no `/latest`) returned `0000` + `msgId`, `resMsg=null` → polling `/latest`:

| t after trigger | CCU Date | |
|---|---|---|
| 10 s | 12:31:17 | same |
| 21 s | **12:34:15** | fresh — advanced at ~20.6 s |
| 31–82 s | 12:34:15 | stable |

Cross-checked with `/status`: baseline 12:24:04 → after `GET /status`, `/latest` advanced to 12:28:20 at ~20.5 s. Same wake effect, different response shape (`SID` vs `msgId`).

### Takeaways

- The trigger (`GET /ccs2/carstatus` no `/latest`) + poll `/latest` + `Date` freshness check pattern is **validated** — fresh data lands ~20 s after the trigger.
- `/ccs2/carstatus` returns `msgId` (CCS2 style, consistent with the `control/*` commands and `check_action_status`); `/status` returns `SID`. Using the CCS2-native `/ccs2/carstatus` (as this PR does) is the right call.
- The `Date` `.000` ms suffix is real (`20260619123117.000`) — the PR's `.000` parsing is needed.
- **Budget headroom:** fresh data landed at **20.5 s** and **20.6 s** across two runs — right at the 10 × 2 s = 20 s poll budget edge (the PR's own live test noted 22.5 s). Consider bumping to ~12–15 polls or a 3 s interval so a slightly-slow cellular round-trip doesn't fall through to the timeout fallback.

Scripts: `tests/integration/test_force_wake_validation_live.py`, `test_ccs2_carstatus_wake_live.py`, `test_endpoints_probe_live.py` (local, not committed — credentials via gitignored `.env`).

blka:
## HTTP call comparison for CCS2 `force_refresh_vehicle_state` (single invocation)

| Variant | HTTP calls | What it returns |
|---|---|---|
| Upstream original | 2 | `/ccs2/carstatus/latest` (cached, does **not** wake car) + `/location/park` |
| PR #1184 (best case) | 3 | trigger `/carstatus` + 1 poll `/latest` (fresh immediately) + `/location/park` |
| PR #1184 (typical) | ~12 | trigger + ~10 polls `/latest` + `/location/park` |
| PR #1184 (timeout) | 12 | trigger + 10 polls (timeout fallback) + `/location/park` |

Delta: +1 to +10 calls per force-refresh, typically ~6× more (2 → 12).

### Why this isn't apples-to-apples

The original (2 calls) was a **bug**: `/ccs2/carstatus/latest` returns the CCU's cached snapshot and does not wake the car. Force refresh in practice did not refresh anything — it returned the same data the periodic poll already had. So:

- Original: 2 calls = 0 actual refresh
- PR: ~12 calls = genuine fresh data from the car (~20 s after the trigger, live-confirmed)

### Mitigating context

1. **This is not the periodic poll.** `force_refresh` is on-demand — the user clicks "Refresh" in HA, or the coordinator calls it after a command (lock/climate) to confirm state. Normal HA polling uses cached `/latest` without waking. So the delta only affects rare actions, not the minute-by-minute scan.
2. **The 20 s budget (10×2 s) is on the edge.** Live tests: fresh data lands ~20.5–22.5 s after the trigger. Often the full budget is consumed → timeout → fallback to last data. I'd propose bumping to ~12–15 polls / 3 s interval so a slow cellular round-trip doesn't fall through to the timeout fallback.

### Conclusion

The increase is real (~6× per force-refresh), but: (a) it compensates a functional bug (force refresh previously didn't work), and (b) it only affects a rare path. The main drawback is **latency, not call count** — force refresh blocks ~20 s on cellular. For HA this is acceptable (it's a user-initiated interaction), but the budget bump is worth considering to reduce timeout-fallback frequency.

---

### Why I'm bringing this up

Based on this reliable force-refresh primitive, I want to build a fix for **CCS2 EU that actually returns the status of sent commands**. Today `check_action_status` polls `/notifications/{vehicleId}/records`, which EU CCS2 never populates — every `/ccs2/control/*` command (lock, climate, charge, windows, …) comes back `UNKNOWN`/`TIMEOUT`, even though the car executes it (confirmed live for lock + windows, and climate in kia_uvo#995). The plan: confirm commands by reading `carstatus/latest` after the command (what the Bluelink app does), building on this PR's reliable `force_refresh`.

@cdnninja — could you take a look at this PR when you have a moment? I'd like to build the command-status fix on top of it, so a review here unblocks that work. Happy to adjust the poll budget or anything else based on your feedback.

blka:
## Refactor: polling loop extracted to base + Date bug fix + HA alignment

Rebased onto `upstream/master` (v4.23.2). Refactored per the design discussion above. Summary for review:

### What changed
- **Polling loop extracted to base** `ApiImplType1._force_refresh_vehicle_state_ccs2` + 3 hooks (`_inspect_ccs2_response`, `_apply_ccs2_state`, `_post_refresh_ccs2_location`). Eliminates the 4× copy-paste of trigger+poll+apply across EU/AU/IN/CN.
- **Date bug fixed (off-by-1h/2h):** the old override parsed `Date` with `.replace(tzinfo=data_timezone)`, treating it as local time. `Date` is UTC (#1147). `_inspect_ccs2_response` parses `Date` as UTC via `parse_datetime(..., dt.timezone.utc).astimezone(data_timezone)` — consistent with `_update_vehicle_properties_ccs2`. With the old code, `last_update.timestamp() > trigger_time` was never satisfied in real life → always timeout → stale fallback. Now fixed.
- **HA principle (unavailable = data unknown):** apply state ONLY when fresh. On timeout (12 polls × 3s = 36s, no fresh data), keep previous vehicle state and log a warning. The old code's "use last fetched data" fallback **masked stale** — removed.
- **Budget bumped 10×2s=20s → 12×3s=36s.** Live tests showed fresh data lands ~20-22s after the trigger; the old 20s budget was at the edge.
- **Per-region hooks:**
  - EU: inherits `_inspect`/`_apply`; `_post_refresh` overrides to `_set_cached_location_park` (`/location/park`).
  - AU: inherits all 3 base defaults (response shape matches base).
  - IN: overrides `_inspect` (`resMsg` root + `state.time` via `get_last_updated_at`), `_apply` (IN non-CCS2 parser), `_post_refresh` (`_update_vehicle_location`).
  - CN: overrides `_inspect` (`resMsg` root + `status.time`), `_apply` (CN non-CCS2 parser); inherits base `_post_refresh` (CN `_get_location` matches base default). CN never sets `ccu_ccs2_protocol_support` → flag stays `None`; `None != 0` is `True` → CCS2 path IS invoked → CN is NOT dead code → hooks needed (not a full override removal).

### Verification
- 386 unit tests pass, 10 snapshots, `ruff` + `pre-commit` (pyupgrade --py312-plus) clean.
- **Live-validated on EU Santa Fe CCS2:** `force_refresh` took 20.5s, `last_updated_at` advanced +18.2s after trigger (fresh) → freshness check passes → apply called. Confirms the Date-UTC fix on real hardware.

### Questions for review
1. **IN/CN `time`/`status.time` timezone.** `IN get_last_updated_at` and `CN parse_datetime(status.time, ...)` treat the time string as `data_timezone` (Asia/Kolkata / Asia/Shanghai). If either server returns `time` as UTC (like EU `Date`), the freshness check would always fail → force-refresh always times out. This is **pre-existing** (the old overrides had the same tz handling); this PR does not regress it. The hook structure makes the fix a one-line change in `_inspect_ccs2_response` once a live IN/CN dump confirms UTC vs local. **Can you confirm the IN/CN CCS2 `time` field semantics, or provide a dump?**
2. **Budget 12×3s=36s** — acceptable for on-demand force-refresh (not cyclic poll)? Any concern about the 36s blocking in HA's `asyncio.to_thread`?
3. **CN `ccu_ccs2_protocol_support` stays `None`.** `None != 0` is `True`, so the CCS2 path is invoked for CN even though CN never sets the flag. This PR keeps CN on hooks (not a removal). Is the `None`-means-CCS2 behavior intentional, or should CN explicitly set the flag to `0` for non-CCS2 vehicles (separate follow-up)?

cdnninja:
@fuatakgun  I am curious on your thoughts on this one.

blka:
Rebased onto current `master` (v4.24.1) — conflicts resolved, CI re-triggered. 424 tests green.

This also resolves kia_uvo #1786 (force-refresh button → `KeyError: 'resMsg'`): the current `master` `_force_refresh_vehicle_state_ccs2` reads `response["resMsg"]["state"]["Vehicle"]` directly from the wake endpoint `GET /ccs2/carstatus`, which returns `{retCode: "S", msgId: "..."}` (async envelope), not the status — hence the `KeyError`. This PR's base `_force_refresh_vehicle_state_ccs2` treats the wake as async (wake → poll `/latest` until `last_updated` advances → apply only when fresh) and `_inspect_ccs2_response` handles a missing `resMsg` gracefully, so force-refresh stops crashing on the wake response. #1786 is good evidence this should land — the force-refresh path is broken on `master` for CCS2 vehicles.

@cdnninja — ready for review.

blka:
@fuatakgun @Lekensteyn — @cdnninja is away for two weeks and suggested one of you take a look.

This fixes the EU/AU/IN/CN CCS2 force-refresh path that currently raises `KeyError: 'resMsg'` (kia_uvo #1786 — two reporters on 3.7.0/3.7.1; DEBUG confirms the wake endpoint returns `{retCode, resCode, msgId}`, not `resMsg.state.Vehicle`). Refactored into base `ApiImplType1._force_refresh_vehicle_state_ccs2` + 3 polymorphic hooks so all four regions share one trigger+poll loop. HA principle: state applied only when fresh (`last_updated > trigger_time`); on timeout the previous state is kept untouched (no stale masking). Budget 12×3s.

Status: rebased onto current master (v4.24.1), 1 commit, all CI green (py3.12/3.13/3.14, CodeQL, Analyze), live-validated EU Santa Fe. I've just refreshed the description to match the refactored implementation. IN/CN `status.time`/`state.time` timezone handling is pre-existing and documented in the tests — deferred pending a dump. Happy to address any feedback.

blka:
@cdnninja @fuatakgun — gentle nudge for a decision. This PR has been open since June 16 (~5 weeks) and fixes the `KeyError: 'resMsg'` in `_force_refresh_vehicle_state_ccs2` (kia_uvo #1786).

kia_uvo #1806 was filed today — same `KeyError: 'resMsg'` at `KiaUvoApiEU.py:459`, IONIQ 5 (CCS2), on the **current** release (kia_uvo 3.8.0 / hyundai_kia_connect_api 4.25.2). So the bug is still shipping and this PR is the fix.

Status:
- Rebased onto current `upstream/master` (`65b4464`), single commit, mergeable.
- Unit tests green (polling loop fresh / timeout-keeps-state / budget 14×10s, per-region hooks, location).
- Live-validated on an EU CCS2 vehicle with the car reachable: fresh data detected after **2 polls = 20.7s** (early-exit); the ~140s window is headroom for an unreachable / deep-sleep vehicle, not the typical path.
- Request budget vs today's single crashing GET: **+2 requests typical** (early-exit once fresh), **+14 max** over ~140s only when the car can't report (≈0.11 req/s, well under any rate limit; on timeout it keeps prior state instead of raising). The wake-trigger + poll-for-fresh sequence and the ~140s/10s timing mirror what the official app already does.

Could you take a look and let me know whether this is mergeable or if you'd like any changes? Happy to address feedback. This would also close the new kia_uvo #1806 duplicate.

hyundai_kia_connect_api/ApiImplType1.py
```diff
@@ -798,6 +799,102 @@ def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None
...
+            return state, None
+
+    def _apply_ccs2_state(self, vehicle: Vehicle, state: dict[str, Any]) -> None:
+        """Parse + write vehicle state. Called only when data is fresh.
+
+        Default impl: CCS2 parser (EU/AU).
+        """
+        self._update_vehicle_properties_ccs2(vehicle, state)
+
+    def _post_refresh_ccs2_location(self, token: Token, vehicle: Vehicle) -> None:
```
cdnninja, line 833: I would think from naming this isn't post but rather just refreshes location.

hyundai_kia_connect_api/ApiImplType1.py
```diff
@@ -798,6 +799,102 @@ def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None
...
+        if not raw_date:
+            return state, None
+        try:
+            last = parse_datetime(str(raw_date).split(".")[0], dt.timezone.utc)
+            return state, last.astimezone(self.data_timezone)
+        except (ValueError, TypeError):
+            _LOGGER.warning(f"{DOMAIN} - CCS2 _inspect: cannot parse Date '{raw_date}'")
+            return state, None
+
+    def _apply_ccs2_state(self, vehicle: Vehicle, state: dict[str, Any]) -> None:
```
cdnninja, line 826: What value does an internal method exists that just calls a single internal method?

hyundai_kia_connect_api/ApiImplType1.py
```diff
@@ -798,6 +799,102 @@ def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None
 
         vehicle.data = state
 
+    def _inspect_ccs2_response(
```
cdnninja, line 802: This method feels complex for the need but maybe I'm missing something?

cdnninja (commented):
A few comments.   As a note we used to just run a sleep, then cached call in the overall approach. That was removed at some point since we got the data from forced call.    Would that solve the same issue but much simpler? 

I worry about how many internal methods and complexity is continued to be added from a maintenance standpoint. 
 
Thoughts?

blka:
Opened a simpler alternative: **#1250** — `fix(EU): simpler wake+sleep+read CCS2 force refresh`.

In response to your review here ("Would that solve the same issue but much simpler? I worry about how many internal methods and complexity is continued to be added"): #1250 replaces `_force_refresh_vehicle_state_ccs2` with wake (`/ccs2/carstatus`) + `sleep(25)` + read cached `/ccs2/carstatus/latest` + apply — no poll loop, no freshness check, no hooks, no new internal methods (~8 lines, reuses existing parser + location). EU only (AU/IN/CN stay covered by this PR #1184).

**Both PRs are open — your call which to merge.** Quick compare:

| | #1250 (sleep+read) | #1184 (poll-until-fresh) |
|---|---|---|
| Requests / force-refresh | 2 (fixed) | 3 typical / 15 max |
| Typical reachable (~20s) | fresh | fresh, early-exit ~20s |
| Slow/deep-sleep | applies cached (stale, visible via last_updated_at) | polls to 140s, catches fresh / keep-previous |
| Complexity | ~8 lines, no new methods | base + 3 hooks + poll loop |
| Regions | EU only | EU/AU/IN/CN |

`last_updated_at` is honest in both (staleness visible, not masked). #1250 is "good enough + simple" for the typical reachable case; #1184 is more robust for slow/deep-sleep + covers all 4 regions. #1250 fixes the same `KeyError: 'resMsg'` (kia_uvo #1786, #1806). Let me know which direction you prefer (or if you'd want #1250's simplicity applied to all 4 regions).

blka:
Closing in favor of #1250, which merged — cdnninja's review asked for the simpler approach. The 4-region base+hooks refactor here is superseded; #1250's minimal EU `wake → sleep → read /latest` fixes the `KeyError: 'resMsg'` (#1786) without the broader refactor. AU/IN/CN keep the existing cached `/latest` force-refresh path. Thanks all.

Issue #1185, opened 2026-06-16
Polling errors after performing actions (seen w/ lock & unlock at least)


cdonovan:
When calling lock/unlock actions the initial request succeeds but the subsequent confirmation polling call to rmt/getRunningStatus returns HTTP 405 (Method Not Allowed) and raises an APIError. This surfaces in Home Assistant logs as:  

```
Action 'unlock vehicle' was sent but confirmation polling failed
...
hyundai_kia_connect_api.exceptions.APIError: check_action_status failed with HTTP 405: '"<!doctype html><html lang=\"en\"><head><title>HTTP Status 405 – Method Not Allowed</title>...'
```

Reproduction steps:
1. Use the integration to lock or unlock a vehicle (USA region). 
2. Observe the action succeeds on the vehicle but the integration logs the error above.

Observed behavior:
- lock/unlock requests succeed, but confirmation polling logs HTTP 405 and raises an exception.

Expected behavior:
- Confirmation polling should use the correct HTTP method/endpoint and return the action status without raising HTTP 405, or gracefully handle non-200 responses so the user log isn't noisy when actions succeed.

Investigation notes:
- In HyundaiBlueLinkApiUSA.check_action_status the code calls POST on `rmt/getRunningStatus` (line ~882). The API appears to reject POST with 405. The control calls (lock_action) use different endpoints and succeed.
- Possible fixes: switch the polling call to GET, make the failure non-exceptional, or update the endpoint path.

Relevant files/lines:
- hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py: check_action_status (lines ~865-900)
- hyundai_kia_connect_api/VehicleManager.py: check_action_status and lock/unlock call paths

Logs/Stack trace: included above.

Additional context:
- This affects USA region implementation. Other regions might be unaffected. The issue appears when actions are sent from Home Assistant through the custom component.

blka:
@cdnninja this one should be done

Issue #1167, opened 2026-06-02
Hazard lights and Horn


ijojog:
For NX5 Tucson, Hazard lights and Horn do not work. I am really interested is making these go off, when my camera detects people around the vehicle for too long, as theft has been really up in my area lately.  Having the hazard lights go off as warning would be great, and later, both the lights and horn.   However, that does not work for the Tucson. Would you amazing geniuses please look into it?

Thank you so very much!!!

blka:
Can you share your region and debug logs while using cureent service pls?

ijojog:
Thank you. Region is USA

Logger: homeassistant.components.websocket_api.http.connection
Source: components/websocket_api/connection.py:332
Integration: Home Assistant WebSocket API ([documentation](https://www.home-assistant.io/integrations/websocket_api), [issues](https://github.com/home-assistant/core/issues?q=is%3Aissue+is%3Aopen+label%3A%22integration%3A+websocket_api%22))
First occurred: 12:56:05 PM (2 occurrences)
Last logged: 12:56:10 PM

[140200385705952] Error handling message: Failed to start hazard lights: start_hazard_lights is not implemented for this region (home_assistant_error) xxxxxx from xxxxxxxxxx (Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/148.0.0.0 Safari/537.36 Edg/148.0.0.0)


and the other

Logger: homeassistant.helpers.script.websocket_api_script
Source: helpers/script.py:2184
First occurred: 12:56:05 PM (2 occurrences)
Last logged: 12:56:10 PM

websocket_api script: Error executing script. Error for call_service at pos 1: Failed to start hazard lights: start_hazard_lights is not implemented for this region

blka:
is this working from the original app?

ijojog:
Yes, absolutely.

blka:
Thanks for confirming it works from the app. The issue is that `HyundaiBlueLinkApiUSA` does not implement `start_hazard_lights` / `start_hazard_lights_and_horn` — these methods only exist for EU/AU/IN regions (CCS2 protocol) and BR.

To add this for USA, we need to know the exact API endpoint and payload the Hyundai/Genesis USA app sends. The EU implementation uses:
- `POST /api/v2/spa/vehicles/{id}/ccs2/control/light` with `{"command": "on"}`
[...]

PR #1191, opened 2026-06-17
fix(USA): use GET for action status polling

blka:
## Summary

Fixes the HTTP 405 Method Not Allowed error on action status polling reported in #1167 after PR #1180.

## Root cause

`check_action_status` polled `/ac/v2/rmt/getRunningStatus` with **POST**. The native myHyundai iOS app (captured by @ijojog in #1167) uses **GET**:

```
GET https://api.telematics.hyundaiusa.com/ac/v2/rmt/getRunningStatus
```

The server rejects POST with HTTP 405. The endpoint name (`getRunningStatus`) also indicates GET.

## Why it was masked

This affected **all** USA action polling (lock, climate, charge, horn, hazard), not just horn/hazard. It was masked for lock/climate/charge because the kia_uvo coordinator catches polling exceptions silently — the action command itself succeeds, so users never noticed the polling failure. Horn/hazard surfaced it because the issue reporter captured the native app traffic and tested specifically.

## Fix

Change `self.sessions.post` → `self.sessions.get` in `check_action_status`. The `service_type` mapping from #1180 stays — those values (`HORN_AND_LIGHTS`, `LIGHTS_ONLY`, `REMOTE_POLL`) are correct per the capture; only the HTTP method was wrong.

## Testing

- Updated `us_action_status_test.py` to mock `sessions.get`
- All 182 unit tests pass
- @ijojog: needs retest to confirm polling now returns SUCCESS/PENDING instead of 405

Ref #1167

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
@@ -885,7 +885,7 @@ class HyundaiBlueLinkApiUSA(ApiImpl):
         max_attempts = 1 if not synchronous else max(1, timeout // 2)
 
         for _ in range(max_attempts):
-            response = self.sessions.post(url, headers=headers)
+            response = self.sessions.get(url, headers=headers)
             response_json = _safe_parse_json(response, "check_action_status")
             if response_json is None:
                 if not synchronous:
--- a/tests/us_action_status_test.py
+++ b/tests/us_action_status_test.py
@@ -50,7 +50,7 @@ def _make_api():
 class TestCheckActionStatus:
     def test_success(self):
         api = _make_api()
-        api.sessions.post.return_value = _FakeResponse(json_data={"status": "SUCCESS"})
+        api.sessions.get.return_value = _FakeResponse(json_data={"status": "SUCCESS"})
         vehicle = _make_vehicle()
         token = _make_token()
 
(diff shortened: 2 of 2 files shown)


blka:
@cdnninja can take care of it? small method update from the morning

blka:
Fixes #1185

ijojog:
Quick update from my Home Assistant server:

After reverting back to the stable `kia_uvo` setup, I tried the hazard service again from Developer Tools.

Result:

```text
Failed to perform the action kia_uvo.start_hazard_lights.
Failed to start hazard lights: start_hazard_lights is not implemented for this region
```

So I think the services/buttons may now be present in the HA side, but my currently installed stable backend/API package still does not include the USA horn/hazard implementation yet.

When I was testing the PR branch earlier, the same physical vehicle actions worked:

```text
Hazard lights only: worked physically
Horn + hazard lights: worked physically
Confirmation polling issue: fixed by #1191
```

So from my side, the issue now seems to be release/update wiring rather than the vehicle/API capability itself.

I am happy to help test anything needed to get this finished faster. If you want me to test a `kia_uvo` branch, an API branch, or a specific manifest requirement line, I can do that and report back with screenshots/logs.

Also, if there is a simple PR or documentation/test result I can contribute, I’m willing to help. I’m not a maintainer/developer on this project, but I can reliably test on a Hyundai USA 2025 Tucson Hybrid and provide app/API behavior, Home Assistant screenshots, and logs.

PR #1192, opened 2026-06-18
fix(EU/AU/IN/CN): resolve drvSeatLoc dynamically instead of hardcoded R

blka:
## Summary

The CCS2 `start_climate` payload sent `drvSeatLoc: "R"` (RHD) for every region, which is wrong for the LHD majority of EU and CN. This resolves it dynamically per region.

This is the drvSeatLoc half of the closed PR #1183. The CCS2 seat-retry half was dropped — no user has reported the specific `retCode=="F"`-from-`seatClimateInfo` case in either tracker, so I did not add speculative defensive code. The full investigation (issue references, egmp-bluelink-scriptable analysis, the 4005 root cause) is documented on #1183.

## Resolution (mirrors egmp-bluelink-scriptable)

- **`ApiImplType1._get_drv_seat_loc(vehicle)`** — base implementation, used by EU and CN. Uses the vehicle odometer unit: mile-based markets (UK, Ireland) → `"R"`, kilometre-based → `"L"`. Falls back to `"L"` when `odometer_unit` is unavailable (correct for the LHD majority of EU/CN).
- **`KiaUvoApiAU._get_drv_seat_loc`** — override → `"R"`. AU is RHD despite using kilometres.
- **`KiaUvoApiIN._get_drv_seat_loc`** — override → `"R"`. IN is RHD despite using kilometres.
- EU and CN use the base implementation.

This mirrors egmp: `europe.ts` uses `distanceUnit === 'mi' ? 'R' : 'L'`, while `australia.ts` hardcodes `drvSeatLoc: 'R'`.

## Why not `vehicle.distance_unit` (the closed PR #1183 approach)

PR #1183 used `vehicle.distance_unit`, but `Vehicle` has **no `distance_unit` attribute** (that field exists only on `DailyDrivingStats`). Every `start_climate` call after a cached state fetch raised:

```
AttributeError: 'Vehicle' object has no attribute 'distance_unit'
```

This PR uses `vehicle.odometer_unit`, which is a real `Vehicle` property. It is populated by `_update_vehicle_properties` from the cached state `odometer.unit` field (EU/AU/CN) or defaults to the kilometre constant (base/IN). It returns `None` when the response omits the unit, which the base logic handles by falling back to `"L"`.

## Known limitation

For EU UK/Ireland users on the new CCS2 state format where the response does not include `odometer.unit`, `odometer_unit` is `None` and `drvSeatLoc` resolves to `"L"` (should be `"R"` for RHD). This is a rare edge case (UK/IE on CCS2 without the unit field) and the same degradation egmp would have. The previous hardcoded `"R"` was correct for UK but wrong for the rest of EU; this is correct for the rest of EU and degrades only for UK without the signal. A per-vehicle LHD/RHD field from the API would be the proper fix, but no such field is exposed in the cached state response.

## Files changed

- `ApiImplType1.py` — add `_get_drv_seat_loc`, use it in `start_climate` CCS2 payload
- `KiaUvoApiAU.py` — override `_get_drv_seat_loc` → `"R"`
- `KiaUvoApiIN.py` — override `_get_drv_seat_loc` → `"R"`
- `tests/test_drv_seat_loc.py` — 10 unit tests (base km/miles/None + AU/IN overrides)

## Testing

- 10 new unit tests pass
- Full suite: 192 passed (was 177)
- `stop_climate` payload is `{"command": "stop"}` only — no `drvSeatLoc`, unchanged
- Pre-commit hooks pass (ruff, codespell, pyupgrade, etc.)

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -965,6 +965,25 @@ class ApiImplType1(ApiImpl):
         token.device_id = self._get_device_id(self._get_stamp())
         return response["msgId"]
 
+    def _get_drv_seat_loc(self, vehicle: Vehicle) -> str:
+        """Return the driver seat location for CCS2 climate payloads.
+
+        "L" for LHD (left-hand drive), "R" for RHD (right-hand drive).
+        Derived from the vehicle's reported odometer unit: mile-based markets
+        (UK, Ireland) are RHD, kilometre-based markets are LHD.
+
+        This base implementation covers EU and CN. RHD regions that use
+        kilometres (AU, IN) override this method to return "R" directly,
+        because the km/miles signal would incorrectly resolve to "L".
+
+        Falls back to "L" when the odometer unit is unavailable (e.g. the
+        cached state response did not include it), which is correct for the
+        LHD majority of EU/CN markets.
+        """
+        if vehicle.odometer_unit in (DISTANCE_UNITS[2], DISTANCE_UNITS[3]):
+            return "R"
+        return "L"
+
     def start_climate(
         self, token: Token, vehicle: Vehicle, options: ClimateRequestOptions
     ) -> str:
@@ -1020,7 +1039,7 @@ class ApiImplType1(ApiImpl):
                 "hvacTempType": 1,
                 "hvacTemp": options.set_temp,
                 "sideRearMirrorHeating": 1,
-                "drvSeatLoc": "R",
+                "drvSeatLoc": self._get_drv_seat_loc(vehicle),
                 "seatClimateInfo": {
                     "drvSeatClimateState": options.front_left_seat,
                     "psgSeatClimateState": options.front_right_seat,
--- a/hyundai_kia_connect_api/KiaUvoApiAU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiAU.py
@@ -662,6 +662,10 @@ class KiaUvoApiAU(ApiImplType1):
         if response["resMsg"] is not None:
             return response["resMsg"]
 
+    def _get_drv_seat_loc(self, vehicle: Vehicle) -> str:
+        """Australia uses RHD vehicles regardless of the odometer unit."""
+        return "R"
+
     def _get_trip_info(
         self,
         token: Token,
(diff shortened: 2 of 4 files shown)


blka:
Live-validated on my own car: EU Hyundai Santa Fe 2026 (CCS2=1, PL, LHD, km).

## Results

| Check | Result |
|-------|--------|
| `vehicle.odometer_unit` after `update_all_vehicles_with_cached_state()` | `'km'` (populated from `state.odometer.unit`) |
| `api._get_drv_seat_loc(vehicle)` | `'L'` ✅ (was hardcoded `'R'`) |
| `AttributeError` from PR #1183's `vehicle.distance_unit` | gone ✅ |
| `start_climate` payload `drvSeatLoc` | `'L'` ✅ |
| `stop_climate` payload | `{'command': 'stop'}` — no `drvSeatLoc` ✅ (unchanged) |

Captured `start_climate` payload (via mocked `requests.post`, so no API call, no car contact, no wasted 4005):

```python
{'command': 'start', 'ignitionDuration': 5, 'strgWhlHeating': None,
 'hvacTempType': 1, 'hvacTemp': 22, 'sideRearMirrorHeating': 1,
 'drvSeatLoc': 'L',   # was hardcoded 'R' on master
 'seatClimateInfo': {...}, 'tempUnit': 'C', 'windshieldFrontDefogState': False}
```

## Note on `odometer_unit` availability

I had flagged a known limitation that `odometer_unit` might be `None` on the new CCS2 state format (where the body has `Drivetrain.Odometer` as a flat number). Live testing disproved that: on the 2026 Santa Fe the EU cached-state response **does** include `odometer.unit`, so `KiaUvoApiEU._update_vehicle_properties` populates `vehicle.odometer_unit = 'km'`. This means the miles → `"R"` path for UK/Ireland will actually work when a UK car reports miles — not just in theory. The fallback to `"L"` when the unit is absent only triggers if a response genuinely omits `odometer.unit`.

## Test method

`tests/integration/test_pr1192_drvseatloc_live.py` — login + cached state fetch (live, read-only), then `requests.post` is monkeypatched to capture the `start_climate`/`stop_climate` payloads without sending the command. No command reaches the car (the 4005 endpoint issue on 2025+ vehicles is documented on #1183 and is unrelated to drvSeatLoc). This file is a local integration test, not committed to the PR — the PR ships the 10 offline unit tests in `tests/test_drv_seat_loc.py`.

PR #1193, opened 2026-06-18
fix(USA): auto-recover get_vehicles/refresh_vehicles on session expiry (1003/1005)

blka:
## Summary

`KiaUvoApiUSA.get_vehicles` and `refresh_vehicles` called `self.session.get()` without the `@request_with_logging` / `@request_with_active_session` decorators, so an expired-session response (errorCode 1003) raised a generic `APIError("Missing payload in response")` instead of `AuthenticationError`. With `test_token` a no-op for USA and `check_and_refresh_token` driven only by the `valid_until` timestamp, a server-side session expiry before `valid_until` left the integration looping `ConfigEntryNotReady` with a dead token — the user had to delete and re-add the account (kia_uvo#1685).

This adds a token-only `_retry_on_auth_error` decorator (symmetric with `request_with_active_session`, which requires `vehicle` / `json_body` these methods don't have) and makes both methods raise `AuthenticationError` on errorCode 1003/1005. The decorator re-logs in, mutates the token in place, and retries once.

This is the USA analog of #1136 (which fixed the same class of bug for EU / `ApiImplType1`).

Fixes kia_uvo#1685

## Test plan

- [x] `get_vehicles` with 1003 response raises `AuthenticationError` and recovers after re-login
- [x] `get_vehicles` success path does not retry
- [x] `get_vehicles` raises `AuthenticationOTPRequired` when re-login needs OTP
- [x] `get_vehicles` retries exactly once (second failure propagates)
- [x] `refresh_vehicles` equivalent coverage
- [x] `pytest tests/ -q` passes (204 passed; 5 pre-existing live-integration errors unrelated)
- [x] `pre-commit run --all-files` passes

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiUSA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiUSA.py
@@ -117,13 +117,16 @@ def _retry_on_auth_error(func):
             _LOGGER.debug(
                 f"{DOMAIN} - Session expired during {func.__name__}, re-logging in"
             )
-            new_token = self.login(
-                token.username,
-                token.password,
-                token,
-                getattr(self, "_otp_handler", None),
-            )
-            if not isinstance(new_token, Token):
+            try:
+                new_token = self.login(
+                    token.username,
+                    token.password,
+                    token,
+                    getattr(self, "_otp_handler", None),
+                )
+            except Exception as e:
+                raise AuthenticationError(f"Re-login failed: {e}") from e
+            if isinstance(new_token, OTPRequest):
                 raise AuthenticationOTPRequired("OTP required to refresh token")
             token.access_token = new_token.access_token
             token.refresh_token = new_token.refresh_token
--- a/hyundai_kia_connect_api/KiaUvoApiUSA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiUSA.py
@@ -370,6 +370,7 @@ class KiaUvoApiUSA(ApiImpl):
         """Refresh the token using the refresh token"""
         return self.login(token.username, token.password, token)
 
+    @_retry_on_auth_error
     def get_vehicles(self, token: Token) -> list[Vehicle]:
         """Return all Vehicle instances for a given Token"""
         url = self.API_URL + "ownr/gvl"
@@ -378,6 +379,13 @@ class KiaUvoApiUSA(ApiImpl):
         response = self.session.get(url, headers=headers)
         _LOGGER.debug(f"{DOMAIN} - Get Vehicles Response {response.text}")
         response = response.json()
+        status = response.get("status") or {}
+        if (
+            status.get("statusCode") == 1
+            and status.get("errorType") == 1
+            and status.get("errorCode") in (1003, 1005)
+        ):
+            raise AuthenticationError("Session invalid")
         if "payload" not in response:
             raise APIError("Missing payload in response")
         result = []
--- a/hyundai_kia_connect_api/KiaUvoApiUSA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiUSA.py
@@ -411,6 +411,7 @@ class KiaUvoApiUSA(ApiImpl):
             return ENGINE_TYPES.EV
         return None
 
+    @_retry_on_auth_error
     def refresh_vehicles(self, token: Token, vehicles: list[Vehicle] | Vehicle) -> None:
         """
         Refresh the vehicle data provided in get_vehicles.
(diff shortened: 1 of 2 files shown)


blka:
@cdnninja rebased onto current `master` (resolves the `dirty` merge state — conflicts were the #1160 `ApiImplSession` `self._session`→`self.session` rename and the #1216 py312 `list[Vehicle] | Vehicle` type syntax). CI is green on py3.12/3.13/3.14 + CodeQL + pre-commit.

Independent field confirmation in [kia_uvo#1685](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1685#issuecomment-5042666430) (comment by @dallen, today): ran the `fix/usa-session-expiry-recovery` branch for several weeks with zero occurrences of the `Config Not Ready: Token refresh failed` loop, and saw the expected `Session expired during get_vehicles, re-logging in` recovery; reverting to 3.8.0 reproduced the bug immediately, re-applying the branch cleared it again.

This is the USA analog of the already-merged #1136 (same class of token-expiry→`AuthenticationError` bug, EU/`ApiImplType1` side). `get_vehicles`/`refresh_vehicles` bypass `request_with_logging` (which already maps 1003/1005→`AuthenticationError` on master), so they need the token-only `_retry_on_auth_error` decorator — `request_with_active_session` can't be reused because it requires `vehicle`/`json_body` kwargs these methods don't have.

hyundai_kia_connect_api/KiaUvoApiUSA.py
```diff
@@ -102,6 +102,37 @@ def request_with_logging_wrapper(*args, **kwargs):
...
+                f"{DOMAIN} - Session expired during {func.__name__}, re-logging in"
+            )
+            new_token = self.login(
+                token.username,
+                token.password,
+                token,
+                getattr(self, "_otp_handler", None),
+            )
+            if not isinstance(new_token, Token):
+                raise AuthenticationOTPRequired("OTP required to refresh token")
```
cdnninja, line 130: Is this assuming OTP is needed for all failures?

hyundai_kia_connect_api/KiaUvoApiUSA.py
```diff
@@ -102,6 +102,37 @@ def request_with_logging_wrapper(*args, **kwargs):
...
+
+    Symmetric with request_with_active_session but for methods that only take
+    a token (get_vehicles / refresh_vehicles). No threading.Lock, matching
+    request_with_active_session's concurrency behavior.
+    """
+
+    def _retry_on_auth_error_wrapper(self, token, *args, **kwargs):
+        try:
+            return func(self, token, *args, **kwargs)
+        except AuthenticationError:
```
cdnninja, line 116: What happens if wrong username password? Auth error is probably the wrong exception here.

cdnninja (changes requested):

blka:
@cdnninja thanks for the catch — both points addressed in `4b14533`:

1. **OTP not assumed for all failures** — the `if not isinstance(new_token, Token)` catch-all was wrong. Narrowed to `isinstance(new_token, OTPRequest)`, so only a genuine `OTPRequest` return from `login()` raises `AuthenticationOTPRequired`. Other unexpected return types are no longer misclassified.

2. **Wrong username/password** — `KiaUvoApiUSA.login()` raises a bare `Exception` on bad credentials (no `sid`, no `otpKey`), which previously propagated unchanged through the retry path. The `self.login(...)` call is now wrapped so any login failure surfaces as `AuthenticationError("Re-login failed: ...")` with the original exception chained as `__cause__`, consistent with the auth-error semantics the coordinator already handles. The `except AuthenticationError` itself is still narrow — `get_vehicles`/`refresh_vehicles` only raise it for errorCode 1003/1005 (session invalid), so wrong-credentials never enters this path from the first call; it can only surface during the re-login retry.

Two tests added for the wrong-credentials re-login path (decorator + `get_vehicles`). `pytest tests/test_usa_session_expiry.py` → 12 passed. CI re-running on py3.12/3.13/3.14 + CodeQL.

Re-review welcome when CI goes green.

cdnninja:
You mention wrong creds don't throw an auth exception based on the error code. If wrong credentials where is an auth exception raised?  As this is how HA allows a user to fix it.

blka:
@cdnninja good push — you're right, that's the real gap. Today `KiaUvoApiUSA.login()` raises a bare `Exception` at the no-`sid`/no-`otpKey` fallthrough (line 365), not `AuthenticationError`. So wrong creds surface as a generic `Exception`: the coordinator maps that to `UpdateFailed` (60s retry loop), and the config flow doesn't catch it as `invalid_auth` either — the user can't fix it. CA (`7402/7403/7404/7602 → AuthenticationError`) and `HyundaiBlueLinkApiUSA` already map auth error codes; `KiaUvoApiUSA.login()` is the one that never inspects `status`.

Within this PR's retry path, `4b14533` already converts any re-login failure to `AuthenticationError` (so HA gets `ConfigEntryAuthFailed` there), but the initial `login()` / `check_and_refresh_token → refresh_access_token → login()` path still raises bare `Exception`. That's the place an auth exception should be raised.

Two things:

1. **What errorCode does USA `/prof/authUser` return for wrong credentials?** I don't have a live capture of a bad-creds response. If you have one from logs (the `status.errorCode`/`errorType`), I'll map it precisely like CA does. Otherwise the safe option is to raise `AuthenticationError` whenever `status.statusCode != 0` at the fallthrough (no `sid`, no `otpKey` = auth rejected in practice).

2. **Scope: include that `login()` fix in this PR, or a separate follow-up?** This PR is scoped to the `get_vehicles`/`refresh_vehicles` session-expiry path; touching `login()` changes behavior for every USA login call (initial setup included). I'm happy to add it here if you prefer a single fix, or split it out — your call.

cdnninja:
I'm good with it being a follow up pr.

PR #1198, opened 2026-06-22
feat: add HyundaiBlueLinkApiUSA.refresh_access_token override

### LINKED ISSUE #1186: Feature: refresh_access_token for Hyundai USA (BlueLink)
## Problem

`HyundaiBlueLinkApiUSA` has no `refresh_access_token` override. When the access token expires (typically every 24h), the integration does a full re-login via `self.login()`, which:

- Takes ~3s instead of ~0.2s for a refresh
- May require user interaction (especially if session expires)
- Causes unnecessary API load

The login response includes `refresh_token`, but it is never used for token renewal.

## Evidence

- `KiaUvoApiEU` already has a working `refresh_access_token` that uses `grant_type=refresh_token` and takes ~0.2s
- `ApiImplType1` (EU, AU, IN, CN) regions all have `_get_refresh_token()` methods that call `oauth2/token` with `grant_type=refresh_token`
- The USA Hyundai login endpoint (`/v2/ac/oauth/token`) returns `refresh_token` in the response — suggesting the OAuth2 refresh grant may be supported
- Issue #1715 proved that **CA does NOT support refresh** (the `refreshToken` field is dead), but USA Hyundai uses a different API (`/v2/ac/`) than CA (`/tods/api/v2/`)

## Proposal

Add `refresh_access_token` to `HyundaiBlueLinkApiUSA` that:

1. Calls `POST /v2/ac/oauth/token` with `grant_type=refresh_token` and the stored `refresh_token`
2. Returns a new `Token` with updated `access_token` and `refresh_token`
3. Falls back to `self.login()` if refresh fails

## Diagnostic script

I created a diagnostic script to test whether the refresh endpoint actually works:

**Branch:** `feat/usa-hyundai-refresh-token-diagnostic`
**File:** `scripts/hyundai_usa_refresh_diagnostic.py`

```bash
cd hyundai_kia_connect_api
pip install -e .
pip install python-dotenv requests
python scripts/hyundai_usa_refresh_diagnostic.py
```

Set `HYUNDAI_USA_USERNAME` and `HYUNDAI_USA_PASSWORD` in a `.env` file.

The script tests:
1. Full login → captures `refresh_token`
2. `POST /v2/ac/oauth/token` with `grant_type=refresh_token`
3. Verifies the new token works (vehicle list fetch)
4. Tries alternative approaches if standard refresh fails

## Request for testing

@ijojog — you tested the horn/hazard lights for USA Hyundai. Could you run this diagnostic script with your credentials? It will:

- Login to your account
- Test if token refresh works (no vehicle commands are sent)
- Report the results

No credentials are logged — only truncated token prefixes are shown.

## Related

- #1715 — CA auth expires (different API, refresh proven dead)
- PR #1182 — Control token caching for EU/AU/IN (different optimization)
- PR #1121 — EU headless login (different auth flow)

blka:
## What

Adds `HyundaiBlueLinkApiUSA.refresh_access_token`, so the USA Hyundai (BlueLink) integration can renew the access token without a full re-login when it expires.

## Why

`HyundaiBlueLinkApiUSA` had no `refresh_access_token` override, so `ApiImpl.refresh_access_token` fell back to a full `self.login()` on every token expiry. The login response already includes a `refresh_token`, but it was unused.

Issue #1186 was previously closed as wontfix because a first diagnostic concluded the refresh grant was unsupported. That conclusion was premature: every refresh attempt returned a **field-validation** error (`C400_1` "Parameter {password} field [Required]") — the endpoint validates `username`+`password` presence **before** inspecting `grant_type`, so the combination `username + password + grant_type=refresh_token + refresh_token` was never sent.

## How — confirmed live

A second diagnostic (issue #1186, @ijojog re-run 2026-06-22) tested a 3-call matrix against `POST /v2/ac/oauth/token`:

| Variant | Result |
|---|---|
| **T1** JSON: `{username, password, grant_type:"refresh_token", refresh_token}` + standard `API_HEADERS` | **HTTP 200 ✅** |
| T2 form-urlencoded, same fields | HTTP 400 `username required` |
| T3 `Authorization: Basic` + form, no username/password | HTTP 403 `client_id required` |

T1 wins. The refreshed `access_token` was verified against `GET /enrollment/details` → HTTP 200. Measured: **0.02s refresh vs 0.74s full login (44.6x faster)**, `refresh_token` rotates, `expires_in` ~1800s (30 min).

So the USA endpoint **does** honor `grant_type=refresh_token`, but only when sent as JSON with username + password + grant_type + refresh_token together, and with the custom `client_id`/`clientSecret` headers (not Basic auth, not form-urlencoded).

## Implementation

`refresh_access_token` POSTs to `/v2/ac/oauth/token` with the four-field JSON body and `self.API_HEADERS`, parses the new tokens, and returns a new `Token` preserving `username`/`password`/`pin`. `valid_until` is `now + expires_in`. Falls back to `self.login()` when:
- `token.refresh_token` is missing/empty
- the POST raises (network error)
- the response carries an `errorCode` (`_check_response_for_errors` raises `APIError`)
- the response has no `access_token`

This mirrors `KiaUvoApiEU.refresh_access_token` fallback behaviour.

## Tests

10 unit tests in `tests/us_refresh_access_token_test.py` (mocked `sessions.post`, no network):
- success path returns rotated credentials, preserves username/password/pin
- **body contract**: POST body contains `username`, `password`, `grant_type:"refresh_token"`, `refresh_token` (locks in the four-field requirement)
- `refresh_token` rotation and preservation when response omits it
- `valid_until` derived from `expires_in`
- fallback to `login` on: missing/None refresh_token, POST exception, error response
- `login` not called on success

## Closes

Closes #1186.

## Notes

- No `kia_uvo` manifest bump needed here — the release pipeline bumps the API requirement automatically.
- Cross-region check: PIN is not used in refresh by any region (PIN is only for vehicle commands), so no PIN field is involved.

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
@@ -222,6 +222,66 @@ class HyundaiBlueLinkApiUSA(ApiImpl):
             pin=pin,
         )
 
+    def refresh_access_token(self, token: Token) -> Token:
+        """Refresh the access token using the stored refresh token.
+
+        Calls the same /v2/ac/oauth/token endpoint as login, but sends
+        username + password + grant_type=refresh_token + refresh_token
+        together in a JSON body. The endpoint validates username/password
+        presence before inspecting grant_type, so all four fields must be
+        sent (sending grant_type + refresh_token alone returns 400
+        "password required"). Falls back to full login if the refresh
+        token is missing or the exchange fails.
+
+        Confirmed working live (2026-06-22, issue #1186): ~0.02s refresh
+        vs ~0.74s full login, refresh_token rotates, expires_in ~1800s.
+        """
+        if token.refresh_token:
+            try:
+                url = self.LOGIN_API + "oauth/token"
+                data = {
+                    "username": token.username,
+                    "password": token.password,
+                    "grant_type": "refresh_token",
+                    "refresh_token": token.refresh_token,
+                }
+                response = self.session.post(url, json=data, headers=self.API_HEADERS)
+                response_json = response.json()
+                _check_response_for_errors(response_json)
+                if response_json.get("access_token") is None:
+                    raise APIError(
+                        f"{DOMAIN} - refresh_access_token: no access_token "
+                        f"in response {response_json}"
+                    )
+                access_token = response_json["access_token"]
+                new_refresh_token = response_json.get(
+                    "refresh_token", token.refresh_token
+                )
+                expires_in = float(response_json["expires_in"])
+                valid_until = dt.datetime.now(dt.timezone.utc) + dt.timedelta(
+                    seconds=expires_in
+                )
+                _LOGGER.debug(
+                    f"{DOMAIN} - Access token refreshed successfully, "
+                    f"expires in {expires_in}s"
+                )
+                return Token(
+                    username=token.username,
+                    password=token.password,
+                    access_token=access_token,
+                    refresh_token=new_refresh_token,
+                    valid_until=valid_until,
+                    pin=token.pin,
+                )
+            except Exception:
+                _LOGGER.warning(
+                    f"{DOMAIN} - Refresh token exchange failed, "
+                    f"falling back to full login"
+                )
+                return self.login(token.username, token.password, token.pin)
+        # No refresh_token available — full login.
+        return self.login(token.username, token.password, token.pin)
+
     def _get_vehicle_details(self, token: Token, vehicle: Vehicle):
         url = self.API_URL + "enrollment/details/" + token.username
         headers = self._get_authenticated_headers(token)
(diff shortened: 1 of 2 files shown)


hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
```diff
@@ -223,6 +223,64 @@ def login(
...
+                    password=token.password,
+                    access_token=access_token,
+                    refresh_token=new_refresh_token,
+                    valid_until=valid_until,
+                    pin=token.pin,
+                )
+            except Exception:
+                _LOGGER.warning(
+                    f"{DOMAIN} - Refresh token exchange failed, "
+                    f"falling back to full login"
```
cdnninja, line 279: Based on indenting how are we falling back to full login?

blka: Fair point — the fallback was relying on fall-through (the `except` only logged, no return, so control dropped past the `if` to the method-level `return self.login(...)`). Correct but not obvious from the indent alone, especially in a narrow diff view.

Made it explicit in bce20f0: the `except` now has its own `return self.login(...)` so both fallback paths visibly return — exchange-failure (inside `except`) and missing/empty `refresh_token` (the `if` is skipped, final return). Added a one-line comment on the final return to mark the no-refresh_token path. No more fall-through. Tests still green (10/10).

PR #1209, opened 2026-06-30
fix(Vehicle): coerce air_temperature/outside_temperature to float (root cause of kia_uvo #1755)

blka:
## Summary

Root-cause API fix for https://github.com/Hyundai-Kia-Connect/kia_uvo/pull/1755.

The HA climate entity crashes the coordinator listener update with `TypeError: Temperature is not a number: 72` because Home Assistant's `display_temp()` rejects non-numeric `current_temperature`. The root cause is in this library: USA (and BR) backends assign `vehicle.air_temperature` without casting to `float`, unlike `ApiImplType1` which does `float(air_temp)`. The string then flows into the `Vehicle` object and onward to HA.

## Root cause

- `ApiImplType1.py` sets `vehicle.air_temperature = (float(air_temp), unit)` — EU/CCS2 paths cast to float and do not crash.
- `KiaUvoApiUSA.py` and `HyundaiBlueLinkApiUSA.py` set `vehicle.air_temperature = (air_temp, unit)` — **no cast**; the raw string (`"72"`) reaches HA.
- `HyundaiBlueLinkApiBR.py` sets `(temp_value, temp_unit)` — same non-cast pattern.
- `Vehicle.air_temperature` setter stored `value[0]` as-is (only special-casing `"OFF"` → `None`); `Vehicle.outside_temperature` stored `value[0]` as-is. No numeric coercion at the choke point.

kia_uvo #1755 is a workaround on the HA side (`climate.py` coerces to float). This PR fixes the source of truth so every consumer (HA climate, HA sensor, and any other) receives a numeric value regardless of which backend populated it.

## Changes

- **`utils.py`** — new `float_or_none(value)` helper: coerces to `float`, returns `None` for `None` / non-numeric / `"OFF"` / empty.
- **`Vehicle.py`** — `air_temperature` and `outside_temperature` setters now store `float_or_none(value[0])` in the public backing field. The diagnostic `_value` fields stay raw (unchanged contract); `_unit` handling unchanged.
- No backend files modified — the setter is the single choke point every backend passes through.

> Note: a ruff-format quirk in 0.15.18 strips parentheses from `except (TypeError, ValueError):`, producing `except TypeError, ValueError:` which is a `SyntaxError` on Python 3.10–3.13. The `# fmt: skip` on that line keeps the parens intact across `ruff format` / pre-commit.

## Testing

- `tests/utils_test.py` — 7 unit tests for `float_or_none`.
- `tests/test_vehicle_temperature_setters.py` (new) — 8 setter tests covering numeric-string → float, `"OFF"` → `None`, non-numeric → `None`, and that `_value` stays raw.
- `tests/test_usa_vehicle_properties.py` — regression test proving the USA end-to-end path turns `airTemp.value = "72"` into `vehicle.air_temperature == 72.0` (float).
- `tests/__snapshots__/test_vehicle_snapshots.ambr` — updated USA snapshots (`'air_temperature': '72'` → `72.0`). Updated via `pytest --snapshot-update`; only the 3 USA `air_temperature` lines changed.

## Pre-existing snapshot failures (not from this PR)

`test_vehicle_snapshots.py` has 8 failures on `upstream/master` already: PR #1207 added `supports_valet_mode` / `valet_mode_active` to `Vehicle` but did not regenerate the syrupy snapshot file. This PR does not touch those fields. A separate PR (`fix/vehicle-snapshots-valet-fields`) regenerates the snapshots and turns the suite green; it should land alongside or before this one.

## Out of scope

- **Other temperature fields** (`ev_battery_temperature_min/max/water`, `ev_first/second_departure_climate_temperature`) — same setter pattern but they are numeric sensors without HA `display_temp()` routing and have no crash reports. Not changed; can be hardened separately if needed.
- **kia_uvo PR #1755** itself — the HA-side workaround. Once this API fix is released and `kia_uvo` bumps its dependency, #1755 becomes a redundant defensive layer and can be kept or closed separately.
- Changing `get_float`'s contract or removing the now-redundant `float(air_temp)` in `ApiImplType1` (left in place — surgical, no functional effect).


hyundai_kia_connect_api/utils.py
```diff
@@ -56,6 +56,16 @@ def get_float(value):
     return value  # original fallback
 
 
+def float_or_none(value):
```
cdnninja: Add type hints.

cdnninja (changes requested):

hyundai_kia_connect_api/utils.py
blka: Done — added type hints: `def float_or_none(value: str | int | float | None) -> float | None:`. Pushed in f122901.

PR #1214, opened 2026-07-02
fix: return None for unset EV timers in _get_time_from_string (#1206)

### LINKED ISSUE #1206: _get_time_from_string crashes on "0000" (unset EV departure/off-peak timer)
This bug applies to GENESIS EU INtegration:

## Summary

`ApiImplType1._get_time_from_string` raises `ValueError: time data '0000' does not match format '%I%M'` whenever the vehicle reports an unset (zero-valued) EV departure timer or off-peak time window. This makes the integration unusable for any EU vehicle that has no scheduled departure/charging timer configured.

## Affected versions

Confirmed in **every release from 4.13.0 through 4.21.0** (current). The bug has now survived at least nine library releases without being touched.

## Affected file / location

`hyundai_kia_connect_api/ApiImplType1.py`, method `_get_time_from_string` (line 201 in v4.21.0).

It is triggered from `KiaUvoApiEU.py::_update_vehicle_properties` for these fields:

- `ev_first_departure_time`
- `ev_second_departure_time`
- `ev_off_peak_start_time`
- `ev_off_peak_end_time`

## Reproduction

1. Configure the integration with **brand=Genesis, region=Europe** (Genesis GV60 in my case, but any EU EV without a scheduled timer should reproduce it).
2. Make sure the vehicle has **no departure timer configured** in the official app — the API then returns `"0000"` for `evFirstDepartureTime` (and friends).
3. Let the coordinator update the vehicle. It crashes with:

```
File ".../hyundai_kia_connect_api/VehicleManager.py", line 143, in update_vehicle_with_cached_state
    self.api.update_vehicle_with_cached_state(self.token, vehicle)
File ".../hyundai_kia_connect_api/KiaUvoApiEU.py", line 379, in update_vehicle_with_cached_state
    self._update_vehicle_properties(...)
File ".../hyundai_kia_connect_api/KiaUvoApiEU.py", line 770, in _update_vehicle_properties
    vehicle.ev_first_departure_time = self._get_time_from_string(...)
File ".../hyundai_kia_connect_api/ApiImplType1.py", line 208, in _get_time_from_string
    d = dt.datetime.strptime(str(value), "%I%M")
File "/usr/local/lib/python3.14/_strptime.py", line 555, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
ValueError: time data '0000' does not match format '%I%M'
```

## Root cause

For `value = "0000"`:

- `int(value[-2:]) == 0`, so the `> 60` branch is skipped.
- `int(value) == 0`, which is not `> 1260`, so the code falls into the `else` branch.
- `datetime.strptime("0000", "%I%M")` then raises, because `%I` requires hours `01..12` — `00` is invalid in 12-hour format.

So the function has no handling at all for the "no timer set" case that the API actually returns in production.

## Proposed fix

Treat `"0000"` / `0` / empty string as "no scheduled timer" and return `None`, and wrap the rest in a try/except so an unexpected value can never crash the entire vehicle update:

```python
def _get_time_from_string(self, value, timesection) -> dt.datetime.time | None:
    if value is None:
        return None
    try:
        # "0000", "0", or empty means "no scheduled timer" -> return None
        if not str(value).strip() or int(value) == 0:
            return None
        lastTwo = int(value[-2:])
        if lastTwo > 60:
            value = int(value) + 40
        if int(value) > 1260:
            return dt.datetime.strptime(str(value), "%H%M").time()
        d = dt.datetime.strptime(str(value), "%I%M")
        if timesection and timesection > 0:
            d += dt.timedelta(hours=12)
        return d.time()
    except (ValueError, TypeError):
        return None
```

The callers in `KiaUvoApiEU._update_vehicle_properties` already write the result straight onto `vehicle.ev_*_departure_time` / `ev_off_peak_*_time`, so returning `None` is the natural "no value" semantics and doesn't need any caller changes.

## Workaround currently in use

I'm runtime-monkey-patching `ApiImplType1._get_time_from_string` from the consuming custom component (`kia_uvo`'s `__init__.py`) with the implementation above. It has worked reliably across nine consecutive library releases (4.13.0 → 4.21.0).

Happy to open a PR if a maintainer can confirm the proposed semantics (`"0000"` → `None`) are correct.

blka:
Closes #1206.

## Problem

`ApiImplType1._get_time_from_string` raised `ValueError: time data "0000" does not match format "%I%M"` whenever the vehicle reported an unset (zero-valued) EV departure timer or off-peak window. The API returns `"0000"` for `evFirstDepartureTime` (and friends) when no timer is configured. This crashed `update_vehicle_with_cached_state` and made the integration unusable for any EU EV without a scheduled timer.

Confirmed across every release 4.13.0 to 4.21.0 (current). Root cause: `%I` requires hours 01..12; `00` is invalid in 12-hour format, and the function had no handling for the "no timer set" case.

## Scope

All Type1 regions using this code path, not just EU as reported:

| Region | File | Format | Crashed on "0000" |
|---|---|---|---|
| EU + AU | `ApiImplType1.py` (base) | `%I%M` | yes |
| IN | `KiaUvoApiIN.py` (bit-for-bit identical override) | `%I%M` | yes |
| CN | `KiaUvoApiCN.py` | `%I%M %p` | yes |

## Changes

- `ApiImplType1._get_time_from_string` — treat `"0000"` / `"0"` / `""` / `0` / `None` as unset -> `None`; guard `timesection=None` (preexisting `TypeError` when `timeSection` absent from the response); wrap the parse in `try/except (ValueError, TypeError)` with `_LOGGER.warning` so an unexpected format never crashes the whole vehicle update. Return type updated to `dt.datetime.time | None` (PEP 604, matches `Vehicle.py`).
- `KiaUvoApiIN._get_time_from_string` — deleted (bit-for-bit identical to base); IN now inherits the fixed base. Reduces duplication.
- `KiaUvoApiCN._get_time_from_string` — same guard + try/except + warning applied to the `%I%M %p` override. `timesection=None` falls to the AM default branch instead of crashing.

Callers in `KiaUvoApiEU._update_vehicle_properties` (and AU/CN equivalents) assign the result straight onto `vehicle.ev_*_departure_time` / `ev_off_peak_*_time`, which are already `Union[datetime.time, None]` — so `None` is the natural "no value" semantics and no caller changes were needed.

## Why None is correct

`"0000"` is what the API returns in production for an unset timer (the reporter has validated this workaround across 9 releases). `Vehicle` already types these fields as `Optional[datetime.time]`, and HA renders `None` as "no scheduled timer". The reporter runtime monkey-patch (in `kia_uvo` `__init__.py`) uses exactly this semantics.

## Testing

- New `tests/test_get_time_from_string.py` — isolated unit tests for base (EU/AU/IN) and CN: unset values -> `None`, real timers, `timesection=None`, unexpected format -> `None`.
- New `tests/fixtures/eu_kia_ev6_unset_timers.json` — copy of the EU EV6 fixture with all 4 EV timers set to `"0000"`; reproduces the issue crash path through `_update_vehicle_properties`.
- New `test_ev_timers_unset_when_api_returns_zero` — full-path assertion that all 4 timer fields are `None` and the update does not raise.
- `tests/__snapshots__/test_vehicle_snapshots.ambr` — regenerated (syrupy) for the new fixture; captures `None` for all 4 timer fields.

## Verification

- `pytest tests/ --ignore=tests/test_headless_login.py -m "not integration"` -> 287 passed, 9 snapshots passed
- `pre-commit run --all-files` -> all hooks pass (ruff, codespell, pyupgrade, yamllint, prettier, mypy --strict)
- `ruff check --target-version py310` -> clean

## Backward compatibility

Behavior is unchanged for valid timers (`"0800"`, `"2200"`, `"1300"`, etc.). The only behavior change: previously-crashing inputs now return `None`. No caller changes required. No `kia_uvo` manifest change in this PR (the downstream bump is pipeline-driven).


hyundai_kia_connect_api/KiaUvoApiCN.py
```diff
@@ -220,20 +220,30 @@ def get_vehicles(self, token: Token) -> list[Vehicle]:
             result.append(vehicle)
         return result
 
-    def _get_time_from_string(self, value, timesection) -> dt.datetime.time:
-        if value is not None:
+    def _get_time_from_string(self, value, timesection) -> dt.datetime.time | None:
```
cdnninja: Could this just be removed to reduce duplicate code since this is also a type1?

cdnninja (changes requested):

hyundai_kia_connect_api/KiaUvoApiCN.py
blka: Good catch — done in 2db80fd. Verified the base `ApiImplType1._get_time_from_string` produces identical results to the CN override for every CN API input (`timeSection` ∈ {0,1}, confirmed against `cn_kia_ev6_2024_with_soc.json`):

| input | base | CN override (was) |
|---|---|---|
| `0800` ts=0 | 08:00 | 08:00 |
| `0800` ts=1 | 20:00 | 20:00 |
| `2200` ts=0/1 | 22:00 | 22:00 |
| `0600` ts=0/1 | 06:00 / 18:00 | 06:00 / 18:00 |
| `0000` ts=* | None | None |

The only divergence is `timeSection` outside {0,1} (e.g. 2), which the API never sends and which **previously crashed in both implementations** — so inheriting base is a strict improvement (crash → 20:00 instead of crash → 08:00), not a behavior change for any real input.

CN now inherits the fixed base, matching the IN override removal earlier in this PR. Dropped the now-redundant `TestGetTypeFromStringCN` — CN is covered by the base parametrized tests via inheritance. Full suite green (277 passed, 9 snapshots), pre-commit + mypy --strict clean.

PR #1219, opened 2026-07-05
fix(us): detect Hyundai PHEV via enrollment evStatus=='P'

blka:
## Problem

`HyundaiBlueLinkApiUSA.get_vehicles` only mapped enrollment `evStatus` values `"N"` (ICE) and `"E"` (EV). Plug-in hybrids (Tucson PHEV, Santa Fe PHEV, IONIQ PHEV) report `evStatus == "P"` in enrollment details, so their `engine_type` was left as `None` — PHEVs were misclassified.

## Fix

Add the `"P"` branch mapping to `ENGINE_TYPES.PHEV`:

```python
elif entry["evStatus"] == "P":
    entry_engine_type = ENGINE_TYPES.PHEV
``"

`ENGINE_TYPES.PHEV` already exists (`const.py`) and is already used by `KiaUvoApiUSA` and `ApiImplType1` (EU).

## Behavioral impact

- **EV sensors**: `_update_vehicle_properties` already parses `vehicleStatus.evStatus.*` unconditionally (not gated on `engine_type == EV`), so PHEV users already received EV sensors when the `evStatus` block was present. No change.
- **Commands**: `start_climate`, `start_charge`, `set_charge_limits` are gated on `engine_type == EV`; PHEV routes to the ICE climate path (`rcs/rsc/start`). Before this fix, PHEV had `engine_type=None` → also used the ICE path. **No regression** — same command path before and after.

## Tests

New `tests/us_phev_detection_test.py` (4 tests): `"P"` → PHEV, `"E"` → EV (regression guard), `"N"` → ICE (regression guard), unknown → `None` (regression guard). All green; full suite green (`pytest tests/`). Ruff + pre-commit green.

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
@@ -923,6 +923,8 @@ class HyundaiBlueLinkApiUSA(ApiImpl):
                 entry_engine_type = ENGINE_TYPES.ICE
             elif entry["evStatus"] == "E":
                 entry_engine_type = ENGINE_TYPES.EV
+            elif entry["evStatus"] == "P":
+                entry_engine_type = ENGINE_TYPES.PHEV
             vehicle: Vehicle = Vehicle(
                 id=entry["regid"],
                 name=entry["nickName"],
--- a/tests/us_phev_detection_test.py
+++ b/tests/us_phev_detection_test.py
@@ -1,0 +1,81 @@
+"""Tests for PHEV detection in HyundaiBlueLinkApiUSA.get_vehicles.
+
+Hyundai USA enrollment details return evStatus == "P" for plug-in hybrids.
+get_vehicles must map this to ENGINE_TYPES.PHEV (previously left as None,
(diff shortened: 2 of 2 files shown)


cdnninja:
To confirm you found P in logs?

blka:
Confirmed on a real 2023 Hyundai Santa Fe Plug-In Hybrid (not my own vehicle — I don't have a PHEV, so I can't produce logs from my own account).

Enrollment `details` returns `evStatus: "P"` for that vehicle. After mapping `P` → `ENGINE_TYPES.PHEV`, it surfaces both EV-side entities (ev_battery 100%, plugged_in on, charging switch, EV range) and the fuel sensor (65%) — ~5 new entities total, no regression for `E` (full EV) or `N` (ICE) since the branch is purely additive.

The PR's test is mocked (`_enrollment_response(ev_status="P")`), not fixture-based. If you can share (or point me to) a sanitized USA Hyundai PHEV enrollment dump, I'll convert `us_phev_detection_test.py` to a fixture-based test matching `eu_kia_sportage_phev_2027_ccs2.json`. Otherwise the mock test + this live confirmation is what I can stand behind.

PR #1220, opened 2026-07-05
fix(us): fail-fast timeout on findMyCar location request

blka:
## Problem

`HyundaiBlueLinkApiUSA._get_vehicle_location` called `self.session.get(url, headers=headers)` with the session default timeout `(connect=10, read=30)`. If the `rcs/rfc/findMyCar` endpoint hangs, the coordinator update blocks for up to 30s on read timeout.

The method already has `try/except Exception` returning `None` on failure, and the coordinator already falls back to cached `vehicleLocation` when location is `None`. So this is purely a fail-faster concern, not a correctness fix.

## Fix

Pass a per-call timeout on the findMyCar request only:

```python
response = self.session.get(url, headers=headers, timeout=(5, 10))
```

- `connect=5` (vs session 10), `read=10` (vs session 30).
- Only this one call. Other `session.get` calls in the file keep the session default.
- On timeout, `requests.exceptions.Timeout` is caught by the existing `except Exception` → returns `None` → coordinator uses cached `vehicleLocation`.
- **No retry added** — retrying would wake the vehicle and risk 12V battery drain. Fail-faster only.

## Tests

New `tests/us_findmycar_timeout_test.py` (2 tests): asserts `session.get` is called with `timeout=(5, 10)`; asserts a `Timeout` exception results in `None` (graceful). Full suite green. Ruff + pre-commit green.

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkApiUSA.py
@@ -350,7 +350,7 @@ class HyundaiBlueLinkApiUSA(ApiImpl):
         url = self.API_URL + "rcs/rfc/findMyCar"
         headers = self._get_vehicle_headers(token, vehicle)
         try:
-            response = self.session.get(url, headers=headers)
+            response = self.session.get(url, headers=headers, timeout=(5, 10))
             response_json = response.json()
             _check_response_for_errors(response_json)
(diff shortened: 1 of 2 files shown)


cdnninja:
Do we have evidence of this happening?

cdnninja:
@blka  Can this be closed out?

blka:
Closing per maintainer request — keeping the session default read timeout for now; happy to revisit if blocking on findMyCar is ever reported in the field.

PR #1238, opened 2026-07-16
fix(ca): treat errorCode 7606 (IP validation) as an invalid token

BenBaril83:
CA binds an access token to the IP address that requested it, and rejects a token used from a different IP with errorCode 7606 rather than expiring it normally (e.g. an ISP lease renewal around a nightly Home Assistant restart). test_token() only recognized 7403/7602 as invalidating errors, so it reported the token as still good. check_and_refresh_token() then skipped the refresh/re-login path and fell through to initialize_vehicles(), which made the same doomed call again via get_vehicles() -- which does not tolerate the error and raises an unhandled APIError, crashing config entry setup (kia_uvo#1715).

Add 7606 to test_token()'s token_errors so check_and_refresh_token() takes its existing refresh/re-login path instead, reusing the persisted device_id (kia_uvo#1715 / PR #1212) to get a token bound to the current IP.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -431,7 +431,13 @@ class KiaUvoApiCA(ApiImpl):
         response = self.sessions.post(url, headers=headers)
         _LOGGER.debug(f"{DOMAIN} - Test Token Response {response.text}")
         response = response.json()
-        token_errors = ["7403", "7602"]
+        # 7403/7602: token expired/deleted. 7606: CA binds the access token to
+        # the IP that issued it, and rejects it outright if the caller's IP has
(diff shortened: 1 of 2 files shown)


blka:
Nice catch — adding `7606` to `test_token()` is the right gate: it routes IP-validation failures through the existing refresh/re-login path and reuses the persisted `device_id` from #1212, so the new token binds to the current IP transparently.

One thing worth recording for future readers: `7606` is correctly **not** added to `error_code_mapping` in `_check_response_for_errors`. Mapping it to `AuthenticationError` would surface as `ConfigEntryAuthFailed` in the kia_uvo coordinator and prompt a reauth UI — but here recovery is automatic (re-login with the persisted device_id), so the current `APIError → UpdateFailed → retry → test_token catches 7606 → re-login` path is the right one. All good from my side.

PR #1240, opened 2026-07-17
fix(BR): read cached status from the CCS2 endpoint (fixes 503 resCode 5031)

paulogabrielfs:
## Problem

Status polling for Brazilian (BR) vehicles fails with a **503** on every update, seen in Home Assistant (kia_uvo) as repeated `Force update failed` / `Cached update failed` errors while the official Bluelink app keeps working:

```
GET /api/v1/spa/vehicles/{id}/status/latest
-> 503 {"retCode":"F","resCode":"5031","resMsg":"Unavailable remote control - Service Temporary Unavailable"}
```

BR vehicles report `ccuCCS2ProtocolSupport: 0`, so `_get_vehicle_state` used the legacy `/status/latest` endpoint. On the BR backend that endpoint is the **remote-control command result**, not a status read — it returns 503 / resCode `5031` until a remote command has been issued.

## What actually works on BR

Tested live against a non-CCS2 Creta 2025 (`type: GN`, `ccuCCS2ProtocolSupport: 0`):

| Endpoint | Result |
|---|---|
| `GET /status/latest` | **503** resCode 5031, always |
| `GET /status` | 200, but **wakes the car** (~20s), legacy flat payload |
| `GET /ccs2/carstatus/latest` | **200, instant, no wake** — CCS2 `state.Vehicle`, with a `lastUpdateTime` |
| `GET /ccs2/carstatus` | ack only; ~10s later the car reports and `/ccs2/carstatus/latest` advances (async force) |

So even though the flag is `0`, BR serves the status in **CCS2** format. That's what the app reads for the passive view without waking the car, and it even includes data the flat `/status` payload lacks (e.g. odometer).

## Fix

- cached read (no wake) → `GET /ccs2/carstatus/latest`
- force refresh (wakes the car, async) → trigger `GET /ccs2/carstatus`, then poll `/ccs2/carstatus/latest` until `lastUpdateTime` advances (with a fallback to the last cached snapshot)

To parse the CCS2 payload without duplicating ~500 lines, `HyundaiBlueLinkApiBR` now extends `ApiImplType1` and reuses its `_update_vehicle_properties_ccs2`. BR keeps overriding its own auth, headers, `get_vehicles`, endpoint selection and the async force-refresh flow; `ApiImplType1` is untouched. Valet mode is explicitly disabled for BR.

## Tests

`tests/test_br_vehicle_properties.py`:
- cached read uses `/ccs2/carstatus/latest` (never `/status*`)
- async force refresh: trigger + poll until fresh, and the fallback when the car doesn't report in time
- CCS2 property parsing (battery, engine, locks, fuel, range, odometer, location, last_updated_at)

Backed by an anonymized `br_hyundai_creta_2025_ccs2_cached.json` fixture (status data only — no VIN/location/tokens). All new tests pass; `flake8`/`black` clean. Verified end-to-end against the live API: cached read populates without waking the car, and force refresh returns fresh data in ~26s.

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkApiBR.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkApiBR.py
@@ -10,25 +10,23 @@ from time import sleep
 from urllib.parse import parse_qs, urljoin, urlparse
 
 from .ApiImpl import (
-    ApiImpl,
     ApiImplSession,
     ClimateRequestOptions,
     WindowRequestOptions,
 )
+from .ApiImplType1 import ApiImplType1
 from .const import (
     BRAND_HYUNDAI,
     BRANDS,
-    DISTANCE_UNITS,
     DOMAIN,
     ENGINE_TYPES,
     ORDER_STATUS,
-    SEAT_STATUS,
     VEHICLE_LOCK_ACTION,
     WINDOW_STATE,
 )
 from .exceptions import APIError, AuthenticationError
 from .Token import Token
-from .utils import get_index_into_hex_temp, normalize_battery_soc, parse_date_br
+from .utils import get_index_into_hex_temp
 from .Vehicle import DayTripCounts, DayTripInfo, MonthTripInfo, TripInfo, Vehicle
 
 _LOGGER = logging.getLogger(__name__)
@@ -53,11 +51,24 @@ _SIGNIN_STEP_MESSAGES = {
 }
 
 
-class HyundaiBlueLinkApiBR(ApiImpl):
-    """Brazilian Hyundai BlueLink API implementation."""
+class HyundaiBlueLinkApiBR(ApiImplType1):
+    """Brazilian Hyundai BlueLink API implementation.
+
+    Extends ApiImplType1 to reuse its CCS2 status parser
+    (``_update_vehicle_properties_ccs2``). BR overrides all of its own auth,
+    headers, endpoint selection and force-refresh flow below; only the CCS2
+    parsing is inherited.
+    """
 
     supports_window_control: bool = True
+    # BR does not implement valet mode; keep it off (ApiImplType1 defaults True).
+    supports_valet_mode: bool = False
     data_timezone = dt.timezone(dt.timedelta(hours=-3))  # Brazil (BRT/BRST)
+
+    # Async force-refresh polling: how long to wait for the vehicle to report
+    # a fresh snapshot to /ccs2/carstatus/latest after triggering /ccs2/carstatus.
+    _FORCE_REFRESH_POLL_INTERVAL = 5  # seconds between polls
+    _FORCE_REFRESH_MAX_POLLS = 12  # ~60s total
 
     def __init__(self, region: int, brand: int, language: str = "pt-BR"):
         if BRANDS[brand] != BRAND_HYUNDAI:
@@ -235,7 +246,7 @@ class HyundaiBlueLinkApiBR(ApiImpl):
         response = self.session.get(url, headers=headers)
         response.raise_for_status()
         response_data = response.json()
-        _LOGGER.debug(f"{DOMAIN} - Got vehicles response")
+        _LOGGER.debug(f"{DOMAIN} - Got vehicles response: {response_data}")
         if "resMsg" not in response_data or "vehicles" not in response_data.get(
             "resMsg", {}
         ):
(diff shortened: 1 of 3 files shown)


blka:
Are you sure, that /status returns cached data? In EU it's used for force refresh and it's awakens the car

paulogabrielfs:
@blka You're right. I dug into it and /status does wake the car. 
What I found testing live against my Creta (ICE, ccuCCS2ProtocolSupport: 0):
`GET /status: 200`, but wakes the car (~20s when asleep), returns the legacy flat resMsg, It's a sync force.
`GET /status/latest`: 503 resCode 5031 "Unavailable remote control", consistently, even right after a successful force. On BR this is the remote-control command result endpoint, not a status read.
`GET /ccs2/carstatus/latest`: 200, instant, does not wake the car, returns the CCS2 state.Vehicle structure with a lastUpdateTime.
`GET /ccs2/carstatus`: returns just an ack ({retCode:S, resCode:0000}), then ~10s later the car reports and /ccs2/carstatus/latest's lastUpdateTime advances, so this is an async force trigger.

So I was wrong, I'll drop it.

The interesting part: even though the vehicle reports ccuCCS2ProtocolSupport: 0, BR serves the cached status through the CCS2 endpoint /ccs2/carstatus/latest. That's probably what the official app reads for the passive status view without waking the car, and it even includes the odometer, which the flat /status payload doesn't.

So I think the correct model for BR looks like the EU CCS2 path:

`GET /ccs2/carstatus/latest`: cached read (no wake)
`GET /ccs2/carstatus`: force refresh (wakes car, async), /ccs2/carstatus (ack), then poll /ccs2/carstatus/latest until lastUpdateTime advances

The payload is the standard CCS2 state.Vehicle shape, so it maps directly onto the existing [_update_vehicle_properties_ccs2](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/6a82a89a99363cf26ea8a0b12bffd4323cc49960/hyundai_kia_connect_api/ApiImplType1.py#L287-L801) in `ApiImplType1`.

But my question before I rework this: would it be best if I move _update_vehicle_properties_ccs2 (https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/6a82a89a99363cf26ea8a0b12bffd4323cc49960/hyundai_kia_connect_api/ApiImplType1.py#L287-L801) and the helpers into a shared base/mixin so BR can reuse it, or keep BR self-contained with its own CCS2 parser? BR currently extends ApiImpl, not ApiImplType1, so reuse needs one of those.

blka:
Dig with claude over the other regions. I believeve one of use the same logic you just found with using ccs2 endpoint even if car is not ccs2, then you can mirror solution to br. Sorry, writing from my mobile right now, can't check exactly

blka:
As br uses apiimpltype1,you should be able to overwrite the needed part in br and do not touch apiimpltype1 which is shared for few regions

paulogabrielfs:
I checked all the region implementations (ApiImplType1, EU, AU, CN, IN) and none of them use the CCS2 endpoint for a non-CCS2 car. They all branch strictly on the flag (ccu_ccs2_protocol_support == 0 → /status/latest, else /ccs2/carstatus/latest) and read the flag straight from get_vehicles without overriding it. So there's no existing "use CCS2 even when the flag is 0" path to mirror.

I confirmed the BR server genuinely returns ccuCCS2ProtocolSupport: 0 for this car (it's not a misread), yet only the CCS2 endpoints return status, /status/latest is the remote-control result endpoint on BR and always 503s. So BR really is the first region where a car reporting 0 still has to use the CCS2 endpoints.

The current PR handles this by always using the CCS2 endpoints in BR (cached /ccs2/carstatus/latest, plus the async force via /ccs2/carstatus + poll), and reusing _update_vehicle_properties_ccs2 by extending ApiImplType1. ApiImplType1 and the other regions are untouched. If you'd rather gate this behind something other than "always CCS2 for BR" (e.g. a BR-specific flag), happy to adjust, but since every BR vehicle we've seen reports 0, keying off the flag wouldn't help here.

blka:
@cdnninja look like it's the second solution for broken force refresh in ApiImplType1 but BR way only. @paulogabrielfs pls see https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1184

paulogabrielfs:
Agreed, my BR force_refresh_vehicle_state is essentially a second copy of the trigger+poll loop [#1184](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1184) moves into ApiImplType1, so it makes sense to converge on that. Since your own control test in #1184 also confirms /ccs2/carstatus/latest doesn't wake the car, which is exactly what BR relies on for the cached read

Two ways to sequence it:
If [#1184](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1184) lands soon, I'll rebase this on top and switch BR's force refresh over to the base _force_refresh_vehicle_state_ccs2 + hooks, dropping the BR-only loop, that resolves the force-refresh part cleanly via your solution.

If [#1184](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1184) isn't going to merge shortly and it's not a blocker for you, I'd suggest we move #1240 forward as-is (it already fixes the 503 and gives BR working status), and once [#1184](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1184) is merged I'll come back and refactor the force path to reuse the base loop + drop the duplication.
Either way the cached read (/ccs2/carstatus/latest) stays BR-specific, and ApiImplType1 isn't touched by this PR

blka:
@paulogabrielfranca see #1250 as a replacement of #1184

blka:
Superseded by #1259 — it adopts this PR's core: BR extends `ApiImplType1`, cached read from `/ccs2/carstatus/latest` reusing `_update_vehicle_properties_ccs2`, `supports_valet_mode = False`, plus the anonymized Creta fixture and tests. Credit to @paulogabrielfs for the CCS2-endpoint investigation and the fixture.

One difference: force-refresh is rebuilt on the merged #1250 wake+sleep+read pattern (rather than this PR's poll-until-timestamp loop), with a `lastUpdateTime` guard that raises `APIError` instead of applying a stale snapshot if the car didn't report in time. The 4xx signin case (HTTP 400 on `/user/signin`, not covered here) is also fixed in #1259.

Closing this in favor of #1259.

paulogabrielfs:
@blka Sounds good, happy to have this land in #1259, thanks for carrying it forward and for the credit.

I'm also happy to live-validate #1259 on my Creta, both the 400 signin path and the force-refresh timing before it merges. 

I'll close this in favor of #1259.

PR #1241, opened 2026-07-18
fix(AU/NZ): send Stamp on refresh + derive login valid_until from expires_in (#1778)

blka:
Fixes the two AU/NZ `KiaUvoApiAU` authentication issues confirmed live in kia_uvo #1778 (https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1778), specifically the diagnostic transcript from @stuartdenne (Kia NZ, region 7, same `KiaUvoApiAU` / `au-apigw.ccs.kia.com.au:8082` gateway).

## 1. `refresh_access_token` must send the `Stamp` header (the `4002`)

The base `ApiImplType1.refresh_access_token` (shared by AU, IN, CN) sends the `oauth2/token` refresh_token grant **without** `Stamp`, on the assumption that the endpoint family does not require it (PR #1188, based on CN's behaviour). Live evidence in #1778 disproves this for AU/NZ:

- Without `Stamp` → `400 errCode 4002 "Invalid parameters"`.
- The byte-identical grant **with** `Stamp` (inside the login flow) → `200` + `access_token`.

The fallback `try/except → self.login()` masked this, so it was benign in isolation but meant AU/NZ did a full re-login on every token expiry.

**Fix:** add a small hook on `ApiImplType1`:

```python
def _refresh_access_token_headers(self) -> dict[str, str]:
    return {}
```

The base merges it into the refresh request headers (`headers.update(self._refresh_access_token_headers())`). `KiaUvoApiAU` overrides only the hook:

```python
def _refresh_access_token_headers(self) -> dict[str, str]:
    return {"Stamp": self._get_stamp()}
```

IN and CN inherit the empty default — no behaviour change. AU still inherits the base `refresh_access_token` method (it overrides the hook, not the method), so the existing `test_au_does_not_override_refresh_access_token` still holds. The #1188 test that locked in the old no-Stamp assumption for AU (`test_au_refresh_does_not_send_stamp`) is flipped to `test_au_refresh_sends_stamp` and the module docstring corrected.

This mirrors the existing `KiaUvoApiEU._get_access_token` pattern of surfacing headers/expires_in through small per-region overrides rather than duplicating the whole method.

## 2. `login()` must derive `valid_until` from `expires_in`, not the 23h hardcode

`KiaUvoApiAU.login()` set `valid_until = now + LOGIN_TOKEN_LIFETIME` (23h), but the AU/NZ gateway returns tokens with `expires_in: 21600` (6h) and the JWT `exp` agrees. From hour 6 to 23 the library believed a dead token was alive: `check_and_refresh_token()` did not fire (valid_until in the future) and `test_token()` (base `return True`) never invalidated, so every API call 401'd until HA reauthed ~twice a day. Live-confirmed on Kia NZ by @stuartdenne.

The base `refresh_access_token` already used `expires_in` correctly; only the initial `login()` path hardcoded it.

**Fix:** `_get_access_token` now surfaces `expires_in` as a 4th return value (mirroring `KiaUvoApiEU._get_access_token`). `login()` uses it when present, falling back to `LOGIN_TOKEN_LIFETIME` (23h) when the server omits it — preserving the old behaviour for any endpoint that does not return `expires_in`.

## Out of scope

- The AU **Hyundai** `/signin` failure (`errCode 4003 "Invalid values"`, gateway returns the new Vite SPA login page) reported by the original AU Hyundai reporter is a separate, Hyundai-gateway-specific server-side change and is **not** addressed here. kia_uvo #1778 tracks it. The Kia NZ/AU gateway `/signin` works fine.
- IN / CN are unchanged (no live evidence they need Stamp on refresh).

## Tests

- `tests/test_au_refresh_access_token.py` (new): AU sends Stamp, IN/CN do not, correct grant/URL, fallback to login still works.
- `tests/test_au_login_token_lifetime.py` (new): `_get_access_token` surfaces `expires_in`; `login()` derives `valid_until` from it; falls back to `LOGIN_TOKEN_LIFETIME` when absent.
- `tests/test_type1_refresh_access_token.py`: the #1188 AU no-Stamp test flipped to assert Stamp is sent; docstring corrected.

`ruff` clean; the new tests pass; no new failures in the committed suite (the 10 `test_vehicle_snapshots` failures are pre-existing on `master`, unrelated `.ambr` staleness).

---

Live-confirmed on the Kia NZ/AU gateway by @stuartdenne (https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1778). I do not have AU/NZ credentials to test end-to-end myself; review/live-test welcome — particularly a Kia AU reporter, and a Hyundai AU reporter for the separate `/signin` issue above.

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -1321,6 +1321,14 @@ class ApiImplType1(ApiImpl):
         _check_response_for_errors(response)
         return response["msgId"]
 
+    def _refresh_access_token_headers(self) -> dict[str, str]:
+        """Extra headers for the refresh_token grant. Default: none.
+
+        AU overrides to add the Stamp header its server requires;
+        IN/CN inherit the empty default (no behavior change).
+        """
+        return {}
+
     def refresh_access_token(self, token: Token) -> Token:
         """Refresh access token using the stored refresh token.
 
@@ -1355,6 +1363,7 @@ class ApiImplType1(ApiImpl):
                     "Accept-Encoding": "gzip, deflate",
                     "User-Agent": USER_AGENT_OK_HTTP,
                 }
+                headers.update(self._refresh_access_token_headers())
                 data = "grant_type=refresh_token&refresh_token=" + token.refresh_token
                 response = self.session.post(url, data=data, headers=headers)
                 response_json = response.json()
--- a/hyundai_kia_connect_api/KiaUvoApiAU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiAU.py
@@ -955,3 +955,7 @@ class KiaUvoApiAU(ApiImplType1):
         token_type = response["token_type"]
         refresh_token = token_type + " " + response["access_token"]
         return token_type, refresh_token
+
+    def _refresh_access_token_headers(self) -> dict[str, str]:
+        """AU requires the Stamp header on the refresh_token grant."""
+        return {"Stamp": self._get_stamp()}
--- a/hyundai_kia_connect_api/KiaUvoApiAU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiAU.py
@@ -21,6 +21,7 @@ from .const import (
     DISTANCE_UNITS,
     DOMAIN,
     ENGINE_TYPES,
+    LOGIN_TOKEN_LIFETIME,
     REGION_AUSTRALIA,
     REGION_NZ,
     REGIONS,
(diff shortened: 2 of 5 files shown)


stuartdenne:
Live-tested v4.25.2 on Kia NZ (region 7, `au-apigw.ccs.kia.com.au:8082`) with the same harness as the #1778 transcript.  Both fixes confirmed working:

- `login()` now stores `valid_until` = now + 6.0h, matching `expires_in: 21600`.  The 23h dead zone is gone.
- The refresh grant goes out with Stamp.

One finding to pass on: **`refresh_access_token` still gets the 4002 and falls back to full login** - and it's no longer the Stamp, it's the token:

```
(Phase 2: refresh_access_token, Stamp sent, token seconds old)
POST .../api/v1/user/oauth2/token   grant_type=refresh_token&refresh_token=[REDACTED]
  status: 400  {"errCode": "4002", "errMsg": "Invalid parameters"}
-> "Refresh token exchange failed, falling back to full login"
```

The token it presents is what `_get_refresh_token()` stores: `refresh_token = token_type + " " + response["access_token"]` (KiaUvoApiAU.py:963) - i.e. a second 6h *access* token.  The genuine `refresh_token` from the auth-code grant is consumed as the `authorization_code` arg and discarded.  The server rejects an access token in the refresh_token slot even when it's seconds old, while the byte-identical Stamp'd grant with the *real* refresh token returns 200 (requests #5/#12 in my transcript, i.e. inside the login flow itself).

Net on v4.25.2 for NZ: the user-facing symptom is fixed - proactive fallback re-login at the true 6h boundary, no more reauth prompts - but the refresh path never actually refreshes; every 6h cycle is a full login.  Keeping the auth-code grant's `refresh_token` in the Token instead of the second access token looks like the remaining one-liner, with the caveat that I haven't tested whether that refresh token stays valid past 6h / survives reuse.

Happy to re-run the harness against any follow-up - same setup as before (largely Claude-driven, reviewed and live-tested by me).

stuartdenne:
Follow-up: raised #1242 with the fix.  One correction to my comment above: probing the grant variants showed the 4002 has two independent triggers, not just token type - the "Bearer " prefix on the stored value is fatal by itself, and so is presenting an access token (even a seconds-old one).  Variant table + live end-to-end results in the PR.

PR #1242, opened 2026-07-18
fix(AU/NZ): store the real refresh token instead of a second access token (#1778)

stuartdenne:
Follow-up to #1241 / kia_uvo #1778.  On v4.25.2 the AU/NZ refresh grant still gets the `4002` and falls back to a full login every cycle (live-confirmed on Kia NZ in [this comment](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1241#issuecomment-5012595311)).  The Stamp was necessary but not sufficient - the *token* the library presents is wrong.

## Cause

`KiaUvoApiAU.login()` minted a second access token via `_get_refresh_token()` and stored it, `"Bearer "`-prefixed, as `Token.refresh_token` - discarding the genuine refresh token returned by the authorization_code grant (it was consumed once as the `authorization_code` arg and dropped).

Probed the `oauth2/token` refresh_token grant on live Kia NZ (all variants with Stamp + Basic auth):

| refresh_token param | result |
|---|---|
| real refresh token, verbatim | **200** |
| `"Bearer "` + real refresh token | 400 `4002 Invalid parameters` |
| raw access token (seconds old) | 400 `4002 Invalid parameters` |
| real refresh token, reused | **200** |

So both deviations in the stored value are *independently* fatal: the `"Bearer "` prefix, and access-tokens-in-the-refresh-slot (regardless of liveness - this corrects my earlier #1241 comment, which attributed it to token type alone).  Also of note: the real refresh token is opaque (not a JWT) and survives reuse - no rotation.

## Fix

- `_get_access_token()` returns the response's `refresh_token` verbatim (the misleading `authorization_code` variable name is gone).
- `login()` stores it directly and no longer calls `_get_refresh_token()`; the method is removed (its only effect was minting the bogus stored value - login now makes one `oauth2/token` call instead of two).

CN/IN/EU have the same `_get_refresh_token` pattern but are untouched - no live evidence from those regions, and EU overrides the refresh flow anyway.

(Noticed in passing, untouched here: the base fallback `self.login(token.username, token.password, token.pin)` in `ApiImplType1.refresh_access_token` passes `pin` into the `otp_handler` positional, so the fallback-minted Token has `pin=None` - worth a separate look.)

**Migration:** existing stored tokens still hold the old `"Bearer <access token>"` value; the first refresh attempt fails as it does today and the existing fallback does a full login, after which the real refresh token is stored.  Self-healing, no action needed.

## Tests

- `tests/test_au_login_refresh_token.py` (new): `_get_access_token` surfaces the refresh token verbatim; `login()` stores it unmodified; `login()` makes no second token request.
- `tests/test_au_login_token_lifetime.py`: login-chain stub updated (no `_get_refresh_token` any more).
- Full suite: 389 passed; the 10 `test_vehicle_snapshots` failures are pre-existing on clean `master` (re-verified today).  `ruff check` + `format` clean.

## Live end-to-end on Kia NZ (this branch)

```
login:                        signin -> 200, oauth2/token -> 200   (refresh_token: raw opaque, 48 chars; valid_until +6.0h)
refresh_access_token:         oauth2/token -> 200, NO /signin      (same refresh_token retained)
check_and_refresh_token
  (valid_until forced past):  oauth2/token -> 200, NO /signin
```

First time the AU/NZ refresh path actually refreshes.  One caveat I could not test quickly: whether the refresh token stays valid well past the 6h access-token lifetime - it is opaque, so no `exp` to read.  My HA install runs this branch's behaviour from tonight (backported into the custom component), so I'll have multi-cycle soak data within a day or two if you want it before merging.

Disclosure, same as my #1778 comment: diagnosis, patch and tests are largely Claude (AI) work - reviewed, run and live-tested against Kia NZ by me.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiAU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiAU.py
@@ -110,10 +110,9 @@ class KiaUvoApiAU(ApiImplType1):
         if authorization_code is None:
             raise AuthenticationError("Login Failed")
 
-        _, access_token, authorization_code, expires_in = self._get_access_token(
+        _, access_token, refresh_token, expires_in = self._get_access_token(
             authorization_code, stamp
         )
-        _, refresh_token = self._get_refresh_token(authorization_code, stamp)
         if expires_in is not None:
             valid_until = dt.datetime.now(dt.timezone.utc) + dt.timedelta(
                 seconds=int(expires_in)
@@ -939,29 +938,12 @@ class KiaUvoApiAU(ApiImplType1):
 
         token_type = response["token_type"]
         access_token = token_type + " " + response["access_token"]
-        authorization_code = response["refresh_token"]
+        # Keep the refresh token verbatim: the oauth2/token refresh_token grant
+        # rejects "Bearer "-prefixed values and access tokens in the
+        # refresh_token slot with 4002 (kia_uvo #1778).
+        refresh_token = response["refresh_token"]
         expires_in = response.get("expires_in")
-        return token_type, access_token, authorization_code, expires_in
-
-    def _get_refresh_token(self, authorization_code, stamp):
-        # Get Refresh Token #
-        url = self.USER_API_URL + "oauth2/token"
-        headers = {
-            "Authorization": self.BASIC_AUTHORIZATION,
-            "Stamp": stamp,
-            "Content-type": "application/x-www-form-urlencoded",
-            "Host": self.BASE_URL,
-            "Connection": "close",
-            "Accept-Encoding": "gzip, deflate",
-            "User-Agent": USER_AGENT_OK_HTTP,
-        }
-
-        data = "grant_type=refresh_token&refresh_token=" + authorization_code
-        response = self.session.post(url, data=data, headers=headers)
-        response = response.json()
-        token_type = response["token_type"]
-        refresh_token = token_type + " " + response["access_token"]
-        return token_type, refresh_token
+        return token_type, access_token, refresh_token, expires_in
 
     def _refresh_access_token_headers(self) -> dict[str, str]:
         """AU requires the Stamp header on the refresh_token grant."""
--- a/tests/test_au_login_refresh_token.py
+++ b/tests/test_au_login_refresh_token.py
@@ -1,0 +1,64 @@
+"""Tests for KiaUvoApiAU.login() refresh-token storage (kia_uvo #1778, NZ finding).
+
+login() previously minted a second access token via a refresh_token grant and
+stored THAT, "Bearer "-prefixed, as Token.refresh_token — discarding the real
+refresh token returned by the authorization_code grant. The AU/NZ gateway
+rejects both deviations independently with 4002 "Invalid parameters"
+("Bearer "-prefixed values, and access tokens in the refresh_token slot), so
+refresh_access_token could never succeed and always fell back to a full login.
+login() must store the authorization_code grant's refresh_token verbatim.
+"""
+
+from unittest.mock import MagicMock
+
+from hyundai_kia_connect_api.KiaUvoApiAU import KiaUvoApiAU
+
+
+def _au_api_with_login_chain() -> KiaUvoApiAU:
+    """AU api with the full login chain stubbed at the helper level."""
+    api = KiaUvoApiAU(region=5, brand=2, language="en")
+    api._get_stamp = lambda: "S"  # type: ignore[assignment]
+    api._get_device_id = lambda stamp: "dev"  # type: ignore[assignment]
+    api._get_cookies = lambda: {}  # type: ignore[assignment]
+    api._get_authorization_code_with_redirect_url = (  # type: ignore[assignment]
+        lambda username, password, cookies: "authcode"
+    )
+    api._get_access_token = (  # type: ignore[assignment]
+        lambda authorization_code, stamp: ("Bearer", "Bearer atk", "raw-rtk", 21600)
+    )
+    return api
+
+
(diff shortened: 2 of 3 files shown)


cdnninja:
I haven't had a chance to compare but how similar is this to the eu code?  If same approach maybe we move it up to the api impl file?

stuartdenne:
@cdnninja last time I looked the implementations were very different.  Probably the best approach would be to point a couple of agents at it - one with a EU account, and mine would cover AU/NZ (which are essentially identical other than app IDs and the keys).  But the rollout cadence has also been historically very different as well, so I think it's probably safer to keep them independent.  Can have Claude take a look at it all to confirm this if you like?

blka:
Reviewed this against the other Type1 regions and EU to answer @cdnninja's question. Short version: the fix is correct and it brings AU in line with what EU already does; the same bug still exists in IN and CN and is worth a follow-up.

## Correctness — confirmed

The bug is long-standing (not introduced by #1241) and matches the live probe: `login()` exchanged the real refresh token (from the `authorization_code` grant) for a *second* access token via `_get_refresh_token()` and stored `"Bearer " + <that second access token>` as `Token.refresh_token`. Both deviations are independently fatal on the AU/NZ gateway — the `"Bearer "` prefix and an access-token in the refresh_token slot both return `4002`. The real refresh token was consumed once and dropped, so `refresh_access_token` could never succeed regardless of the Stamp header.

#1242 stores `response["refresh_token"]` verbatim and removes the second-mint `_get_refresh_token()` (its only caller was `login()`). Correct.

## @cdnninja's question — EU already uses this approach

EU does exactly what #1242 does for AU: it stores the real refresh token verbatim and does not mint a second access token. `KiaUvoApiEU._get_access_token` (the `refresh_token`-grant variant) returns the input refresh token as the 3rd element with the comment *"Genesis doesn't issue a new refresh_token in this response, so we keep the old one."* So #1242 aligns AU with EU — same principle.

Hoisting `login()` itself up to `ApiImpl` is not feasible: EU's login has two structurally different paths (a 48-char refresh-token-as-password path matched by `^[A-Z0-9]{48}$`, and a plaintext path via headless `_login_with_password`), while AU/IN/CN use CCSP `/signin` + the `authorization_code` grant. The shared piece is the *principle* — "store the auth-code grant's `refresh_token` verbatim; don't exchange it for a second access token" — not the login skeleton. Keeping the regions independent (as @stuartdenne suggests) is right for this PR.

## The same bug is still present in IN and CN

Worth flagging now so it's not rediscovered per-region. The pattern is identical to pre-#1242 AU:

**IN** — `KiaUvoApiIN.login()`:
```
_, access_token, authorization_code = self._get_access_token(stamp, authorization_code)
_, refresh_token = self._get_refresh_token(stamp, authorization_code)
```
and `KiaUvoApiIN._get_refresh_token()` returns `token_type + " " + response["access_token"]` — i.e. `"Bearer " + <second access token>`. IN's `_get_access_token` does extract `response["refresh_token"]` but then hands it to `_get_refresh_token`, which consumes it. Same discard-and-mint-a-second-access-token bug. IN also still hardcodes `valid_until = now + 23h` (the AU equivalent was fixed in #1241 but IN was not touched), and IN uses the base `refresh_access_token` (no `Stamp`), so IN's refresh would `4002` for both reasons.

**CN** — `KiaUvoApiCN.login()`:
```
_, access_token, authorization_code = self._get_access_token(authorization_code)
_, refresh_token = self._get_refresh_token(authorization_code)
```
`KiaUvoApiCN._get_refresh_token()` likewise returns `token_type + " " + response["access_token"]`. Same bug. CN additionally sets `valid_until = now + LOGIN_TOKEN_LIFETIME` without extracting `expires_in` from either grant.

IN/CN haven't surfaced as user reports because the refresh silently falls back to a full login every cycle (as AU did pre-#1241) and their full login still works — there's no Hyundai-AU-style `/signin` breakage on those gateways.

### Suggested follow-up

The IN/CN fix is the same shape as #1242: drop the `_get_refresh_token()` call in `login()`, store `response["refresh_token"]` verbatim from `_get_access_token()`. I'd want live credentials to confirm before merging a token-handling change there, but the code-pattern is identical to the AU/NZ case @stuartdenne just live-probed, so confidence is high. IN should also pick up the `valid_until`-from-`expires_in` treatment (and CN should start extracting `expires_in`). Happy to prepare either as a follow-up PR if useful, or leave it to whoever has IN/CN creds.

## Tests

The three new tests are good — especially `test_au_login_makes_no_second_token_request`, which guards against the second-mint coming back. One tiny note: the `test_au_login_token_lifetime.py` I added in #1241 stubs `_get_refresh_token` in its login-chain mock; after #1242 that stub is dead (login no longer calls it). Harmless, but it can be dropped in a follow-up cleanup if anyone touches that file.

All looks good to me — this is the missing half of the #1778 AU/NZ fix.

stuartdenne:
Can confirm the refresh grant is functioning as expected after a week of testing.

- Zero reauth prompts in six days (previously ~twice a day); data collection continuous all week.
- Stored token is the expected steady state: opaque 48-char refresh token, access tokens minting on the ~6h `expires_in` cadence.
- The currently-stored refresh token exchanged successfully today: Stamp'd `grant_type=refresh_token` -> 200 + `expires_in: 21600` - run from a different machine and IP than the HA instance, so this gateway does not IP-bind tokens the way CA does (errCode 7606).
- No refresh-token rotation observed in six days; reuse keeps working, consistent with the 2026-07-19 probes.
- Caveat: my HA logger was at `default: error` all week, so no per-cycle refresh-vs-fallback counts exist for the soak window (logging enabled now; happy to follow up with counts if useful).

PR #1251, opened 2026-07-24
feat: CCS2 scheduled-charging control + EV5-appMode flat endpoints (#1764)

blka:
## Summary

Closes #1764 — adds scheduled-charging control (on/off, off-peak window, time-vs-target priority) for CCS2 vehicles, including the ccNC/EV5-appMode EVs (EV6/EV9) where the existing write path was a no-op.

Three pieces:

1. **Read** — parse `Green.Reservation` from the CCS2 carstatus (`OffPeakTime.{StartHour,StartMin,EndHour,EndMin,Mode}`, `Departure.Schedule{1,2}`, departure `Climate`). Exposes `ev_off_peak_start_time`, `ev_off_peak_end_time`, `ev_schedule_charge_enabled`, `ev_off_peak_charge_only_enabled`, `ev_first_departure_*`. Mode `0`=off / `2`=target-priority / `3`=time-priority.

2. **Writer fix (combined path)** — `offPeakPowerFlag` was inverted: `2 if off_peak_charge_only_enabled else 1` sent target-priority when time-priority was requested. The reader maps flag `1`→time / `2`→target, so the writer now sends `1 if off_peak_charge_only_enabled else 2`. Affects the combined `/ccs2/reservation/chargehvac` path (CCSP EV, ccs2 PHEV/HEV).

3. **EV5-appMode flat endpoints (new)** — ccNC/EV5-appMode EVs (EV6/EV9) do not apply the combined `/ccs2/reservation/chargehvac` payload (server returns `resCode 0000` but nothing changes). They use two flat endpoints:
   - `POST /ccs2/reservation/charge` — flat body `{reservFlag, offpeakPowerFlag, reservStartTime, reservEndTime}`
   - `POST /ccs2/reservation/hvac` — departure climate `{reservedHVACInfo1, reservedHVACInfo2}`

   Dispatch by `ccu_ccs2_protocol_support` (ccNC head unit) + `engine_type == EV`, both already populated by `get_vehicles`. Combined path unchanged for everyone else. `offpeakPowerFlag` is lowercase on the flat body (not the nested `offPeakPowerInfo.offPeakPowerFlag`); `reservHVACSet` has no `heating1` (unlike `reservFatcSet`).

Live-confirmed on a Kia EV6 EU (ccs2/ccNC): the flat `/charge` endpoint flips `OffPeakTime.Mode` (3↔2) and shifts the off-peak window; the combined `chargehvac` endpoint is a no-op for this vehicle.

## Pre-merge checks (self-audit)

1. **HA principles**: dispatch reads `ccu_ccs2_protocol_support` + `engine_type` (already on Vehicle); no unknown-masked-as-current, no wake on a cheap path. `engine_type` None → falls back to combined (no regression).
2. **Pattern conformance**: `_schedule_ev5_flat` lives in `ApiImplType1` next to `schedule_charging_and_climate`, decorated `@_retry_on_device_id_error`, uses `_get_control_headers`/`_check_response_for_errors`/returns `msgId` — mirrors the combined path. Dispatch is a branch inside the entry method.
3. **No bloat**: one-line docstring; single int form for `offpeakPowerFlag`; no dual canonical forms.
4. **Structural hooks**: combined path structure unchanged; typed exceptions untouched.
5. **Right layer**: write logic in `ApiImplType1`.
6. **Type safety**: full type hints; `offpeakPowerFlag` derived from a bool option (not an API enum), times from `strftime`.
7. **Surgical scope**: `ccs2 + EV` → flat; CCSP EV, ccs2 PHEV/HEV, non-ccs2 → combined (unchanged).
8. **No dead code**: no speculative changes; EV5 flat is the confirmed fix.

## Test plan

- [x] `pytest` — 450 passed, 13 snapshots
- [x] `ruff check` + `ruff format` — clean
- [x] `pre-commit run --all-files` — ruff/codespell/pyupgrade/prettier pass (mypy manual stage)
- [x] New `tests/test_schedule_ev5_flat.py` (12 tests): dispatch (ccs2+EV→flat, CCSP/PHEV/HEV→combined), flat charge payload (lowercase `offpeakPowerFlag`, flag mapping, `reservFlag`, window time/timeSection), hvac payload (no `heating1`, `reservHVACflag`), return = charge msgId
- [x] CCS2 read-parser tests — 74 passed
- [x] Snapshot refresh is additive (96 lines added, 0 removed) — captures tier1/tier2 fields the parser already produces

## Caveat / follow-up

The dispatch `ccs2 + EV` routes all ccNC EVs to the flat endpoints. Live-confirmed on EV6 (ccs2/ccNC/EV). Other ccNC EVs (e.g. IONIQ 5, EV9) are expected to behave the same (ccNC head unit ⇒ flat endpoints), but a per-model live check on additional ccNC EVs would be a useful follow-up. If any ccNC EV turns out to need the combined path, the dispatch can be narrowed.

The HA integration side (kia_uvo) — Time/Select/Switch entities and coordinator helpers that call `schedule_charging_and_climate` — is a separate PR in the kia_uvo repo, gated on this library release + manifest bump.


hyundai_kia_connect_api/ApiImplType1.py
```diff
@@ -737,11 +737,65 @@ def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None
...
+        off_peak = get_child_value(state, "Green.Reservation.OffPeakTime") or {}
+        try:
+            vehicle.ev_off_peak_start_time = dt.time(
+                int(off_peak.get("StartHour") or 0), int(off_peak.get("StartMin") or 0)
+            )
+            vehicle.ev_off_peak_end_time = dt.time(
+                int(off_peak.get("EndHour") or 0), int(off_peak.get("EndMin") or 0)
+            )
+        except (TypeError, ValueError):
+            _LOGGER.warning("%s - CCS2 OffPeakTime malformed: %s", DOMAIN, off_peak)
```
cdnninja: Won't this warn on all none EVs?

cdnninja: Maybe instead of try catch right out of the gate set to if off_peak and allow off peak to be none like our normal.

hyundai_kia_connect_api/ApiImplType1.py
```diff
@@ -737,11 +737,65 @@ def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None
...
+        elif mode == 3:
+            vehicle.ev_schedule_charge_enabled = True
+            vehicle.ev_off_peak_charge_only_enabled = True
+        elif mode is not None:
+            _LOGGER.warning("%s - unknown CCS2 OffPeakTime.Mode: %s", DOMAIN, mode)
+        # mode is None -> no OffPeakTime block (HEV/unconfigured) -> attrs stay None
+
+        # Departure schedules
+        schedule1 = (
+            get_child_value(state, "Green.Reservation.Departure.Schedule1") or {}
```
cdnninja: Ideally want to keep unused values as None.  Easier to be consistent on the kia_uvo side.

hyundai_kia_connect_api/ApiImplType1.py
```diff
@@ -737,11 +737,65 @@ def _update_vehicle_properties_ccs2(self, vehicle: Vehicle, state: dict) -> None
...
+        )
+        try:
+            vehicle.ev_first_departure_time = dt.time(
+                int(schedule1.get("Hour") or 0), int(schedule1.get("Min") or 0)
+            )
+            vehicle.ev_second_departure_time = dt.time(
+                int(schedule2.get("Hour") or 0), int(schedule2.get("Min") or 0)
+            )
+        except (TypeError, ValueError):
+            _LOGGER.warning("%s - CCS2 departure time malformed", DOMAIN)
```
cdnninja: Same as above.

hyundai_kia_connect_api/ApiImplType1.py
```diff
@@ -1104,7 +1164,7 @@ def set_default_departure_options(
                         "time": options.off_peak_start_time.strftime("%I%M"),
                     },
                 },
-                "offPeakPowerFlag": 2 if options.off_peak_charge_only_enabled else 1,
+                "offPeakPowerFlag": 1 if options.off_peak_charge_only_enabled else 2,
```
cdnninja, line 1175: I thought this was switched due to reports of it being wrong a few months back? Not sure though.

hyundai_kia_connect_api/ApiImplType1.py
blka: Fixed in 9948c86 — applied your `if off_peak:` suggestion. When the `Green.Reservation.OffPeakTime` block is absent (HEV / unconfigured EV like the EV9 fixtures), all off-peak attrs stay `None` and no warning is logged. The `try/except` only wraps the `dt.time(...)` construction inside the truthy-block branch, so a malformed-but-present block still warns instead of crashing. Snapshot for the two EV9 fixtures updated: `ev_off_peak_start/end_time` `00:00:00` → `None`.

hyundai_kia_connect_api/ApiImplType1.py
blka: Fixed in the same 9948c86 — the None-when-absent rule is now applied to the departure schedules too. Guard is `if schedule1 and "Hour" in schedule1:` because the EV9 fixtures ship `Schedule1 = {"Enable": False}` (truthy block, no `Hour/Min`): that's a stub, not a disabled-but-remembered schedule, so time/days stay `None`. `ev_*_departure_enabled` is still derived from `Enable` earlier in the function and is unaffected. EV6 (`Enable: 0` + `Hour: 2`) still parses to `02:40` — the remembered window is preserved when a schedule is merely toggled off. EV9 snapshots: `ev_first/second_departure_time` `00:00:00` → `None`.

hyundai_kia_connect_api/ApiImplType1.py
blka: Symmetric to Schedule1 — `if schedule2 and "Hour" in schedule2:`. Same 9948c86.

hyundai_kia_connect_api/ApiImplType1.py
blka: Verified the history — the flag direction was never corrected after being introduced. `offPeakPowerFlag` landed in #767 (5558347, 2025-04-09) as `2 if off_peak_charge_only_enabled else 1` and `git log -S offPeakPowerFlag` shows no later change to that line. Three independent sources confirm the pre-#1251 value was inverted:

1. The decompiled enum: write flag `1 = time priority`, `2 = target priority` (`off_peak_charge_only_enabled=True` is time-priority → must be `1`).
2. Live confirmation on a CCS2 EV6 (Mode 3 = time): sending `offpeakPowerFlag: 2` flipped the app to Mode 2 (target), proving `2 = target`.
3. The EV5 flat path this PR adds uses `1 if off_peak_charge_only_enabled else 2` and live-confirmed (priority + window) on the same EV6.

So the old `2 if True else 1` was sending time-priority requests as `2` (target) — silent wrong behaviour. #1251 flips it to `1 if True else 2`, consistent across both write paths. The direction is now locked by `test_schedule_charging_off_peak_flag_not_inverted` and `test_schedule_charging_off_peak_flag_target_priority` in `test_ccs2_vehicle_properties.py`, so it can't silently regress again.

cdnninja:
Good from my standpoint.  Just had merge conflicts.

PR #1252, opened 2026-07-24
fix: retry Brazilian vehicle status with CCS2 header

eduardobittencourt:
## What changed

- Retry Brazilian Hyundai legacy vehicle status requests once with
  `ccuCCS2ProtocolSupport: 1` when the first response is HTTP 503 with API
  code `5031`.
- Keep the existing request path and response parser unchanged.

## Root cause

The Brazilian API currently returns `ccuCCS2ProtocolSupport: 0` from
`GET /api/v1/spa/vehicles` for at least some vehicles, so the client sends
that value when requesting `/status/latest`.

The status endpoint has started rejecting that request with:

```json
{
  "retCode": "F",
  "resCode": "5031",
  "resMsg": "Unavailable remote control - Service Temporary Unavailable"
}
```

Repeating the same request with `ccuCCS2ProtocolSupport: 1` returns HTTP 200
and the normal legacy `resMsg` payload. This indicates the vehicle-list
capability flag can lag behind the status API's required protocol header.

## Impact

Brazilian Hyundai accounts affected by this server-side inconsistency can
load and update their vehicles again. The retry is limited to legacy vehicles
and the specific `5031` response, so other server failures are not masked.

## Validation

- Reproduced against the live Brazilian API: initial request returned HTTP
  503 / `5031`; retry with the header set to `1` returned HTTP 200.
- Tested the patch in Home Assistant with a Brazilian Hyundai account; the
  integration and its entity platforms loaded successfully.
- `pytest -q`: 400 passed, 10 skipped.
- Pre-commit checks passed for the changed files.


eduardobittencourt:
Closing this in favor of #1240, which addresses the same Brazilian HTTP 503 / resCode 5031 issue with a more complete solution. It switches cached status reads to the CCS2 endpoint, handles the asynchronous CCS2 force-refresh flow, reuses the existing CCS2 parser, and includes broader validation.\n\nFor additional context, while investigating this I also confirmed against the live BR API that retrying the legacy `/status/latest` request with `ccuCCS2ProtocolSupport: 1` returns HTTP 200, despite the vehicle list reporting `0`. That narrower workaround is what this PR implemented, but #1240 better matches the API behavior and preserves the richer CCS2 payload.

PR #1256, opened 2026-07-28
chore: adopt ruff 0.16 default rules (selective ignore + mechanical fixes)

blka:
## Summary

`#1255` (pre-commit autoupdate) bumped ruff to v0.16.0, which expanded the default lint rule set from **59 → 413 rules** (+354). With no `[tool.ruff.lint]` config, every PR rebased onto master inherited a `pre-commit.ci` FAIL from pre-existing code the feature PRs did not touch (#1255 itself merged with pre-commit.ci red; #1253/#1254 passed only because they were checked before #1255 merged).

This PR adopts the new defaults with a documented `ignore` list and fixes the 34 mechanical violations. No behavior change — every transformation is behavior-equivalent or strictly more informative.

## What changed

**Config — `pyproject.toml` (new `[tool.ruff.lint]` block):**
- `N999` — permanent. PascalCase module names (`ApiImpl`, `KiaUvoApiEU`, …) are the package's public API; `kia_uvo` imports them directly. Renaming is a breaking major-version change.
- `BLE001` — permanent. `except Exception` is the library's resilience pattern for flaky vendor HTTP; narrowing would let new exception types crash the HA integration.
- `DTZ001/004/005/006/007/901` — deferred. Naive datetimes in token-expiry + region timestamp handling. Tz-aware migration is all-or-nothing (mixed aware/naive raises `TypeError` on comparison); needs its own PR with a timestamp-comparison audit.

**Code fixes (34 sites, ~15 package + ~10 test files, 1–2 lines each):**
- `SIM102/103/117/118` — collapsible-if, needless-bool, nested-with, `in dict.keys()`.
- `C408` — `dict()` → literal. `RUF015` — `next(iter(...))` over slice. `PERF102` — `.values()`.
- `RUF012` — 6 `temperature_range*` list defaults → `tuple(...)` (instances reassign to `range(...)`; `ClassVar` would forbid that). `_action_service_types` → `ClassVar[dict[str, str]]` (shared class-level cache, mutated per-instance via item assignment).
- `RUF013` — `x: str = None` → `x: str | None` (3 sites incl. `scripts/`).
- `UP031` — `%064x` → f-string (3 region `_get_device_id`); vestigial `# pylint: disable` removed.
- `UP047` — `TypeVar("T", bound=IntEnum)` → PEP 695 `def to_int_enum[T: IntEnum](...)` (bound preserved).
- `G201` — `.error(..., exc_info=True)` → `.exception(...)` (1 site).
- `TRY201` — `raise e` → `raise` (preserves traceback).
- `EXE001` — shebangs removed from `docs/conf.py` + `setup.py` (vestigial); `scripts/ca_api_diagnostic.py` made executable (shebang preserved).
- `I001` — import sort.

## Verification

- `uvx ruff@0.16.0 check hyundai_kia_connect_api/ tests/ scripts/ca_api_diagnostic.py docs/conf.py setup.py` → `All checks passed!`
- `uvx ruff@0.16.0 check --fix …` → no-op (no remaining fixes).
- `pytest tests/ --ignore=tests/integration` → 426 passed, 11 snapshots.
- `pre-commit run --all-files` → green (ruff, ruff-format, codespell, prettier, mypy).
- `#1251` untouched.

## Two minor decisions — would appreciate your call, @cdnninja

1. **G201 `_LOGGER.exception` form (`KiaUvoApiEU.py:1037`)** — the original was `_LOGGER.error(f"... {e}", exc_info=True)`. G201 wants `.exception()`; ruff's `TRY401` then fires on `except Exception as e:` because the binding is "avoidable". I restored `{e}` + `# noqa: TRY401` to keep the exception text in the log **message line** (the original intent — full logs still get the traceback either way). Alternative: drop `{e}` for a clean no-noqa form (exception text lives only in the traceback). I went with preserving message-line info; flag if you prefer the no-noqa form.

2. **DTZ* deferral** — 28 naive-datetime sites (`datetime.now()`, `strptime()` without tz, `datetime.min`) across `Token.py` expiry + all region timestamp handling. Made tz-aware is all-or-nothing (mixed aware/naive raises `TypeError` on the token-expiry comparisons), so I deferred it to a separate PR with a timestamp audit rather than fold a semantic migration into this chore. OK to defer, or would you rather it ride here?

(Out of scope, FYI: 3 `_LOGGER.warning(..., exc_info=True)` sites exist in `HyundaiBlueLinkApiUSA.py`/`KiaUvoApiUSA.py` — G201 only fires on `.error`, so `.warning`-level sites are intentionally untouched.)

## Follow-up

Separate tz-aware datetime migration PR for `DTZ*` (28 sites).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 #
 # hyundai_kia_connect_api documentation build configuration file, created by
 # sphinx-quickstart on Fri Jun  9 13:47:02 2017.
--- a/hyundai_kia_connect_api/ApiImpl.py
+++ b/hyundai_kia_connect_api/ApiImpl.py
@@ -236,7 +236,7 @@ class ApiImpl:
         vehicle: Vehicle,
         use_email: bool,
         provider: int = 1,
-        API_KEY: str = None,
+        API_KEY: str | None = None,
     ) -> None:
         if vehicle.location_latitude and vehicle.location_longitude:
             if (
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -173,23 +173,22 @@ class ApiImplType1(ApiImpl):
             dte = fuel_system.get("DTE", {})
             ice_value = dte.get("ICE")
 
-            if ice_value is not None:
-                # Check if the value is a number and unreasonably large
-                # A reasonable ICE range should be < 1,000,000 km/miles
-                # Values larger than this are likely corrupted API data
-                if isinstance(ice_value, (int, float)):
-                    # Also check if it exceeds JavaScript's Number.MAX_SAFE_INTEGER
-                    # (2^53 - 1 = 9,007,199,254,740,991) to prevent JSON issues
-                    max_safe_integer = 9007199254740991
-                    max_reasonable_range = 1000000
+            # Check if the value is a number and unreasonably large
+            # A reasonable ICE range should be < 1,000,000 km/miles
+            # Values larger than this are likely corrupted API data
+            if ice_value is not None and isinstance(ice_value, (int, float)):
+                # Also check if it exceeds JavaScript's Number.MAX_SAFE_INTEGER
+                # (2^53 - 1 = 9,007,199,254,740,991) to prevent JSON issues
+                max_safe_integer = 9007199254740991
+                max_reasonable_range = 1000000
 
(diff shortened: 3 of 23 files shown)


blka:
Follow-up tracked in #1257 (migrate naive datetimes to tz-aware — the 28 `DTZ*` sites this PR ignores). The `pyproject.toml` ignore comment ("Deferred to a separate tz-aware migration PR.") now points there in spirit; happy to add `#1257` to that comment if you want it explicit.

blka:
What about 1 / 2?

PR #1259, opened 2026-07-29
fix(BR): surface 4xx signin errors and read status from CCS2 endpoint

blka:
## What changed

**Login (4xx error handling):** `_raise_auth_error` reads the response body before raising, so 4xx from `/user/signin` (and `/oauth2/authorize`, `/oauth2/token`) surfaces a readable `AuthenticationError` instead of a raw `HTTPError`. The BR backend returns HTTP 400 for password-expired (step 5), `errCode`/`errMsg`, or non-JSON WAF responses. This extends #1239 (which only handled HTTP 200 + `{step:N}`) to 4xx, across all three BR auth call-sites (`_get_cookies`, `_get_authorization_code`, `_get_auth_response`).

**Status (CCS2 endpoint):** BR extends `ApiImplType1` and reads cached status from `/ccs2/carstatus/latest`, reusing `_update_vehicle_properties_ccs2`. BR vehicles report `ccuCCS2ProtocolSupport: 0` but the cached status is served in CCS2 format (including location); the legacy `/status/latest` endpoint is the remote-control command result on BR and always returns 503 / resCode 5031. Force refresh uses the #1250 wake+sleep+read pattern with a `lastUpdateTime` guard: if the snapshot did not advance after the wake, raise `APIError` (→ `UpdateFailed` → entities unavailable) rather than apply stale data. `supports_valet_mode = False` (BR has no valet). Drops the legacy flat status parser (~190 lines).

## Supersedes

Supersedes #1240 — credit to @paulogabrielfs for the CCS2-endpoint investigation and the anonymized Creta fixture. Force-refresh is rebuilt on the merged #1250 wake+sleep+read pattern instead of #1240's poll-until-timestamp loop.

## Related issues

**hyundai_kia_connect_api:** #1239 (the 200+step fix this extends to 4xx), #1240 (superseded), #1250 (force-refresh pattern), #1256 (ruff 0.16 base this branches from)
**kia_uvo:** Hyundai-Kia-Connect/kia_uvo#1792 (BR connect failure — both the 503 status and 400 login cases), Hyundai-Kia-Connect/kia_uvo#1515

## Test plan

- [x] `pytest tests/test_br_vehicle_properties.py tests/test_br_login.py -q` — 22 passed
- [x] `pytest tests/ -q` (excluding local-only browser/integration tests) — 448 passed, 11 snapshots
- [x] `ruff check` / `ruff format` / `pre-commit run` — all pass
- [ ] Live validation welcome: a BR account hitting the 400 on `/user/signin` should now see the actionable message (e.g. "password has expired"); cached status should load without the 503/5031.

### DIFF
--- a/tests/test_br_login.py
+++ b/tests/test_br_login.py
@@ -1,0 +1,127 @@
+"""Tests for Brazilian Hyundai BlueLink auth error handling.
+
+Regression for kia_uvo #1792 / #1515: the BR signin endpoint returns HTTP 400
+(with a JSON body describing the failure) for password-expired / blocked /
+verification-required accounts. Previously ``raise_for_status()`` raised a raw
+``HTTPError`` before the body was parsed, so users saw "400 Client Error: Bad
+Request for url: .../user/signin" with no actionable reason.
+
+These tests cover ``_raise_auth_error`` across the three BR auth call-sites
+(``_get_cookies``, ``_get_authorization_code``, ``_get_auth_response``) plus the
+existing HTTP 200 + ``{step:N}`` path (#1239) as a regression guard.
+"""
+
+from unittest.mock import MagicMock
+
+import pytest
+
+from hyundai_kia_connect_api.const import BRAND_HYUNDAI, BRANDS, REGION_BRAZIL, REGIONS
+from hyundai_kia_connect_api.exceptions import AuthenticationError
+from hyundai_kia_connect_api.HyundaiBlueLinkApiBR import HyundaiBlueLinkApiBR
+
+_BR_REGION = next(k for k, v in REGIONS.items() if v == REGION_BRAZIL)
+_HYUNDAI_BRAND = next(k for k, v in BRANDS.items() if v == BRAND_HYUNDAI)
+
+
+def _resp(status_code: int, json_data: dict | None = None, text: str = "") -> MagicMock:
+    r = MagicMock()
+    r.status_code = status_code
+    if json_data is not None:
+        r.json.return_value = json_data
+    else:
+        r.json.side_effect = ValueError("not JSON")
+    r.text = text
+    return r
+
+
+@pytest.fixture
+def br_api() -> HyundaiBlueLinkApiBR:
+    return HyundaiBlueLinkApiBR(region=_BR_REGION, brand=_HYUNDAI_BRAND)
+
+
+class TestRaiseAuthError:
+    """Unit-test the helper directly."""
+
+    def test_no_op_on_success(self, br_api):
+        # 2xx must not raise.
+        br_api._raise_auth_error(_resp(200, {"redirectUrl": "x"}), "signin")
+
+    def test_400_step_surfaces_readable_reason(self, br_api):
+        with pytest.raises(AuthenticationError, match="password has expired"):
+            br_api._raise_auth_error(_resp(400, {"step": 5}), "signin")
+
+    def test_400_errcode_errmsg(self, br_api):
+        with pytest.raises(AuthenticationError, match="errCode=4003"):
+            br_api._raise_auth_error(
+                _resp(400, {"errCode": 4003, "errMsg": "Invalid credentials"}), "signin"
+            )
+
+    def test_400_non_json_falls_back_to_snippet(self, br_api):
+        # Cloudflare / WAF HTML response, no JSON body.
+        with pytest.raises(AuthenticationError, match="Response not JSON"):
+            br_api._raise_auth_error(
+                _resp(403, None, text="<html>Attention Required</html>"),
+                "cookie request",
+            )
+
+    def test_400_unknown_shape_lists_keys_only(self, br_api):
+        with pytest.raises(AuthenticationError, match="keys="):
+            br_api._raise_auth_error(_resp(400, {"foo": "bar"}), "token request")
+
+
+class TestGetAuthorizationCode:
+    """The signin call-site: 4xx must surface the body, not a raw HTTPError."""
+
+    def test_400_step5_raises_authentication_error_not_httperror(self, br_api):
+        br_api.session = MagicMock()
+        br_api.session.post.return_value = _resp(400, {"step": 5})
+        with pytest.raises(AuthenticationError, match="password has expired"):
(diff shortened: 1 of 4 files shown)


guicampos1:
Live validation from a BR account — Hyundai Creta 2026 Limited, Santa Catarina, Brazil.

**Setup:** kia_uvo 3.8.1, HA Core on Python 3.14. Installed this branch by pointing the `requirements` entry in `custom_components/kia_uvo/manifest.json` at the branch tarball instead of `hyundai_kia_connect_api==4.25.3`.

### Cached status — works

Before: every coordinator refresh failed with `503 Server Error` on `/api/v1/spa/vehicles/{id}/status/latest`, integration stuck on `Config Not Ready`. Same traceback as kia_uvo#1792, but reproduced on 3.8.1 — the issue was filed on 3.7.1, so the bug is still present on the current release.

After: cached status loads from the CCS2 endpoint, no more 503 / resCode 5031. All entities populate correctly — odometer, fuel level, door and lock state, location. Your read of the BR backend is confirmed: `ccuCCS2ProtocolSupport: 0` but the cached snapshot really is served in CCS2 format.

One benign warning at startup, from `ApiImplType1`: `Refresh token exchange failed, falling back to full login`. Single occurrence, full login succeeded right after, no impact.

### Summary

The core fix in this PR resolves the BR breakage. Cached status is the part that matters day to day and it's working. 

Happy to run further tests or provide anonymized fixtures.

paulogabrielfs:
Live-validated #1259 on my Creta (ICE, ccuCCS2ProtocolSupport: 0):

Cached status, works perfectly, loads from /ccs2/carstatus/latest, no 503, all entities populate (odometer, fuel, locks, location), and matches what @guicampos1 reported.

Force refresh (sleep(25) + raise-on-stale) works when the car is reachable: one run woke it and returned fresh data within the 25s window. 
But I hit a case worth flagging: with the car in overnight deep sleep, I triggered the wake and polled /ccs2/carstatus/latest every 3s for 46s and lastUpdateTime never advanced, the car only started reporting a couple minutes later. 
In that state #1259 raises APIError after 25s (entities go unavailable) and only succeeds on a retry.

So the fixed 25s is fine for an awake car, but on a deep-sleeping one the first manual refresh can intermittently fail. 
Two options if you want to smooth it: fall back to the last cached snapshot instead of raising, or poll a few more times past 25s before giving up. 

Not a blocker, just sharing the live data, since it lines up with the ~20–22s edge you measured in #1184. 
Happy to test any variant.

Login: the normal login path still works on the refactored _raise_auth_error.

blka:
@paulogabrielfs thanks for the live data — really useful, and it lines up with the ~20–22s edge in #1184.

The raise-on-stale is deliberate, not an oversight. Force refresh is an explicit user action expecting fresh data; if the wake doesn't produce a newer `lastUpdateTime`, applying the cached snapshot would mask unavailable as current. Per HA principles (`UpdateFailed` → entities unavailable, not stale-as-fresh), raising `APIError` so the coordinator surfaces the failure is the intended path. Your deep-sleep case (fail on first attempt, succeed on retry) is the expected behavior for an unreachable car, not a regression.

On the two options:
- **Fall back to cached:** rejected for the reason above — it's exactly what the `lastUpdateTime` guard exists to prevent.
- **Poll past 25s:** your own run shows it wouldn't have helped here — you polled 46s and `lastUpdateTime` still hadn't advanced, the car reported ~2 min later. Extending the window only blocks force-refresh longer before failing on the same deep-sleep case. The retry on the next poll already covers it.

So no change planned for this PR. Happy to revisit if more deep-sleep data shows a tighter wake→report window that a modest extension would reliably catch.

tests/test_br_login.py
```diff
@@ -0,0 +1,127 @@
+"""Tests for Brazilian Hyundai BlueLink auth error handling.
```
cdnninja, line 1: Longer term can't these just be parameters without unique tests for each region?

cdnninja:
This feels like it should be multiple PRs. Our goal should also be to slowly move things out of the BR child whenever possible as they were grouped because of very similar.  _raise_auth_error looks unique to the region but should it be?   Can it follow standards from type1?

blka:
Re: multiple PRs — fair point, the two changes are technically separable (login 4xx touches auth methods ~115-239 + `test_br_login.py`; the CCS2 status read is the `@@ -267,195 +308,60` rewrite + `test_br_vehicle_properties.py` + fixture). I'm keeping them in one PR because both fix the same BR breakage reported in Hyundai-Kia-Connect/kia_uvo#1792 — the 503 status path and the 400 login path are two symptoms of one bug, and the status fix is only reachable once login works. Two BR accounts live-validated both together in one session. Splitting would just fragment the fix for the issue this PR closes.

Re: `_raise_auth_error` / type1 standards — Type1's `_check_response_for_errors` operates on the SPA envelope (`retCode`/`resCode`/`resMsg`), which the BR OAuth2 auth endpoints don't return. They return `{step}`, `{errCode, errMsg}`, or non-JSON WAF HTML — a different schema, so there's no Type1 standard to follow here. Other regions handle auth 4xx ad hoc (AU/IN/CN: bare `AuthenticationError("Login Failed")`; CA: OTP-specific; USA: its own; EU: headless). The genuinely generic part of `_raise_auth_error` is "read the body before raising so the message is actionable" — that skeleton could be hoisted to `ApiImpl` as a follow-up, with BR supplying the key map and `_SIGNIN_STEP_MESSAGES`. I'd rather do that hoist as its own PR, not expand scope here.

tests/test_br_login.py
blka: Agreed in principle — if the body-reading helper gets hoisted to `ApiImpl` (see my reply on the issue thread), a parametrized test over `(region, schema, expected)` would make sense. Today each region's auth endpoints return a different schema (BR: `step`/`errCode`/`errMsg`; CA: OTP `error_desc`; USA: `enrolledVehicleDetails`; EU: headless OAuth), so a shared parametrization reduces to per-region fixtures anyway. These tests pin BR's `_raise_auth_error` until there's a shared helper to test.

PR #1260, opened 2026-07-30
fix(CCS2): start_climate sideRearMirrorHeating + RHD front seat swap

### LINKED ISSUE #1195: EU/CCS2 start_climate ignores heating parameter — sideRearMirrorHeating hardcoded to 1 (rear window + side mirror heating always on)
**Region and Brand of car**
EU, Hyundai Kona EV (CCS2 protocol vehicle)

**Library version**
hyundai_kia_connect_api 4.11.0 (via kia_uvo Home Assistant integration)

**Description of the bug**
On CCS2-protocol vehicles, `start_climate` always turns on the rear-window + side-mirror heating, regardless of the `heating` option passed. Sending `heating=0` ("Off") has no effect — the rear defroster and side mirror heaters still activate. This makes it impossible to run cooling (e.g. in summer) without also enabling the rear/side defrosters.

**Root cause**
In `ApiImplType1.py`, `start_climate()`, the CCS2 branch (`vehicle.ccu_ccs2_protocol_support` is `True`) builds the payload with `sideRearMirrorHeating` **hardcoded to `1`** instead of deriving it from `options.heating`:

```python
payload = {
    "command": "start",
    "ignitionDuration": options.duration,
    "strgWhlHeating": options.steering_wheel,
    "hvacTempType": 1,
    "hvacTemp": options.set_temp,
    "sideRearMirrorHeating": 1,          # <-- hardcoded, ignores options.heating
    "drvSeatLoc": "R",
    "seatClimateInfo": { ... },
    "tempUnit": "C",
    "windshieldFrontDefogState": options.defrost,
}
```

The legacy (non-CCS2) branch in the same method correctly honors the value via `"heating1": int(options.heating)`, and the USA implementation (`KiaUvoApiUSA.py`) derives the heater states from `options.heating`:

```python
"rearWindow": 1 if options.heating in [1, 2, 4] else 0,
"sideMirror": 1 if options.heating in [1, 4] else 0,
```

So `heating=0` is respected on legacy and USA paths, but ignored on the EU/CCS2 path.

**Expected behavior**
`sideRearMirrorHeating` should be derived from `options.heating`, so that `heating=0` keeps the rear window + side mirrors off. Suggested fix (mirroring the USA logic):

```python
"sideRearMirrorHeating": 1 if options.heating in [1, 2, 4] else 0,
```

**To Reproduce**
1. Use an EU CCS2 vehicle (e.g. Hyundai Kona EV).
2. Call `kia_uvo.start_climate` with `climate: true`, `heating: "0"` (and e.g. seat cooling).
3. Observe that the rear-window heater / side-mirror heater turn on anyway (visible via the car and the `back_window_heater` / `steering_wheel_heater` state sensors).

**Expected vs actual**
- Expected: with `heating: "0"`, rear window + side mirror heating stay **off**.
- Actual: they turn **on** every time climate is started.

**Affected scope**
Likely all CCS2 vehicles using `ApiImplType1` (EU/AU/IN/CN), since the hardcoded value is in the shared CCS2 payload. (Note: `drvSeatLoc` is hardcoded to `"R"` in the same payload as well.)

**Additional context**
Per the integration's `start_climate` docs, `heating` is the only parameter controlling the rear-window/side-mirror/back defrosters (`0`=Off, `2`=Rear Window Only, `3`=Steering Only, `4`=Steering + Side & Back Defroster); there is no separate parameter, so `heating=0` is the intended way to keep them off.

### LINKED ISSUE #1447: UK (RHD) vehicle; kia_uvo.start_climate vehicle heating is reversed.
UK (RHD) Ioniq 5 - when using "kia_uvo.start_climate" the heating actions seats are wrong way round.

The integration has the following entities for seats;

flseat - Front Left Seat Heat Cool Setting
frseat - Front Right Seat Heat Cool Setting

As a UK RHD vehicle, the driver sits in "frseat" 
The action triggers absolutely fine, but frseat triggers the passenger (flseat) and vice versa.

I cannot test cooling nor can I test the rear seats, as my vehicle isn't specced with these.

It's a low priority issue, as I've just changed my automations, but thought worth bringing up.

blka:
## Summary

Two correctness fixes in the CCS2 branch of `ApiImplType1.start_climate` (reached by `KiaUvoApiEU` and `KiaUvoApiAU` for CCS2 vehicles — `ccu_ccs2_protocol_support` true). IN/CN override `start_climate` with their own `/control/engine` path and are unaffected; BR/USA/CA are not `ApiImplType1` subclasses.

### Commit 1 — `sideRearMirrorHeating` derived from `options.heating`

The CCS2 branch hardcoded `sideRearMirrorHeating: 1`, so the rear-window + side-mirror heaters turned on for every climate start regardless of the `heating` option (`heating=0` still enabled them). Derive it instead, mirroring the Kia USA `rearWindow`/`sideMirror` logic: on for `1`/`2`/`4`, off for `0` and steering-only `3`.

Closes Hyundai-Kia-Connect/hyundai_kia_connect_api#1195. Relates to Hyundai-Kia-Connect/kia_uvo#1739.

### Commit 2 — swap front driver/passenger seats for RHD

`seatClimateInfo` mapped `drvSeatClimateState` from `front_left_seat` and `psgSeatClimateState` from `front_right_seat` unconditionally — an LHD assumption. For RHD vehicles (`drvSeatLoc == "R"`, driver on the right) this inverted the seats: `frseat` activated the passenger. Swap the front seats when `drvSeatLoc` is `"R"`; rear seats are physical left/right and stay unchanged. LHD behaviour is unaffected.

This is the complement to the existing `_get_drv_seat_loc` fix: that tells the car which side the driver is on (the `drvSeatLoc` field); this maps the HA physical seats (`flseat`/`frseat`) to the logical `drv`/`psg` seats consistently with that side.

Closes Hyundai-Kia-Connect/kia_uvo#1447.

## ⚠️ Verification basis for the RHD swap — please confirm

The swap assumes the car interprets `drvSeatClimateState`/`psgSeatClimateState` as **logical** driver/passenger seats remapped by `drvSeatLoc` (i.e. `drvSeatLoc="R"` + `drvSeatClimateState=X` heats the right/driver seat with `X`). This is supported by the #1447 report (UK Ioniq 5, filed *after* the `drvSeatLoc` fix was in place): with `drvSeatLoc` correct, `frseat` still activated the passenger — i.e. the car remaps `drv`/`psg` by `drvSeatLoc`, and the pre-fix `front_left→drv` mapping heated the wrong seat.

That evidence is from a single vehicle (UK Ioniq 5). AU CCS2 (also RHD, same CCSP/CCS2 protocol) is assumed to interpret the same way but has no seat-swap report. **Maintainer/reporter confirmation that this matches app/vehicle behaviour — especially for AU — would be welcome.** If the car instead treats `drvSeatClimateState` as a fixed physical seat (not remapped by `drvSeatLoc`), the swap direction would need revisiting.

The status **read** side uses physical `flSeatHeatState`/`frSeatHeatState` keys (side-agnostic), so it is correct as-is; only the **write** side carried the LHD assumption, and this PR brings write into consistency with the physical read.

## Testing

New `tests/test_ccs2_climate_payload.py` (8 tests, payload assertions against the CCS2 `start_climate` payload):
- `sideRearMirrorHeating`: off for `heating=0`/`3`, on for `1`/`2`/`4`.
- RHD: `drvSeatClimateState = front_right_seat`, `psgSeatClimateState = front_left_seat`; rear seats unchanged.
- LHD: `drvSeatClimateState = front_left_seat` (unchanged behaviour, regression guard).

`ruff` (0.16.0) + `pre-commit` clean; full suite green (436 + 11 snapshots). `mypy --strict` introduces no new errors on the touched lines.

## Related

- Hyundai-Kia-Connect/hyundai_kia_connect_api#1195 — `sideRearMirrorHeating` hardcoded (commit 1)
- Hyundai-Kia-Connect/kia_uvo#1739 — EU EV4 rear/mirror always on (commit 1, same root cause)
- Hyundai-Kia-Connect/kia_uvo#1447 — UK RHD Ioniq 5 seat reversal (commit 2)

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -1181,6 +1181,13 @@ class ApiImplType1(ApiImpl):
                 + vehicle.id
                 + "/ccs2/control/temperature"
             )
+            # For RHD vehicles the driver sits on the right, so the front
+            # physical seats map to passenger/driver respectively.
+            drv_seat_loc = self._get_drv_seat_loc(vehicle)
+            if drv_seat_loc == "R":
+                drv_seat, psg_seat = options.front_right_seat, options.front_left_seat
+            else:
+                drv_seat, psg_seat = options.front_left_seat, options.front_right_seat
             payload = {
                 "command": "start",
                 "ignitionDuration": options.duration,
@@ -1190,10 +1197,10 @@ class ApiImplType1(ApiImpl):
                 # heating 1/2/4 engage the rear-window + side-mirror heaters;
                 # 0 (off) and 3 (steering wheel only) leave them off.
                 "sideRearMirrorHeating": 1 if options.heating in (1, 2, 4) else 0,
-                "drvSeatLoc": self._get_drv_seat_loc(vehicle),
+                "drvSeatLoc": drv_seat_loc,
                 "seatClimateInfo": {
-                    "drvSeatClimateState": options.front_left_seat,
-                    "psgSeatClimateState": options.front_right_seat,
+                    "drvSeatClimateState": drv_seat,
+                    "psgSeatClimateState": psg_seat,
                     "rrSeatClimateState": options.rear_right_seat,
                     "rlSeatClimateState": options.rear_left_seat,
                 },
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -1187,7 +1187,9 @@ class ApiImplType1(ApiImpl):
                 "strgWhlHeating": options.steering_wheel,
                 "hvacTempType": 1,
                 "hvacTemp": options.set_temp,
-                "sideRearMirrorHeating": 1,
+                # heating 1/2/4 engage the rear-window + side-mirror heaters;
+                # 0 (off) and 3 (steering wheel only) leave them off.
+                "sideRearMirrorHeating": 1 if options.heating in (1, 2, 4) else 0,
                 "drvSeatLoc": self._get_drv_seat_loc(vehicle),
                 "seatClimateInfo": {
                     "drvSeatClimateState": options.front_left_seat,
--- a/tests/test_ccs2_climate_payload.py
+++ b/tests/test_ccs2_climate_payload.py
@@ -89,3 +89,59 @@ class TestSideRearMirrorHeating:
     def test_heating_eu_steering_side_back_enables(self):
         options = ClimateRequestOptions(set_temp=21, heating=1)
         assert self._heating(options) == 1
+
+
+class TestRhdSeatSwap:
+    """For RHD (drvSeatLoc == "R") the front driver/passenger seats are swapped.
+
+    The HA service fields flseat/frseat are physical (front-left/right). The
+    API fields drvSeatClimateState/psgSeatClimateState are logical
+    (driver/passenger). For RHD vehicles the driver sits on the right, so the
+    physical front-right seat is the driver. Regression for
+    Hyundai-Kia-Connect/kia_uvo#1447 (UK Ioniq 5: frseat activated passenger).
+    """
+
+    def setup_method(self):
+        self.api = _make_api()
+
+    def _seat_info(self, vehicle, options):
+        return _post_payload(self.api, vehicle, options)["seatClimateInfo"]
+
+    def test_lhd_driver_is_front_left(self):
+        """LHD: driver sits on the left -> drvSeatClimateState = front_left."""
+        from hyundai_kia_connect_api.const import DISTANCE_UNITS
+
+        vehicle = _make_vehicle(ccs2=True)
+        vehicle._odometer_unit = DISTANCE_UNITS[1]  # km -> LHD
+        options = ClimateRequestOptions(
+            set_temp=21, front_left_seat=1, front_right_seat=2
+        )
+        info = self._seat_info(vehicle, options)
+        assert info["drvSeatClimateState"] == 1  # front_left
+        assert info["psgSeatClimateState"] == 2  # front_right
+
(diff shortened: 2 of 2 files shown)


blka:
Review update — verified against the official app behaviour for EU CCS2 vehicles:

**Front-seat RHD swap (commit 2).** The official app sends `seatClimateInfo` with fixed key names (`drvSeatClimateState` / `psgSeatClimateState`) and, for EU/RU vehicles, flips which physical seat feeds each key based on `drvSeatLoc`:
- `drvSeatLoc == "R"` (RHD): `drvSeatClimateState` ← front-right seat, `psgSeatClimateState` ← front-left seat
- otherwise (LHD): `drvSeatClimateState` ← front-left seat, `psgSeatClimateState` ← front-right seat
- rear seats (`rlSeatClimateState` / `rrSeatClimateState` / 3rd-row) are never swapped

This matches this PR exactly. For AU (RHD, non-EU app branch), the app uses a pre-assigned driver/passenger seat mapping; because the library exposes physical `front_left_seat` / `front_right_seat` inputs, the swap is still required to land `drvSeatClimateState = driver`, so applying it for AU RHD is correct (converges to the same result).

**`sideRearMirrorHeating` (commit 1).** The library's `heating` selector is the exposed control surface — `0`=Off, `1`=Steering+Side+Back (EU), `2`=Rear Window, `3`=Steering Only, `4`=Steering+Side+Back (non-EU). Deriving `sideRearMirrorHeating = 1 if heating in (1, 2, 4) else 0` matches the intent of each mode: the modes that include rear/side-mirror heating (1/2/4) turn it on; Off (0) and Steering-Only (3) leave it off. This resolves #1195 (the field was hardcoded `1`, so even `heating=0` enabled the mirror heaters).

Issue #1273, opened 2026-08-12
And new log in issues again today (at some point after 2026-08-10,23:08:05 UTC)


Jesus-M:
I noticed we are having again log-in issues.

On 2026-08-11,00:08:05 Irish time (2026-08-10,23:08:05 UTC) I started automatically my car charge my Kona-odo script worked fine, but on 2026-08-11,07:58:03 Irish time (2026-08-11,06:58:03 UTC) when I stopped the charge, the script failed and it is failing ever since.

```
Traceback (most recent call last):
  File "/cinemateka/scripts/Kona-odo.py", line 22, in <module>
    vm.check_and_refresh_token()
  File "/usr/lib/python3.12/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 196, in check_and_refresh_token
    if self.login() is True:
       ^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/hyundai_kia_connect_api/VehicleManager.py", line 98, in login
    result = self.api.login(
             ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 182, in login
    access_token, refresh_token, expires_in = self._login_with_password(
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 293, in _login_with_password
    raise AuthenticationError(
hyundai_kia_connect_api.exceptions.AuthenticationError: Authentication failed — returned to login page. Check username and password.
root@cinemateka:[system-connections]#
```

The script is very simply and has not changed for a while:

```
root@cinemateka:[scripts]# cat Kona-odo.py
#!/usr/bin/python3

from datetime import datetime
from pytz import timezone
import sys
import logging
from hyundai_kia_connect_api import *

import importlib.util
spec = importlib.util.spec_from_file_location("libKona", "/cinemateka/libs/libKona.py")
libKona = importlib.util.module_from_spec(spec)
sys.modules["libKona"] = libKona
spec.loader.exec_module(libKona)

force_refresh_interval = 3600

Start = datetime.now(timezone('UTC'))
TimeFormat = "%Y-%m-%d,%H:%M:%S"

[...]

PR #1276, opened 2026-08-15
fix(EU): detect IdP WAF block on login step 1 and raise a clear error

xtux77:
Since ~2026-08-11 the EU IdP WAF rejects **step 1** of `_login_with_password()` (`GET /auth/api/v2/user/oauth2/authorize`) **before any credential is sent**, because the registered `redirect_uri` carries a non-standard `:8080` port. The request 302-redirects to `/error?status=400` with an `"abusing request"` body, so no session cookies are set and the later signin bounces back to the login page — surfacing the misleading `AuthenticationError: Check username and password`.

Root cause and reproduction are documented in #1273 (blocked with `:8080`, passes without a port but then the IdP correctly rejects the port-less URI). URL-encoding the colon (`%3A8080`) does **not** bypass it either — the WAF normalises before matching.

### What this PR does
This does **not** fix the login itself (the block is server-side on Hyundai/Kia). It only detects the block right after step 1 and raises an explicit `APIError` instead of the misleading credentials message, so users stop resetting perfectly valid passwords.

- Captures the step-1 response and checks for the `"abusing request"` / `/error?status=400` signal (it surfaces in both `r.text` and the followed `r.url`).
- Raises `APIError` (not `AuthenticationError`) on purpose: it is not a credentials failure and must not trigger retry-on-bad-credentials logic. `APIError` (rather than `ServiceTemporaryUnavailable`) also avoids retry loops hammering a WAF that is already blocking us.
- No behavioural change on the happy path: when step 1 is not blocked, the response is discarded exactly as before.

Refs #1273.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -35,6 +35,7 @@ from .const import (
     VALET_MODE_ACTION,
 )
 from .exceptions import (
+    APIError,
     AuthenticationError,
     ConsentRequiredError,
 )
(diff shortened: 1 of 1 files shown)


blka:
Honestly I am not sure if ApiError is the right option it there as it could be temporary - see https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1840

xtux77:
Closing this — it has been overtaken by #1277.

That PR replaces the legacy authorize (the `:8080` redirect the IDPConnect WAF was blocking) with the OneApp `client_id` flow, so the EU login no longer goes through the request this PR was detecting. A clearer error message for a call we no longer make isn't worth carrying.

@blka you were right to question whether `APIError` was the right type given the block could be temporary — the point is moot now that the flow itself changed. Thanks for looking at it, and thanks for the quick turnaround on the actual fix; confirmed working here after bumping the library.

Happy to revisit if the WAF ever starts affecting the new flow.

PR #1277, opened 2026-08-15
fix(eu): bypass IDPConnect WAF via OneApp/CCI login for Hyundai+Kia

blka:
## Problem

EU Hyundai/Kia password login has been broken since 2026-08-11 (#1273). The IDPConnect WAF now blocks the legacy `GET .../oauth2/authorize` request for `client_id=6d477c38` (Hyundai) / `fdc85c00` (Kia) with a `:8080` `redirect_uri`:

```
302 → /error?status=400&error=Bad+Request
      &error_description=Bad Request. It was classified as an abusing request and blocked
```

Because that request is refused, no session cookies are set and `POST /auth/account/signin` bounces back to the login page, producing the misleading `AuthenticationError: returned to login page. Check username and password`.

The block is **server-side** with **no client-side workaround** in the legacy flow: the only `redirect_uri` the IdP accepts is the `:8080` one its own WAF now blocks.

## Fix

Switch EU Hyundai/Kia password login to the OneApp client_id flow (`4f4953b5` Hyundai / `01b36c86` Kia), whose authorize is **not** WAF-blocked:

1. `GET authorize` (OneApp client_id, `oneapp.{hyundai,kia}.com/redirect`) — passes WAF
2. `GET certs` → RSA public key
3. `POST /auth/account/signin` (RSA, `state=ccsp`, full email) → auth code
4. `POST cci-api-eu.{hyundai,kia}.com/domain/api/v1/auth/token?code=` → CCI token set
5. `POST cci-api-eu/domain/api/v1/auth/token-exchange?serviceType=CCS` → CCS token

The CCS token is accepted by the **existing legacy `ccapi:8080`** vehicle/control endpoints as a Bearer access_token, so `Stamp`, `device_id` (`notifications/register`) and **every `ccapi:8080` call stay unchanged** — only the login/token-acquisition path changes.

Refresh uses `cci-api-eu/domain/api/v2/auth/token-refresh` with the full CCI token set, then re-exchanges the CCS token.

## What is preserved

- Legacy 48-char refresh path (`_get_access_token`) — backward compat for users with a pre-WAF refresh token, and for Genesis.
- **Genesis EU** keeps the legacy flow entirely (`PORT=443`, not WAF-affected).
- All other regions (USA/CA/AU/IN/CN/BR) are separate API classes — untouched.
- `Stamp`, `device_id`, `_get_cookies`, `_set_session_language`, all vehicle/control endpoints — untouched.

## Verification

- **Live tested, both brands** (EU account):
[...]

PR #1284, opened 2026-08-18
feat(eu-cci): shared GspaApiEU base + KiaCciApiEU (rework of #1284)

### LINKED ISSUE #1273: And new log in issues again today (at some point after 2026-08-10,23:08:05 UTC)

blka:
## Rework: shared EU GSPA base class + Kia EU support

This rework restructures the EU region into a shared base class with
brand-specific subclasses, and adds Kia EU support on top of it.

### What changed

- **`GspaApiEU` (new shared base)** — everything brand-neutral:
  - CCI (OneApp) password login flow, confirmed against production endpoints
  - Token lifecycle: `refresh_access_token`, `test_token`, `_exchange_ccs_token`, `_refresh_cci_token`
  - X-Stamp request-signing machinery and the GSPA secure-request layer (`_gspa_get`)
  - CCI header construction (`_get_cci_headers`)
  - `get_vehicles` / `_parse_vehicles_from_cci` (both brands use the same CCI
    envelope — `ccspCarId` / `ccspVehicle` fields confirmed on the shared
    `available-vehicles` endpoint)
- **`HyundaiCciApiEU`** — brand constants + Hyundai-specific parsers
  (driving info/history, breakdowns/DTC, ccs2 vehicle-properties, cached state)
  plus the Hyundai remote-action paths (`force_refresh_vehicle_state` /
  `prewakeup`).
- **`KiaCciApiEU` (new)** — Kia EU support via the same CCI login flow with the
  Kia OneApp client id. Login and vehicle listing are inherited from the base;
  cached-state parsing, extended reads, and the remote-action paths
  (`force_refresh_vehicle_state` / `prewakeup`) raise `NotImplementedError`
  until live-response fixtures from a real Kia vehicle are available.
- **`VehicleManager` wiring** — `REGION_EUROPE_CCI` + KIA → `KiaCciApiEU`;
  GENESIS → explicit `APIError` (no Genesis CCI flow yet); default Hyundai.

### Unified cipher engine

The GSPA request-signing cipher is unified into one parameter-driven engine
(`gspa/cipher_keys.py`): a pure-Python AES-128 implementation with
per-parameter affine layers. Hyundai and Kia differ only by data-driven
parameter sets (`hyundai_cipher_params.json` / `kia_cipher_params.json` plus
mask LUTs), selected by brand — no new dependencies. Hyundai vectors remain
bit-exact unchanged (behavior-preserving proof in tests); Kia parameters are
validated against a pinned reference vector.

### Review fixes

Final-review pass addressed in a follow-up commit on this branch:

- X-Stamp signing decoupled from the library region id (the stamp region is
  now a class constant; regression tests exercise the exact production wiring)
- GSPA error handling reordered: non-JSON bodies (e.g. gateway blocks) surface
  as `APIError` with a body preview instead of a raw parse error; 403 reports
  the API resCode/message
- `force_refresh_vehicle_state` / `prewakeup` moved to `HyundaiCciApiEU`
  (spec §3); Kia inherits/raises `NotImplementedError`
- `smart_key_battery_warning_is_on` uses `bool_or_none` (unknown ≠ off)

### Live verification

- **Hyundai** (Santa Fe HEV, regression gate for the refactor): login OK,
  `get_vehicles` returns the vehicle with correct id/VIN/model.
  GSPA stored-status (stamp-signed read) re-verified **after** the
  stamp-region fix: HTTP 200 with the expected payload on live credentials.
- **Kia** (new machinery): CCI login OK (same token shape as Hyundai);
  `available-vehicles` returns HTTP 200 with the shared `contents` envelope
  (empty list — test account has no Kia vehicle).

### Tests

- Full suite: 589 passed + 15 skipped (snapshot tests skip locally when
  `syrupy` is not installed; CI installs it)
- New: `test_stamp_region_regression.py` (stamp region through the exact
  production wiring + bit-exact pin), `test_gspa_api_eu_base.py` (base
  machinery), `test_kia_cci_api_eu.py` (Kia constants, cipher selection, login
  stub flow, shared vehicle-list path, manager wiring), `test_kia_cipher.py`
  (Kia parameter set + pinned vectors)
- Existing Hyundai CCI tests migrated to patch the shared base

### Notes

- Kia EU login + vehicle listing are usable now; the cached-state parser is
  gated on a real Kia vehicle response (Kia accounts without vehicles get an
  honest empty list).
- Genesis EU CCI is out of scope for this PR (explicit `APIError`).

### Follow-up (separate PRs)

- MQTT push (gspa-ccs-eu service hub)
- GSPA control (lock/climate/charge/windows)
- OTA firmware
- kia_uvo companion: config flow for `REGION_EUROPE_CCI` + manifest bump

Fixes #1273. Follow-up to #1277/#1278/#1279.


hyundai_kia_connect_api/Token.py
```diff
@@ -33,6 +33,8 @@ class Token:
     non_ccs_token: str | None = None
     non_ccs_refresh_token: str | None = None
     id_token: str | None = None
+    # CCS user ID (uid claim from ccs_token JWT) — used for GSPA X-Stamp.
+    ccs_user_id: str | None = None
```
cdnninja: Could we just name this User ID to make it so it can be used by other regions if they call it something different?

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -0,0 +1,1629 @@
...
+        if language not in SUPPORTED_LANGUAGES_LIST:
+            _LOGGER.warning(f"Unsupported language: {language}, fallback to en")
+            language = "en"
+
+        self.region: int = region
+        self.LANGUAGE: str = language
+        self.brand: int = brand
+
+        if BRANDS[self.brand] == BRAND_HYUNDAI:
+            # Confirmed production endpoints.
```
cdnninja: Not sure this comment brings value.

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -0,0 +1,1629 @@
...
+            self.ONEAPP_CLIENT_ID: str = "4f4953b5-02e1-4dbc-8599-87e983ee1be5"
+            self.ONEAPP_REDIRECT_URI: str = "https://oneapp.hyundai.com/redirect"
+            self.CCI_API_URL: str = "https://cci-api-eu.hyundai.com"
+            self.LOGIN_FORM_HOST: str = "https://idpconnect-eu.hyundai.com"
+            self._cci_package_id: str = "com.hyundai.oneapp.eu"
+            self._cci_client_name: str = "hyundai"
```
cdnninja: This feels like duplicate of brand?

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -0,0 +1,1629 @@
...
+        elif BRANDS[self.brand] == BRAND_KIA:
+            raise NotImplementedError(
+                "Kia CCI EU not yet implemented — use KiaUvoApiEU for Kia EU."
+            )
+        elif BRANDS[self.brand] == BRAND_GENESIS:
+            raise NotImplementedError(
+                "Genesis CCI EU not yet implemented — use KiaUvoApiEU for Genesis EU."
+            )
+        else:
+            raise APIError(f"Unknown brand {BRANDS[self.brand]} for CCI EU API")
```
cdnninja: This isn't an API error.

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -0,0 +1,1629 @@
...
+        self.CCSP_API_URL: str = "https://gspa-ccs-eu.hyundai.com"
+        self.CCSP_SERVICE_ID: str = "6d477c38-3ca4-4cf3-9557-2a1929a94654"
+        self._cci_client_version: str = "1.3.3"
+        self._cci_client_os_version: str = "18.7"
+        self._cci_notification_provider: str = "APNS"
+
+        self.session = ApiImplSession()
+
+    # ------------------------------------------------------------------
+    # Login
```
cdnninja: This comment doesn't seem to follow the rest of the file.

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -0,0 +1,1629 @@
...
+    ) -> Token:
+        """Login via CCI flow and return a Token with all CCI fields.
+
+        Generates a local device_id (UUID), runs the CCI password login,
+        registers the device on CCI, and extracts the CCS user-id for
+        GSPA X-Stamp computation.
+        """
+        device_id = str(uuid.uuid4())
+
+        info = self._login_with_password(username, password, device_id)
```
cdnninja: info isn't a great variable name.

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -0,0 +1,1629 @@
...
+        self._register_device(token)
+
+        # Extract CCS user-id for GSPA X-Stamp (best effort).
+        self._fetch_ccs_user_id(token)
+
+        return token
+
+    def _login_with_password(
+        self, username: str, password: str, device_id: str
+    ) -> dict:
```
cdnninja: Would it be easier to just return the token here instead of creating a dict then mapping it a second time?  I am good either way.

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -0,0 +1,1629 @@
...
+    # ------------------------------------------------------------------
+
+    def refresh_access_token(self, token: Token) -> Token:
+        """Refresh access token using the stored CCI token set.
+
+        CCI flow: POST v1/auth/token-refresh with the full token set,
+        then re-exchange the CCS token. Falls back to full login if
+        the refresh token is missing or the exchange fails.
+        """
+        if getattr(token, "cci_access_token", None) or getattr(
```
cdnninja: Does this need getattr? I would think it always exists as it is defined in the dataclass token. If token.cci_access_token or token.non_ccs_token.

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -0,0 +1,1629 @@
...
+                region=self.region,
+                tsid=tsid,
+                epoch_seconds=epoch_seconds,
+                user_id=user_id,
+            )
+            return stamp, tsid
+        except NotImplementedError:
+            raise
+        except Exception as e:
+            _LOGGER.debug(f"{DOMAIN} - X-Stamp computation error: {e}")
```
cdnninja: Wouldn't this cause it to fail so you want the exception to be created / pass on?

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -0,0 +1,1629 @@
...
+            "Content-Type": "application/json",
+            "User-Agent": USER_AGENT_OK_HTTP,
+        }
+        stamp, tsid = self._get_stamp(token)
+        if stamp and tsid:
+            headers["X-Stamp"] = stamp
+            headers["X-Request-Id"] = tsid
+        return headers
+
+    def _ensure_ccs_token(self, token: Token) -> None:
```
cdnninja: The method name isn't very clear.

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -0,0 +1,1629 @@
...
+            return None
+
+    def _update_vehicle_driving_history(self, vehicle: Vehicle, state: dict) -> None:
+        """Parse 30-day driving history into power_consumption_30d and daily_stats."""
+        driving_info_list = state.get("drivingInfo", [])
+        if not driving_info_list:
+            return
+
+        for item in driving_info_list:
+            if not isinstance(item, dict):
```
cdnninja, line 120: I am not sure I understand this.  What does this do?  If not matter the value it continues.

cdnninja (commented):
Few minor comments at my first run though.

hyundai_kia_connect_api/HyundaiCciApiEU.py
cdnninja: Following up on this.

hyundai_kia_connect_api/HyundaiCciApiEU.py
cdnninja: Following up on this.

blka: Kept the dict return — `_login_with_password` doesn't have `pin` (the user's PIN), which `login()` adds to the Token. Also, `_register_device` and `_fetch_user_id` need the Token object, so the steps are: 1) get CCI/CCS tokens (dict), 2) create Token with pin, 3) register device, 4) fetch user-id. I can refactor to return Token directly if preferred — pin could be passed to `_login_with_password` as well.

hyundai_kia_connect_api/HyundaiCciApiEU.py
blka: This filters the `drivingInfo` response for the summary period (`drivingPeriod == 0`). The API returns a list of driving periods — period 0 is the total/summary that contains `totalDistance` (odometer), `totalPwrCsp` (total power consumption), and `calculativeOdo` (calculated odometer). Other periods (1, 2, ...) are individual trips. The `continue` skips non-summary entries so only the aggregate is parsed into `total_driving_range` and `power_consumption_30d`.

blka:
Friendly ping on this one — no movement since late August. The follow-up work I have queued depends on this landing, so happy to address anything blocking whenever you get to it.

PR #1285, opened 2026-08-19
fix(BR): surface 4xx body on /spa/vehicles instead of raw HTTPError

blka:
## Problem

`GET /spa/vehicles` returned a bare `400 Bad Request` with the response body discarded (`raise_for_status()`), so users saw `400 Client Error: Bad Request for url: https://br-ccapi.hyundai.com.br/api/v1/spa/vehicles` with no actionable reason.

Reported in [Hyundai-Kia-Connect/kia_uvo#1846](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1846). Same recurring symptom as [Hyundai-Kia-Connect/kia_uvo#1395](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1395) (Nov 2025, closed by stale-bot, never root-caused): login succeeds — the OAuth2 token is issued — but the vehicles call fails opaquely.

BR was the **only** Type1 region using `raise_for_status()` on SPA endpoints. Every other region (EU, CA, AU, IN, CN, USA) and the base `ApiImplType1` use `.json()` + `_check_response_for_errors()` — so a SPA-envelope 4xx is classified into a typed error (`DeviceIDError`, `DuplicateRequestError`, …) instead of a raw `HTTPError`.

## Change (revised per review)

An earlier revision added a `_raise_api_error` helper. @cdnninja flagged it as clunky vs. the established pattern — correctly. This revision **drops the helper** and conforms BR to the Type1 pattern:

- `get_vehicles`: `self.session.get(...).json()` → `_check_response_for_errors(response)` → parse. No `raise_for_status()`.
- `_get_cached_vehicle_state` (`/spa/vehicles/{id}/ccs2/carstatus/latest`): same pattern.

The shared `_check_response_for_errors` already classifies SPA envelopes into typed exceptions and logs unknown response shapes, so a vehicles 4xx now surfaces a readable typed error (and the body for unknown shapes) instead of a bare `400 Bad Request`. No region-specific helper, no second body-reading method.

The auth helper `_raise_auth_error` from #1259 stays — the auth endpoints (`/user/oauth2/*`, `/user/signin`) return a different schema (`step`/`errCode`/`errMsg`, non-JSON WAF), for which no Type1 standard applies; that was reviewed and merged in #1259.

## What this does NOT do

This is a **diagnostic** fix, not a root-cause fix. The root cause of the `/spa/vehicles` 400 is still unknown (account-state / server-side flakiness, as in #1395). But with this change the next reporter sees a typed error / the response body instead of a bare `400 Bad Request`, which is the prerequisite to diagnosing it.

## Verification

- `ruff check .` ✅ · `ruff format --check .` ✅ · pre-commit ✅
- `pytest tests/` → 534 passed (15 snapshots)
- New `TestGetVehicles`: SPA-envelope 4xx → typed `DeviceIDError` (not raw `HTTPError`), non-SPA 4xx → `InvalidAPIResponseError` (not raw `HTTPError`), 200 success path.
- `test_br_vehicle_properties._response` helper now sets `status_code = 200` (realistic success mock; required because the conforming path no longer calls `raise_for_status()`).

## Related

- [Hyundai-Kia-Connect/kia_uvo#1846](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1846) — this report
- [Hyundai-Kia-Connect/kia_uvo#1395](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1395) — same symptom, closed stale
- #1259 — BR auth 4xx + CCS2 status (does not cover `get_vehicles`)

### DIFF
--- a/hyundai_kia_connect_api/HyundaiBlueLinkApiBR.py
+++ b/hyundai_kia_connect_api/HyundaiBlueLinkApiBR.py
@@ -12,7 +12,7 @@ from urllib.parse import parse_qs, urljoin, urlparse
 from requests import Response
 
 from .ApiImpl import ApiImplSession, ClimateRequestOptions, WindowRequestOptions
-from .ApiImplType1 import ApiImplType1
+from .ApiImplType1 import ApiImplType1, _check_response_for_errors
 from .const import (
     BRAND_HYUNDAI,
     BRANDS,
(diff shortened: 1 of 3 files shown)


cdnninja:
Does this feel clunky to you as far as consistency to other regions and multiple methods doing similar?

blka:
@cdnninja Fair point — it was clunky. BR was the only Type1 region using `raise_for_status()` on SPA endpoints; every other region (and the base `ApiImplType1`) does `.json()` + `_check_response_for_errors()`. Dropped the helper in cce6a32 and conformed: `get_vehicles` and `_get_cached_vehicle_state` now follow that pattern, so a SPA-envelope 4xx is classified by the shared `_check_response_for_errors` into a typed error instead of a raw `HTTPError`. The auth helper (`_raise_auth_error`) stays — the auth endpoints return a different schema (`step`/`errCode`/`errMsg`, non-JSON WAF) that the SPA classifier doesn't cover, which is what #1259 addressed.

Issue #993, opened 2026-01-11
EU/Kia: GPS coordination details are received via API but not extracted.


jk5836:
I'm in the EU (Sweden) with a kia E-Niro (my 21), When it retrieves the data the following error occurs...

2026-01-11 18:26:58.146 DEBUG (SyncWorker_4) [hyundai_kia_connect_api.KiaUvoApiEU] hyundai_kia_connect_api - _get_location response: {'retCode': 'S', 'resCode': '0000', 'resMsg': {**'gpsDetail'**: {'coord': {'lat': xx.xxxxx, 'lon': yy.yyyyy, 'alt': 0, 'type': 0}, 'head': 0, 'speed': {'value': 0, 'unit': 0}, 'accuracy': {'hdop': 0, 'pdop': 0}, 'time': '20260111172656'}, 'drvDistance': {'rangeByFuel': {'evModeRange': {'value': 165, 'unit': 1}, 'totalAvailableRange': {'value': 165, 'unit': 1}}, 'type': 2}}, 'msgId': '...msgid...'}
2026-01-11 18:26:58.147 ERROR (SyncWorker_4) [hyundai_kia_connect_api.KiaUvoApiEU] hyundai_kia_connect_api - _get_location failed: 'coord'
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/hyundai_kia_connect_api/KiaUvoApiEU.py", line 816, in _get_location
    return **response["resMsg"]["coord"]**
           ~~~~~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'coord'

No coordination data is available..

I guess this should be: return response["resMsg"]**["gpsDetail]"]**["coord"]

cdnninja:
Could you create a pull request?

erikgeurts:
It looks like what's needed is to reverse the change in this PR: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/912

PR #1292, opened 2026-08-28
fix(EU): convert forced refresh location time from UTC

### LINKED ISSUE #1291: EU forced refresh stamps location time one UTC offset early
- Hyundai / Kia Connect version: 4.27.1
- Python version: 3.12
- Operating System: Linux (Debian, container)

### Description

On a **non-CCS2 EU vehicle**, a forced refresh stamps `location_last_updated_at` one UTC offset
early: two hours in CEST, one hour in CET. A cached update of the same position gives the correct
time. Thus the sensor moves backwards each time a force refresh occurs.

The two paths read different endpoints, and the endpoints do not use the same clock:

| path | endpoint | field | clock |
| --- | --- | --- | --- |
| `update_vehicle_with_cached_state` | `/location/park` | `resMsg.time` | `Europe/Berlin` (`data_timezone`) |
| `force_refresh_vehicle_state` | `/location` | `resMsg.gpsDetail.time` | **UTC** |

`KiaUvoApiEU._get_location` returns the `gpsDetail` of `GET /location`, and
`force_refresh_vehicle_state` puts it into `state["vehicleLocation"]`:

```python
state = self._get_forced_vehicle_state(token, vehicle)
state["vehicleLocation"] = self._get_location(token, vehicle)
```

`_update_vehicle_properties` then labels that time with `data_timezone`:

```python
parse_datetime(
    get_child_value(state, "vehicleLocation.time"), self.data_timezone
)
```

`data_timezone` is `ZoneInfo("Europe/Berlin")`, and for a compact `YYYYMMDDHHMMSS` value
`utils.parse_datetime` only *labels*, it does not convert. The same function *does* convert when
the format states its zone: `"%a, %d %b %Y %H:%M:%S GMT"` is read as UTC and then converted with
`astimezone(timezone)`. So the library already converts where the convention of the source is
known. Only the compact branch cannot know it, and here the caller does.

The payload carries no zone marker: `gpsDetail` holds `coord`, `head`, `speed`, `accuracy` and
`time`, and no `Offset`.

### What I Did

I see this on my own EU non-CCS2 vehicle. Four consecutive forced refreshes each returned a
location time exactly the current UTC offset behind the cached time for the same fix, while the
cached read of that same fix was correct.

The clearest public evidence is in Hyundai-Kia-Connect/kia_uvo#1639. That log holds **both** endpoints for the same
car, 0.6 s apart, because a forced update and a cached update overlapped in two worker threads:

```
2026-05-02 16:54:55.005 DEBUG (SyncWorker_2) ... - _get_location response:
    {'resMsg': {'gpsDetail': {..., 'time': '20260502145453'}}}
2026-05-02 16:54:55.648 DEBUG (SyncWorker_5) ... - _get_location response:
    {'resMsg': {'gpsDetail': {..., 'time': '20260502165453'}}}
```

Same method name, same second, two values two hours apart. SyncWorker_2 is the forced update
(`Received forced vehicle data` appears on that thread at 16:54:51.734), so its call is
`_get_location` → `/location`. SyncWorker_5 is the cached update
(`get_cached_vehicle_status response` at 16:54:55.434, carrying `'time': '20260502165453'`), so
its call is `_set_cached_location_park` → `/location/park`. The log wall clock is 16:54:55 in
CEST. `20260502145453` is that instant in UTC; `20260502165453` is it in `Europe/Berlin`.

Hyundai-Kia-Connect/kia_uvo#931 is the same picture two years earlier. The reporter states his offset up front —
"Log times are in CET ( UTC+2) - watch to 'time' in the json" — and notices the split himself:

> There is `'time': '20240920132506'` in the `get_location response` but
> `'time': '20240920152506'` in the `get_cached_vehicle_status response`

Two further logs show the forced value alone, each within a second or two of the wall clock of
the log line, the difference being request latency:

| report | offset | log wall clock | `/location` `gpsDetail.time` |
| --- | --- | --- | --- |
| Hyundai-Kia-Connect/kia_uvo#1451 | +01:00 | 10:59:08 | `20260112095907` |
| #993 (Sweden) | +01:00 | 18:26:58 | `20260111172656` |

All of these are central European accounts, so the shift they show is the offset of
`data_timezone` itself. I have not found a capture from a user outside that zone that would
separate "the field is UTC" from "the field is the user's local time"; if anyone has one, it
would be useful here.

**One caution about reading any of these logs.** `_set_cached_location_park` writes the same
`_get_location response:` message as `_get_location`, as the extract above shows, so the
message alone does not say which endpoint answered. Also, between #1083 and #1095 — releases
4.8.3 to 4.10.1, 12 to 19 April 2026 — `_get_location` itself pointed at `/location/park`. The
reports above are all outside that window.

### Affected versions

Present in 4.27.1, the current release. `_get_location` has read `/location` since long before
4.8.2, and again from 4.10.2 onward. Only 4.8.3 to 4.10.1 was free of this, because in those
releases `_get_location` read `/location/park` instead.

### Two questions you will probably ask

**Why not change `_update_vehicle_properties`?** That same branch also receives the
`vehicleLocation` embedded in the cached status response, which really is `data_timezone` time. A
conversion there would move the cached feed wrong by the same one or two hours in the other
direction.

**If `/location` is UTC, why is `vehicleStatus.time` on the same forced response not also wrong?**
It is a different field from a different endpoint (`/status`), and no report shows it as shifted.
I did not change it. If someone can show a paired capture where it is also UTC, that deserves its
own issue.

### Not part of this report

- On a CCS2 vehicle, `_get_location` is never called: `force_refresh_vehicle_state` sends CCS2
  cars to `_force_refresh_vehicle_state_ccs2`, which ends at `_set_cached_location_park`. In
  #1096 there is an observation that a direct location call on a ccNC car returned local time.
  That is consistent with this report: a different protocol and a different code path.
- `ApiImplType1` labels `Location.TimeStamp` with `data_timezone` as well. On EU and AU the park
  override hides it, but perhaps not for every region. It is a different payload shape and needs
  its own capture.
- `_set_cached_location_park` reads `resMsg.coord`, yet `/location/park` can answer with
  `resMsg.gpsDetail.coord` — as in the 16:54:55.648 line above, where the override therefore
  matched nothing and returned in silence. On the captures I have this changes no value, because
  the `vehicleLocation` embedded in the same cached read carries the same coordinates and time.
  I note it only because it is the same extract, and someone reading that log will wonder.

### Suggested fix

Convert where the source is known to be UTC, and leave the cached feed alone. This is what
PR #1147 did for the CCS2 `Date` field:

```python
    get_child_value(state, "Date"), dt.timezone.utc
).astimezone(self.data_timezone)
```

The same treatment applies to the `gpsDetail` of `/location`, on the forced path only. Converting
from UTC also avoids the ambiguous hour that relabelling a local string would introduce at the
autumn DST change.

I have a patch with tests and will open a pull request that refers to this issue.

rogiervandergeer:
## Summary

On a non-CCS2 EU vehicle a forced refresh stamps `location_last_updated_at` one UTC offset early:
two hours in CEST, one hour in CET. The cached path gives the correct time for the same fix, so
the sensor moves backwards on every force refresh.

The cause is that the two paths read endpoints with different clocks, and both end in the same
line of `_update_vehicle_properties`:

| path | endpoint | field | clock |
| --- | --- | --- | --- |
| `update_vehicle_with_cached_state` | `/location/park` | `resMsg.time` | region local |
| `force_refresh_vehicle_state` | `/location` | `resMsg.gpsDetail.time` | **UTC** |

`utils.parse_datetime` only labels a compact `YYYYMMDDHHMMSS` value with the zone it receives, it
does not convert it, so a UTC value becomes an `Europe/Berlin` instant.

Full analysis and the user reports are in #1291.

## Changes

- `hyundai_kia_connect_api/KiaUvoApiEU.py` — a new `_set_location_from_gps_detail`, called from
  `force_refresh_vehicle_state` after `_update_vehicle_properties`. It parses `gpsDetail.time`
  as UTC and converts into `data_timezone`, which is the treatment that #1147 already applies to
  the CCS2 `Date` field.
- `hyundai_kia_connect_api/utils.py` — a docstring on `parse_datetime` that states the difference
  between its two formats. Comment only.
- `tests/test_eu_forced_location_timezone.py` — new.
- `tests/utils_test.py` — one test that pins the difference between the two formats.

## Why the change is at the call site, and not in `_update_vehicle_properties`

That branch serves two feeds. Besides the forced `/location` payload, it also receives the
`vehicleLocation` embedded in the cached status response, which really is region local time. A
conversion inside `_update_vehicle_properties` would move the cached feed wrong by the same one
or two hours in the other direction. `test_cached_embedded_location_time_stays_local` fails if
someone later makes that change.

Writing the value and then overriding it is the shape this file already uses: on the cached path
and on the CCS2 forced path, `_set_cached_location_park` also runs after the properties are set.

The raw payload in `vehicle.data` keeps the original string, so a later diagnosis still sees what
the server sent.

## Scope

EU, non-CCS2, forced refresh only.

- `KiaUvoApiEU` has no subclasses, and `_get_location` has exactly one caller.
- A CCS2 vehicle never reaches this code: `force_refresh_vehicle_state` sends it to
  `_force_refresh_vehicle_state_ccs2`, which ends at `_set_cached_location_park`.
- `KiaUvoApiAU._get_location` reads `/location/park` and is untouched.
- `KiaUvoApiCN` reads `/location` but with a different response shape and no DST. I have no
  capture for it and did not change it.

## Tests

Four new tests for the EU path, one for `parse_datetime`.

Without the source change, `test_forced_location_time_is_parsed_as_utc` fails by two hours and
`test_forced_location_time_is_utc_in_winter` fails by one hour, so a constant offset cannot pass
both. The other three pass before and after; they guard the cached feed, the offline vehicle and
the missing timestamp.

`pytest` on the full suite gives `540 passed`.

`pre-commit run --all-files` is clean, and so is
`flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics`.

## Related issues

Closes #1291.

Same symptom reported downstream in Hyundai-Kia-Connect/kia_uvo#1639,
Hyundai-Kia-Connect/kia_uvo#931 and Hyundai-Kia-Connect/kia_uvo#1451, and here in #993.
#1147 is the CCS2 `Date` precedent that this follows.

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -672,8 +672,13 @@ class KiaUvoApiEU(ApiImplType1):
             self._force_refresh_vehicle_state_ccs2(token, vehicle)
         else:
             state = self._get_forced_vehicle_state(token, vehicle)
-            state["vehicleLocation"] = self._get_location(token, vehicle)
+            location = self._get_location(token, vehicle)
+            state["vehicleLocation"] = location
             self._update_vehicle_properties(vehicle, state)
+            # Must run after _update_vehicle_properties: that method labels
+            # vehicleLocation.time with data_timezone, which is correct for the
+            # cached feed but wrong for the GET /location payload attached above.
+            self._set_location_from_gps_detail(vehicle, location)
         # Only call for driving info on cars we know have a chance of supporting it.
         # Could be expanded if other types do support it.
         if (
@@ -1293,6 +1298,35 @@ class KiaUvoApiEU(ApiImplType1):
         except Exception as e:
             _LOGGER.exception(f"{DOMAIN} - _get_location failed: {e}")  # noqa: TRY401
             return None
+
+    def _set_location_from_gps_detail(
+        self, vehicle: Vehicle, gps_detail: dict | None
+    ) -> None:
+        """Apply the gpsDetail of GET /location, whose time is UTC.
+
+        Two feeds fill vehicle.location from the same compact YYYYMMDDHHMMSS
+        shape but on different clocks: this one is UTC, while the
+        vehicleLocation embedded in the cached status response and
+        GET /location/park are region local. parse_datetime only labels a
+        compact value, so _update_vehicle_properties leaves this feed one UTC
+        offset early. Convert here, as the CCS2 'Date' field does (#1147).
+        """
+        if not gps_detail or not get_child_value(gps_detail, "coord.lat"):
+            return
+        # A missing timestamp must stay None, which HA shows as "unknown".
+        # See kia_uvo #1771.
+        location_last_updated_at = parse_datetime(
+            get_child_value(gps_detail, "time"), dt.UTC
+        )
+        if location_last_updated_at is not None:
+            location_last_updated_at = location_last_updated_at.astimezone(
+                self.data_timezone
+            )
+        vehicle.location = (
+            get_child_value(gps_detail, "coord.lat"),
+            get_child_value(gps_detail, "coord.lon"),
+            location_last_updated_at,
+        )
 
     def _get_forced_vehicle_state(self, token: Token, vehicle: Vehicle) -> dict:
         url = self.SPA_API_URL + "vehicles/" + vehicle.id + "/status"
--- a/hyundai_kia_connect_api/utils.py
+++ b/hyundai_kia_connect_api/utils.py
@@ -180,6 +180,15 @@ def get_index_into_hex_temp(value):
 
 
 def parse_datetime(value, timezone) -> datetime.datetime | None:
+    """Parse an API timestamp into an aware datetime.
+
+    The two accepted formats are not equivalent. "Tue, 24 Jun 2025 16:18:10
+    GMT" states its own zone, so it is read as UTC and converted into
+    ``timezone``. The compact "20250624161810" states no zone, so it is only
+    labelled with ``timezone``; no conversion happens. A caller whose compact
+    value is UTC must therefore pass ``datetime.UTC`` and convert itself
+    (see the CCS2 'Date' field, #1147).
+    """
     # Missing timestamp must surface as None (HA renders "unknown") rather than
     # a 2000-01-01 sentinel that renders as "27 years ago". See kia_uvo #1771.
     if value is None:
--- a/tests/test_eu_forced_location_timezone.py
+++ b/tests/test_eu_forced_location_timezone.py
@@ -1,0 +1,156 @@
+"""EU forced refresh: GET /location returns gpsDetail.time in UTC. See kia_uvo #931.
+
+The cached feeds (the vehicleLocation embedded in the status response, and
+GET /location/park) return region local time, so only the forced value needs
+a conversion. The last test guards that difference.
+"""
+
+import datetime as dt
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from hyundai_kia_connect_api.KiaUvoApiEU import KiaUvoApiEU
+from hyundai_kia_connect_api.Vehicle import Vehicle
+
+# One instant, two clocks: the forced /status payload reports local wall clock
+# (18:18:10 CEST), GET /location reports UTC (16:18:10Z).
+FORCED_STATUS = {
+    "retCode": "S",
+    "resCode": "0000",
+    "resMsg": {"time": "20260624181810", "doorLock": True},
+}
+LOCATION = {
+    "retCode": "S",
+    "resCode": "0000",
+    "resMsg": {
+        "gpsDetail": {
+            "coord": {"lat": 52.52, "lon": 13.405, "alt": 10, "type": 0},
+            "head": 180,
+            "speed": {"value": 0, "unit": 0},
+            "time": "20260624161810",
+        }
+    },
+}
+
+
+@pytest.fixture
+def eu_api() -> KiaUvoApiEU:
+    api = KiaUvoApiEU.__new__(KiaUvoApiEU)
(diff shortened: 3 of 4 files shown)


blka:
The evidence gap in #1291 is narrower than it looks. The captures already cited there span both DST seasons, and the shift tracks DST:

| report | month | forced value vs log wall clock |
| --- | --- | --- |
| Hyundai-Kia-Connect/kia_uvo#931 | September (CEST) | −2 h |
| Hyundai-Kia-Connect/kia_uvo#1639 | May (CEST) | −2 h |
| Hyundai-Kia-Connect/kia_uvo#1451 | January (CET) | −1 h |
| #993 | January (CET) | −1 h |

That rules out every clock except UTC:

- If the field carried the user's local time — or the account's, or the car's — the forced value would equal the log wall clock: a zero shift. The observed shifts are 1–2 h.
- If it carried a fixed server-side zone other than UTC, the shift would be constant all year. It is 1 h in winter and 2 h in summer, so it follows DST exactly.

A zone whose offset equals UTC year-round is UTC. A capture from an account outside CET would, I think, only re-confirm this.

The fix itself also looks right to me: applying the conversion after `_update_vehicle_properties` matches the shape `_set_cached_location_park` already uses on the other two paths, and parsing as UTC instead of relabelling avoids the ambiguous hour at the autumn DST change.

cdnninja:
@blka to confirm you are good with this being merged? Your comment isn't clear to me.

blka:
Yes — good to merge from my side. Plain restatement of what I was arguing: the forced-refresh location time is sent in UTC (winter reports are off by 1 h, summer by 2 h, so the offset follows DST — only UTC does that). The PR converts it to the correct UTC-aware parse instead of relabelling it. No open concerns; happy to answer anything else.

PR #1294, opened 2026-08-29
fix(CA): fetch location when none is known after restart/reload

### LINKED ISSUE #1844: Canada: Location and Geocoded location sensors unavailable again
### Region

CA

### Brand of Car

Kia

### Version of the Integration

3.10.1

### Describe the Bug

This issue seems to be back, for me at least, as of today. v3.10.1. Last time this happened it was gone for weeks and then came back.

### Debug Logs (if an error occurred)

Debug logging doesn't provide any more information.  That's just from enabling debug logging from the HACS integration.  

```
Logger: hyundai_kia_connect_api.KiaUvoApiCA
Source: /usr/local/lib/python3.14/site-packages/hyundai_kia_connect_api/KiaUvoApiCA.py:932
First occurred: 2:15:05 PM (1 occurrence)
Last logged: 2:15:05 PM

hyundai_kia_connect_api - Get vehicle location failed
```

### Diagnostics dump (if available)

_No response_

### To Reproduce

Reboot HASS, reload the integration or force a refresh.

### Expected Behavior

Location and geocoded location sensors should show location.

### Screenshots

_No response_

### Additional Context

_No response_

blka:
## Summary

In the CA region the vehicle location comes exclusively from the Find-My-Car call (`fndmcr`, PIN-protected) — the cached status's `vehicleLocation` field is never read. `KiaUvoApiCA` only calls it when the odometer increased since the service reading ("the car has moved").

Coordinates live only in the in-memory `Vehicle` object: after a Home Assistant restart / integration reload they are gone, and for a parked car the moved-gate never fires again. The device_tracker and the geocoded-location sensor then stay unavailable until the car physically moves — which can be weeks (Hyundai-Kia-Connect/kia_uvo#1844: "gone for weeks and then came back", reproduces on reboot / reload / force refresh).

## Changes

- `KiaUvoApiCA.update_vehicle_with_cached_state` / `force_refresh_vehicle_state` (sibling call sites): the location gate now also fires while `vehicle.location_latitude` is None. Once a location exists, behaviour is unchanged (moved-gate only), so the extra API load is bounded to the first successful fetch after a restart or a failed one.
- `tests/test_ca_location_unknown_fetch.py`: parked+unknown (fetch — the #1844 case), parked+known (no fetch — no API-load regression), moved (fetch — existing behaviour), plus the force-refresh sibling site. The no-`time` fndmcr case from the #1844 dump (coordinates set, timestamp unknown) is covered by the mocked result.

## Scope

CA region only. HyundaiBlueLinkApiUSA has a structurally similar moved-gate (`HyundaiBlueLinkApiUSA.py` ~line 894) with its own location fallback path — left untouched. kia_uvo needs no change: the entities already render unavailable-while-unknown and become available once the first fetch succeeds.

Fixes Hyundai-Kia-Connect/kia_uvo#1844

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiCA.py
+++ b/hyundai_kia_connect_api/KiaUvoApiCA.py
@@ -476,9 +476,16 @@ class KiaUvoApiCA(ApiImpl):
         # Service Status Call
         service = self._get_next_service(token, vehicle)
 
-        # Get location if the car has moved since last call
+        # Get location if the car has moved since last call, or if no location
+        # is known yet: coordinates live only in the in-memory Vehicle object,
+        # so after a restart/reload a parked car would otherwise never get a
+        # location fetch and the location sensors stay unavailable until the
+        # car physically moves (Hyundai-Kia-Connect/kia_uvo#1844).
         if vehicle.odometer:
-            if vehicle.odometer < get_child_value(service, "currentOdometer"):
+            if (
+                vehicle.odometer < get_child_value(service, "currentOdometer")
+                or vehicle.location_latitude is None
+            ):
                 location = self.get_location(token, vehicle)
                 self._update_vehicle_properties_location(vehicle, location)
         else:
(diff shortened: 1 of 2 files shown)


cdnninja:
One thing for us to watch for with this.  If location fails it will then try wake the car on every call after this.  Not sure if that could have a adverse outcome.

blka:
Thanks — valid concern, and confirmed: `get_location` swallows all errors and returns `None`, so a persistently failing Find-My-Car response would leave the coordinates `None` and re-run the fetch (`vrfypin` + `fndmcr`, two POSTs) on every poll for the whole session — ~96 calls/day at the default 30-minute scan, in exactly the failure mode from Hyundai-Kia-Connect/kia_uvo#1844 (`Get vehicle location failed`).

Hardening PR: Hyundai-Kia-Connect/hyundai_kia_connect_api#1296 — the unknown-location fetch is now one attempt per in-memory `Vehicle` session (a restart/reload starts a fresh attempt, which is the #1844 recovery); the moved-gate is unchanged. The kia_uvo side (geocoded sensor staying "no longer provided" after a reload) is Hyundai-Kia-Connect/kia_uvo#1856.

PR #1298, opened 2026-09-02
fix: Skip no-op HVAC write on EV5 flat scheduled charging

blka:
## Problem

On ccNC/EV5-appMode EVs (EV6, EV9 — `ccu_ccs2_protocol_support==1` + `engine_type==EV`), scheduled charging writes go to two separate endpoints: `/ccs2/reservation/charge` followed by `/ccs2/reservation/hvac`. When a user only changes charging settings, the HVAC write is a no-op (all-off payload with unchanged departure data), and the backend rejects it with `resCode 4004 "Duplicate request"`:

```
POST /ccs2/reservation/charge  -> {"retCode":"S","resCode":"0000"}
POST /ccs2/reservation/hvac    -> {"retCode":"F","resCode":"4004","resMsg":"Duplicate request"}
```

The second failure raises `DuplicateRequestError` for the whole action, so the HA service call reports an error even though the charge write succeeded. Live debug log from [kia_uvo discussion #1764](https://github.com/Hyundai-Kia-Connect/kia_uvo/discussions/1764) (EV6, 2026-08-19) shows exactly this pattern; the combined `/reservation/chargehvac` path (CCSP / EV2–EV4) sends a single write and is not affected.

## Fix

In `_schedule_ev5_flat`, skip the HVAC write when nothing climate-related is requested — both departures disabled **and** `climate_enabled` false. Any climate-related write (enabled departure or `climate_enabled`) still POSTs `/hvac` unchanged.

**Known limitation:** an explicit all-off service call can no longer clear existing HVAC schedules on the car via the API. This is a rare case (all-off writes previously mostly failed with 4004 anyway); clearing schedules through the OEM app still works.

The duplicate-guard mechanism itself (backend rejecting back-to-back writes on these endpoints) is a hypothesis supported by the log above; a separate live test with genuinely different back-to-back writes is planned to confirm.

## Testing

- New `TestEV5HvacSkip` tests: all-off options → only the charge POST; departure or climate enabled → both POSTs.
- Existing 12 flat-dispatch tests updated for the skip and still green; full suite 551 passed, snapshots unchanged.

### DIFF
--- a/hyundai_kia_connect_api/ApiImplType1.py
+++ b/hyundai_kia_connect_api/ApiImplType1.py
@@ -1280,6 +1280,19 @@ class ApiImplType1(ApiImpl):
         _LOGGER.debug(f"{DOMAIN} - EV5 schedule charge response: {charge_response}")
         _check_response_for_errors(charge_response)
 
+        # EV5 climate is a separate endpoint from the charge write. When
+        # nothing climate-related is requested the write is a no-op, and
+        # sending it right after the charge write is rejected by the backend
+        # as a duplicate request (4004), failing the whole action — skip it
+        # instead. Live evidence: KiaUvo discussion #1764 (EV6 debug log
+        # 2026-08-19: charge 0000, then hvac 4004 duplicate).
+        if (
+            not departures[0].enabled
+            and not departures[1].enabled
+            and not options.climate_enabled
+        ):
+            return charge_response["msgId"]
+
         _LOGGER.debug(f"{DOMAIN} - EV5 schedule hvac request: {hvac_payload}")
         hvac_response = self.session.post(
             base_url + "/hvac",
--- a/tests/test_schedule_ev5_flat.py
+++ b/tests/test_schedule_ev5_flat.py
@@ -97,7 +97,10 @@ class TestEV5Dispatch:
     def test_ccs2_ev_uses_flat_endpoints(self, api):
         calls = _mock_post(api)
         v = _make_vehicle(ccs2=1, engine_type=ENGINE_TYPES.EV)
-        api.schedule_charging_and_climate(MagicMock(spec=Token), v, _options())
+        # climate part present (departure enabled) so both endpoints are hit
+        api.schedule_charging_and_climate(
+            MagicMock(spec=Token), v, _options(first_departure_enabled=True)
+        )
         urls = [c["url"] for c in calls]
         assert any(u.endswith("/ccs2/reservation/charge") for u in urls)
         assert any(u.endswith("/ccs2/reservation/hvac") for u in urls)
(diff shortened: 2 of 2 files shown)


hyundai_kia_connect_api/ApiImplType1.py
```diff
@@ -1280,6 +1280,19 @@ def _schedule_ev5_flat(
...
+        # EV5 climate is a separate endpoint from the charge write. When
+        # nothing climate-related is requested the write is a no-op, and
+        # sending it right after the charge write is rejected by the backend
+        # as a duplicate request (4004), failing the whole action — skip it
+        # instead. Live evidence: KiaUvo discussion #1764 (EV6 debug log
+        # 2026-08-19: charge 0000, then hvac 4004 duplicate).
+        if (
+            not departures[0].enabled
+            and not departures[1].enabled
+            and not options.climate_enabled
```
Lekensteyn, line 1292: Is this condition correct?

In the myHyundai app, I can only configure scheduled charging when a departure time is set. EU, non-CCS2 car. When both "Departure 1" and "Departure 2" are switched off, then both "Scheduled charging" and "Scheduled climate control" are also automatically disabled in the app.

Furthermore, wouldn't the current implementation make it impossible to turn off climate control unless a departure date is set?

Lekensteyn (commented):
The current scheduled charging Python API is somewhat annoying to use. I wonder if we should design a better API to control the settings? For now we can probably stick to a minimal change.

Look at all the extra effort that we have to go through to change the settings. One has to extract current settings from various fields or else they will be reset/rejected. I also added a polling loop since the cloud API returns immediately, async. To reduce unnecessary calls while checking for updates, I also patched the API:
https://github.com/Lekensteyn/bluelink-cli/commit/cd76482930b6996b45eade44ba7972afd7006625

If we want to preserve the current API, perhaps we could treat `None` as "I don't want to change it"? Then `schedule_charging_and_climate` could check for `options.climate_enabled is None` or `options.charging_enabled is None`, and skip changes.

blka:
Thanks for the review. Agreed on keeping this PR minimal.

On the `None` proposal — worth noting the current semantics are the opposite: `schedule_charging_and_climate` coerces every `None` to a hard default before building the payload (`enabled→False`, `charging_enabled→False`, `climate_enabled→False`, `temperature→21.0`, …), so today `None` means "write the default", i.e. an explicit reset. That's exactly why the kia_uvo integration rebuilds the full options object from vehicle state before every partial update, and why the HA service call currently resets any field the user omits.

`None` = "leave unchanged" would be a cleaner contract. One constraint: the endpoints take full state per write, so there is no per-field skip over the wire. The only place "unchanged" can mean "skip" is at endpoint granularity on the flat EV5 path — all charge fields `None` → skip `/reservation/charge`, all climate fields `None` → skip `/reservation/hvac` (which would make the skip in this PR a consequence of the general rule rather than a special case). The combined `/ccs2/reservation/chargehvac` path can't drop half its payload, so `None` fields there still need filling from current state.

I'd take this to a separate issue rather than expanding this PR — happy to open one with the consumers list and the endpoint-granularity caveat.

Lekensteyn:
The "Duplicate request" message is suspicious. Does it mean, "no concurrent changes allowed"? Or does it require sending requests after a delay of, say, a second?

If #1303 gets implemented, this PR won't be necessary right? As said before, it will break when users try to disable climate and don't want to touch charge settings.

blka:
Answering the technical question with live-probed data from today (EU CCI/GSPA): "Duplicate request" (400-004) is **channel occupancy, not content dedup** — the server rejects any new remote command while one is still in flight for the VIN:

```
[MT] Precondition : VIN : … URI : /rc/vpush is duplicated request
```

So "no concurrent changes allowed" is the right reading; there is no fixed delay (1 s or otherwise) that clears it — the channel frees when the car acknowledges the command or the polling session expires. A no-op write still occupies the channel, which is exactly the failure mode this PR tried to skip.

And agreed on the strategy: closing this in favor of #1303's leave-unchanged semantics, which avoids the redundant writes that trigger the duplicate rejection in the first place.

blka:
Closing in favor of #1303 (leave-unchanged semantics — see reply above).

Issue #1302, opened 2026-09-03
API design: scheduled charging/climate — None = "leave unchanged" + endpoint-granularity skip


blka:
## Problem

`ScheduleChargingClimateRequestOptions` fields default to `None`, but `schedule_charging_and_climate` coerces every `None` to a hard default before building the payload (`enabled→False`, `days→[0]`, `time→00:00`, `charging_enabled→False`, `climate_enabled→False`, `temperature→21.0`, `defrost→False`, …). So today `None` means "write the default" — an explicit reset, not "leave unchanged".

Consequence for consumers: to change one setting they must extract current settings from various vehicle fields and pass the rest, or the other settings get reset/rejected (raised by @Lekensteyn in #1298). The kia_uvo integration works around this by rebuilding the full options object from vehicle state before every partial update (`_build_schedule_options_from_vehicle` in the coordinator), and the HA service call currently resets any field the user omits.

## Proposed contract

`None` on an option field means **"leave unchanged"**. Two design notes:

1. **No per-field skip over the wire.** The write endpoints take full state per request, so "unchanged" can only mean "skip" at **endpoint granularity**:
   - Flat EV5 path (ccNC EVs, `ccu_ccs2_protocol_support` + EV): all charge-related fields `None` → skip `POST /ccs2/reservation/charge`; all climate-related fields `None` → skip `POST /ccs2/reservation/hvac`. This makes the no-op HVAC skip in #1298 a consequence of the general rule rather than a special case.
   - Combined `/ccs2/reservation/chargehvac` path (CCSP / EV2–EV4): cannot drop half the payload, so `None` fields still need to be filled from current vehicle state (or documented defaults).
2. **Within an endpoint**, fields left `None` must be filled from current state — either by the consumer, or by a library helper (see below).

## Companion helper

Centralize the "read current settings from vehicle state" logic in the library: a `ScheduleChargingClimateRequestOptions.from_vehicle(vehicle)` classmethod (or equivalent). The kia_uvo coordinator currently reimplements this mapping (`ev_first_departure_*`, `ev_off_peak_*`, `ev_schedule_charge_enabled`, … → options fields) with `or False`/`or [0]` fallbacks that can write guessed defaults when state is unknown.

## Consumers affected

- kia_uvo HA service `schedule_charging_and_climate` (omitted fields currently reset on the car)
- kia_uvo coordinator entity helpers (departure enable, off-peak mode/window, defrost, climate enable)

[...]

PR #1303, opened 2026-09-04
feat: treat None schedule options as leave-unchanged with auto-fill and endpoint skip

### LINKED ISSUE #1312: EV5 flat path – `offpeakPowerFlag` inverted since #1304
**Summary:** On `POST /ccs2/reservation/charge` (ccNC/EV5 flat path), `offpeakPowerFlag` `1` = off-peak **only**, `2` = off-peak **prioritised**. #1304 swapped this for the flat path as well; since then the library writes the opposite of what it reads. Affects 4.31.0.

**Evidence** – Kia PV5 (EU). Mode as reported in `Green.Reservation.OffPeakTime.Mode` (parser: 2 = prioritised, 3 = only; matches the app, "time priority" = 3). Each write also changed the window end, so a readback with the new end proves the write was applied.

| Mode before | flag sent | Mode after | write applied |
|---|---|---|---|
| 3 | 2 | **2** | end 03:20 → 03:40 |
| 2 | 2 | 2 | end 03:40 → 03:50 |
| 2 | 1 | **3** | mode changed |
| 3 | 1 | 3 | end 03:20 → 03:30 |

**Consequence:** a write that should leave the mode unchanged reads Mode 3 as "only", sends `2`, and the car switches to "prioritised" (and vice versa). In kia_uvo, changing only the off-peak start or end time flips the mode.

**Suggested fix:** restore `offpeakPowerFlag: 1 if off_peak_charge_only_enabled else 2` in `_schedule_ev5_flat` and the two EV5 flag tests. The combined `/reservation/chargehvac` path (#1304's reported case, Niro) was not tested here.

blka:
## Summary

Implements #1302 (from @Lekensteyn's review on PR #1298).

`None` fields on `ScheduleChargingClimateRequestOptions` previously coerced to hard defaults before the payload was built — omitting a field silently reset that setting. This PR changes the semantics:

- **`None` = leave unchanged.** An explicit value (including `False`) is written; omitted fields are resolved from current `Vehicle` state instead of hard defaults.
- **New `ScheduleChargingClimateRequestOptions.from_vehicle(vehicle)`** classmethod builds a complete options object from vehicle state, sharing the same fill logic.
- **Unknown vehicle state** (field is `None`) → documented default (`False` / `[9]` / `00:00` / `21.0` / unit `0`) plus a single `_LOGGER.warning` listing the missing fields. Known state is never silently replaced.
- **Flat EV5 (ccNC) path skips inactive endpoints entirely:** a charge scope with no explicit fields sends no `POST /ccs2/reservation/charge`; a climate scope with no explicit fields sends no `POST /ccs2/reservation/hvac`. Both scopes all-`None` → `ValueError` (no schedule changes requested).
- **Combined `/ccs2/reservation/chargehvac` path** (CCSP / EV2–EV4) always sends both scopes — the endpoint cannot drop half the payload — so both are resolved from vehicle state (harmless no-op write when nothing changes).

Departures remain None-or-complete: `first_departure=None` leaves that departure unchanged; a provided `DepartureOptions` gets its `None` fields auto-filled the same way.

## EV5 flat-path `offpeakPowerFlag` enum is not inverted (closes #1312)

Live testing on a Kia PV5 (@fellek, #1312) shows the flat `POST /ccs2/reservation/charge` endpoint uses the **opposite** enum from the combined `/ccs2/reservation/chargehvac` path: `1` = off-peak only, `2` = prioritised. #1304 swapped the flat path along with the combined path, so since 4.31.0 the library writes the opposite mode on EV5-appMode vehicles.

This PR reverts the flat-path half of #1304 (the combined path keeps #1304's corrected enum — its reported Niro case was not contradicted by the live test) and adds warning comments at both call sites so the asymmetry is not "fixed" again.

## Breaking change

`None` no longer means "write the default". Code that relied on the old reset behavior must pass explicit values (or use `from_vehicle` to reproduce the previous read-back-then-write flow). Release tier: **minor** per author decision; the semantics change is documented in the changelog.

One default changed on the combined path: the `off_peak_end_time` fallback is now `00:00` (previously it fell back to the `off_peak_start_time` value); it is normally resolved from vehicle state.

## Merge order / interaction with PR #1298

PR #1298 (skip no-op HVAC) adds a manual vehicle-state check so the hvac endpoint is skipped when nothing changes. This PR makes that behavior a consequence of the general rule: a climate scope with no explicit fields is skipped automatically, and the kia_uvo workaround of rebuilding the full options object from vehicle state becomes unnecessary. If both are merged, #1298's manual skip is redundant and can be dropped in its rebase (this PR supersedes it). Either merge order works.

## Live testing

@fellek ran the PR against a real Kia PV5 (EU, `ccu_ccs2_protocol_support=1` → `_schedule_ev5_flat`): charge-only change → single `POST /ccs2/reservation/charge`, `retCode S`; climate without resolvable temperature → `ValueError` before any request; missing days/time → one combined warning. Details in [the live-test comment](https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1303#issuecomment-5683470423). The two-scope `4004` gap observed there is tracked in #1316 as a follow-up.

## Testing

- `tests/test_schedule_options_fill.py` (new, 12 tests): scope detection, auto-fill from vehicle state, unknown-state default + warning (caplog), `from_vehicle`
- `tests/test_schedule_ev5_flat.py`: 8 new tests for scope skip (charge-only → 1 POST, climate-only → 1 POST, all-None → `ValueError` + 0 POSTs, explicit-off → 2 POSTs), auto-fill; EV5 flag tests assert the flat-path enum (1 = only, 2 = prioritised); 15 existing tests/snapshots unchanged
- Full suite: 725 passed; `ruff` clean; `mypy --strict` no new errors


hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -90,6 +102,211 @@ class DepartureOptions:
...
+            value is not None
+            for value in (
+                options.climate_enabled,
+                options.temperature,
+                options.temperature_unit,
+                options.defrost,
+            )
+        )
+        or options.first_departure is not None
+        or options.second_departure is not None
```
Lekensteyn, line 153: As pointed out here at https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/pull/1298#discussion_r3918146396, turning off the two departure options also disables both climate and charge settings.

I haven't tried enabling charge settings without enabling either departure option since the app doesn't allow for that, but technically I could try it with the API. In my current application, I explicitly check for any departure configuration. If it doesn't exist or if neither are enabled, I add a schedule and enable it to match the app. See https://github.com/Lekensteyn/bluelink-cli/blob/cd76482930b6996b45eade44ba7972afd7006625/bluelink_cli.py#L189-L195

Logically, either departure time only makes sense for climate. And maybe for scheduled charging, without an off-peak only setting enabled. For off-peak only, the departure time is not needed.

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -90,6 +102,211 @@ class DepartureOptions:
...
+    departure.enabled = _fill_option_value(
+        departure.enabled,
+        v_enabled,
+        False,
+        f"{prefix}.enabled",
+        vehicle.id,
+        warn_unknown,
+    )
+    departure.days = _fill_option_value(
+        departure.days, v_days, [0], f"{prefix}.days", vehicle.id, warn_unknown
```
Lekensteyn: At least with a non-CCS2 model, I observed `[9]` meaning "no date", and thereby hiding the on/off switch for the departure option. Using `[0]` will add the departure option for Sunday with an off switch.
```suggestion
        departure.days, v_days, [9], f"{prefix}.days", vehicle.id, warn_unknown
```

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -90,6 +102,211 @@ class DepartureOptions:
...
+                    options.first_departure = departure
+                else:
+                    options.second_departure = departure
+            _fill_departure_options(
+                departure, vehicle, number, warn_unknown=warn_unknown
+            )
+        # One climate set models both departures (known limitation, #1302).
+        options.climate_enabled = _fill_option_value(
+            options.climate_enabled,
+            vehicle.ev_first_departure_climate_enabled,
```
Lekensteyn, line 316: As pointed out in this comment at https://github.com/Lekensteyn/bluelink-cli/blob/cd76482930b6996b45eade44ba7972afd7006625/bluelink_cli.py#L58-L65

> vehicle.ev_first_departure_climate_enabled is EU-only

Please check all other fields related to the info needed for scheduled charging. The fields exist on the `Vehicle` object, but only `KiaUvoApiEU.py` sets it. Do other regions support it?

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -90,6 +102,211 @@ class DepartureOptions:
...
+            False,
+            "climate_enabled",
+            vehicle.id,
+            warn_unknown,
+        )
+        options.temperature = _fill_option_value(
+            options.temperature,
+            vehicle.ev_first_departure_climate_temperature,
```
Lekensteyn, line 323: As pointed out in this comment at https://github.com/Lekensteyn/bluelink-cli/blob/cd76482930b6996b45eade44ba7972afd7006625/bluelink_cli.py#L58-L65, temperature retrieval is not implemented for CCS2. There is an open comment about it in the code, presumably added by you. That needs to be addressed or this will likely fail to keep the temperature setting for CCS2 models.

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -90,6 +102,211 @@ class DepartureOptions:
...
+    vehicle_id: str,
+    warn_unknown: bool,
+) -> _T:
+    """None = leave unchanged: explicit value, else vehicle value, else default."""
+    if current is not None:
+        return current
+    if source is not None:
+        return source
+    if warn_unknown:
+        _LOGGER.warning(
```
Lekensteyn: Let's reduce the amount of noise. Collect any unknown values at the caller, and only print a warning once. Users can't do anything about it, it's probably a programming mistake (like the missing temperature for CCS2 reported below).

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -90,6 +102,211 @@ class DepartureOptions:
...
+        )
+        options.temperature = _fill_option_value(
+            options.temperature,
+            vehicle.ev_first_departure_climate_temperature,
+            21.0,
+            "temperature",
+            vehicle.id,
+            warn_unknown,
+        )
+        if options.temperature_unit is None:
```
Lekensteyn: Can we remove the temperature unit from the `ScheduleChargingClimateRequestOptions` instance and always use the unit from the vehicle object? I assume that this unit depends on the car region, and never changes. If so, it doesn't make sense to burden the user with it?

I see you also added the temperature unit to https://github.com/Hyundai-Kia-Connect/kia_uvo/commit/9735d2620c16233d6d3bebf8fe132485c1ba6809

I think it is okay to break the API, the `_build_schedule_options_from_vehicle` implementation presumably needs changes to take advantage of the new `None` convention. And if we are worried about crashes when kia_uvo users use newer versions, we keep the field on the class for now as no-op. Mark it as deprecated, and remove in a future version.

Lekensteyn (changes requested):
Several comments on improving the API and user experience. Did you actually test this on a real vehicle? Please do.

hyundai_kia_connect_api/ApiImpl.py
blka: I checked the apps' payload behavior before changing anything: the older EU Bluelink app (now decommissioned) does **not** clear `reservFlag` or `reservFatcSet` when the departures are switched off — the round-trip keeps the charge and climate switches independent of `reservChargeSet` (charge on = `reservFlag`, climate on = `airCtrl`, departure on = `reservChargeSet`). Where did you see the auto-disable — in the current myHyundai app's UI, or car-side? I can't reproduce it on my end either (I only have a HEV, so EV schedule writes aren't testable for me).

On the library level, the `None` = leave-unchanged semantics added in this PR should cover your bluelink-cli case: unspecified fields are resolved from the vehicle's current state, so the add-a-schedule workaround is no longer needed to avoid clobbering other settings.

hyundai_kia_connect_api/ApiImpl.py
blka: Applied — the fallback (and an explicit empty day list, which the apps normalize the same way) is now the `[9]` no-days sentinel. Thanks for the confirmation that `[9]` means "no date" — that matches what the app sends for an empty selection.

hyundai_kia_connect_api/ApiImpl.py
blka: Verified: the departure-climate fields (`ev_*_departure_climate_*`) are currently only parsed by `KiaUvoApiEU`; the charge/off-peak fields are parsed by Kia EU, AU, CN and the shared CCS2 status parser. I documented that in `_fill_schedule_options_from_vehicle`. CCS2 departure-climate parsing is an open TODO pending a confirmed response shape; other regions resolve to defaults (now reported with a single warning).

hyundai_kia_connect_api/ApiImpl.py
blka: Addressed: until the CCS2 departure-climate state is parsed (TODO in `ApiImplType1`), a write that activates the climate scope without a resolvable temperature now raises `ValueError` instead of silently sending the 21.0 default and resetting the car's setting. This applies to both the combined and the flat path.

hyundai_kia_connect_api/ApiImpl.py
blka: Done — the fill collects the missing option names and emits a single warning listing them.

hyundai_kia_connect_api/ApiImpl.py
blka: Done — the field stays on the class as a deprecated no-op (kia_uvo still passes it; I'll rework `_build_schedule_options_from_vehicle` on the kia_uvo side separately), and the unit is always derived from the vehicle's reported setting. I agree with the direction: the unit is a car-region property, and the kia_uvo code currently passing it was already broken (it forwarded the unit string, not the wire int) — a good sign it never belonged in the options.

blka:
Separating this from the review replies: before responding, I checked how the official EU apps handle these settings — the older Bluelink Europe app (now decommissioned, but its payload behavior is still the best reference for the SPA v2 shapes these endpoints accept) and the current myHyundai/myKia apps — since I don't have an EV to test the schedule writes live (I only have a HEV).

**Day list.** The older Bluelink Europe app uses day constants 0–6 for Sun–Sat, plus `8 = NEXT_START` and `9 = NOT_SELECTED_DAY`. The UI's 1–7 day selection maps to 0–6, and an empty selection is sent as `day: [9]`. On the way back, any day value outside 0–6 renders as "no days selected" — so `[9]` is exactly "no repeat days" and `[0]` selects Sunday, matching your observation.

**Departure dependency.** The old app enforces, at the UI level: enabling scheduled charging requires a departure time (and off-peak times) to be set first, and enabling scheduled climate requires a departure time. The reverse does not happen in the payload, though: switching the departures off keeps `reservFlag` and `reservFatcSet` as they were — charge on/off is read back from `reservFlag`, climate from `airCtrl`, departure from `reservChargeSet`, all independently. So "turning off both departures also disables both scheduled charging and climate" is not what the old app sends; where did you observe it — which app version (the current myHyundai UI?), and on which vehicle? It may be app-UI state or car-side behavior rather than something the API needs to replicate.

**Current myHyundai app (EV models).** The new app sends the combined charge+hvac object over a different endpoint family than the SPA v2 `/ccs2/reservation/*` used here: `reservFlag` there is a charge-mode enum rather than a plain boolean, the day list can be sent as `null` when no weekly cycle is configured, and the climate set travels inside each departure block (climateSet → 1/0, heating1 always 0, temp as value/unit/hvacTempType). The current myKia app shares the same models. So the two generations differ on at least: `[9]` vs `null` days, and bool vs charge-mode `reservFlag`.

Given that, I kept the library behavior aligned with the payload shapes that the SPA v2 endpoints accept ([9] sentinel, bool-ish `reservFlag`), with the days fallback and normalization changes from the review.

blka:
Honest answer to your earlier question: no, I have not run this against a real vehicle. The car I can test against is an EU Hyundai on the CCI/GSPA flow, which does not go through the legacy `/ccs2/reservation/chargehvac` combined endpoint or the EV5 flat path this PR targets — so these code paths can't be exercised live with what I have. The leave-unchanged/skip logic is covered by the unit tests; if a volunteer with an EV5 (or a CCSP-flow EU vehicle) wants to verify against the real backend, that would close the gap.

fellek:
@blka thanks for your PR. I will test this with KIA PV5 based on KIA PVB.

fellek:
**Live test on a real KIA vehicle** (Kia PV5, EU, `ccu_ccs2_protocol_support=1`, EV → `_schedule_ev5_flat`).

Tested: `master` @ `a2c5485` (4.31.0) + the three commits of this PR, called directly via `VehicleManager` (no HA). Every write was read back with a force refresh (`Green.Reservation.OffPeakTime`).

**Works as intended**
- Charge-only change → only `POST /ccs2/reservation/charge`, `retCode S`, no `/hvac`, no `4004`. On 4.31.0 the no-op `/hvac` always fails with `4004`, so the call raises although `/charge` was applied.
- Climate scope without temperature → `ValueError` before any request.
- Missing departure days/time → one combined warning.

**Rebase note:** `61b1b74` conflicts with #1304 in `_schedule_ev5_flat`. After keeping #1304's line, `TestEV5AutoFill::test_vehicle_state_reaches_charge_payload` fails (expects flag `1` for "only"). That old expectation is actually right for EV5 – see the separate issue on `offpeakPowerFlag`.

**Remaining gap:** when both scopes are sent, `/hvac` follows `/charge` after ~0.2 s and gets `4004` (channel busy). The call raises `DuplicateRequestError` and the `/charge` msgId is lost, although `/charge` was applied. Suggestion: wait for the charge action before sending `/hvac`, or report the partial success.

**Other observations**
- The vehicle floors window times to 10 minutes: sent 02:15 → reads back 02:10 (`/charge` still returns `S`).
- `POST` returns after 0.4–1.9 s (acceptance only). One confirmation observed: `result: success` 6.1 s after the record was created. In 4 of 5 cases the deviceId became invalid 7–11 s after the command (`4002`); after re-registration `notifications/records` is empty (records are per deviceId), and `check_action_status` returns `UNKNOWN` immediately instead of polling until the timeout (related to #1143).
- The `ValueError` message contains the vehicle id.
- 
**My summary:** the approach is right and does what it promises on a real KIA vehicle. Before merging it needs 

- [ ] a rebase on `master` and 
- [ ] the EV5 `offpeakPowerFlag` fix from the separate issue – without it, "None = leave unchanged" still flips the off-peak mode on EV5. 

The two-scope `4004` gap can be a follow-up.

blka:
Thanks for the live test — this is the first real-vehicle validation of both PRs, and it settles the questions I couldn't answer from a HEV.

Both items on your checklist are done in the force-pushed branch (now `5edb30c`, 4 commits):

1. **Rebased on `master`** (`a2c5485`, which includes #1304). The conflict was exactly where you saw it, in `_schedule_ev5_flat`: the rebased code keeps this PR's `if charge_active:` scope-skip structure, but the payload now carries #1304's corrected enum — `"offpeakPowerFlag": 2 if only else 1`.

2. **The failing test is fixed.** After the rebase, `TestEV5AutoFill::test_vehicle_state_reaches_charge_payload` still asserted the old inverted enum (flag `1` for "only" from vehicle state). Flipped to `2`, consistent with the corrected vehicle data model semantics from #1304 and with the pre-existing flat-path tests. Full suite green after the rebase: 725 passed, ruff clean, pre-commit (incl. mypy) passed.

On the two-scope `4004` gap: agreed, that's a follow-up. Sequencing the `/hvac` POST behind the charge action's confirmation (or reporting partial success when `/charge` was applied and `/hvac` hits channel-busy) is a behavioral change worth its own discussion — and your log confirms the current failure mode precisely: the exception loses the `/charge` msgId even though `/charge` was applied.

Your other observations match known EU-side issues, not something this PR changes: the 10-minute flooring of window times is vehicle-side normalization; the `4002` deviceId invalidation 7–11 s after a command is the known device-id rotation root cause, and the immediate `UNKNOWN` from `check_action_status` after re-registration (records being per-deviceId) is #1143 territory. All logged, none introduced here.

blka:
Follow-up to my previous comment: I misread your "that old expectation is actually right for EV5" as meaning the opposite, so the rebase I announced there carried #1304's flat-path enum and flipped the auto-fill test the wrong way. Your #1312 evidence table resolves it: on the flat path `1` = only and `2` = prioritised.

Head is now `eeec12d` (5 commits): `_schedule_ev5_flat` writes the flat-path direction again, the EV5 flag tests assert it, and both call sites have comments documenting that the flat enum is the opposite of the combined path. The combined path keeps #1304's corrected enum. PR body has a dedicated section and `Closes #1312`.

The two-scope `4004` gap remains tracked separately in #1316.

fellek:
@blka thanks for your commits, I can confirm on live test: the last commit solved the issue.

blka:
@fellek That’s great news; I’m glad another issue has been fixed compared to the original app and its behaviour. @cdnninja, as the scripts from bump kia_uvo don’t display the exact information, please add this change manually in Sunday’s edition. 
P.S. I’m still waiting for the review of GSPA Control PR 😁

fellek:
That's great, I will look into it. I just started to analyse the driving-insight data endpoints, which is not documented until now. that might be an interesting modul for this api.

blka:
That direction is welcome — you're not starting from zero: the library already carries two generations of these endpoints, and the third layer is mapped from traffic captures for Hyundai:

1. **Legacy SPA (implemented, `KiaUvoApiEU`):** `POST /drvhistory` with `{"periodTarget": 1|0}` — alltime summary (`resMsg.drivingInfo`: `totalPwrCsp`, `regenPwr`, `calculativeOdo`) and 30-day per-day stats (`resMsg.drivingInfoDetail`: `drivingDate`, `totalPwrCsp`, `motorPwrCsp`, `climatePwrCsp`, `eDPwrCsp`, `batteryMgPwrCsp`, `regenPwr`). Known server-side flakiness with empty responses is tracked in #1315.

2. **GSPA flow (implemented for Hyundai EU, `HyundaiCciApiEU`):** `GET /gspa/v1/driving-info/vehicles/{carId}` and `GET /gspa/v1/driving-history/vehicles/{carId}` (same `drivingInfo`/`drivingInfoDetail` response family as the legacy endpoint).

3. **Current "driving insights" layer (mapped for Hyundai EU from traffic captures):** `GET /domain/api/v1/driving-insight/odometer?ccId={deviceId}_BLU&vin={VIN}` on the CCI domain API — the same endpoint the official app uses for the odometer tile — plus a separate DI API behind `POST /domain/api/v1/auth/token-exchange?serviceType=DI` — `/v0100/car/trip/driving-info`, `/v0100/drive-history/trip[/days|/detail]`, `/v0100/trip-tags[/trips/{tripId}]` — per-trip data with tags, well beyond anything the library exposes today.

The open gap is brand parity: 1–2 are verified on the Hyundai side; the Kia-brand equivalents (`cci-api-eu.kia.com` / `gspa-ccs-eu.kia.com` / the `di-kia-eu` paths) are unconfirmed. Your PV5 sits exactly there — if your analysis turns up the endpoint paths and response shapes a Kia on the CCI flow gets, I can compare them 1:1 against the Hyundai shapes and extend `KiaCciApiEU`. Even a bare list of paths would settle the biggest question.

blka:
@Lekensteyn gentle nudge on this one: all six threads from your review are addressed — the two remaining answered threads are now resolved, and the branch has moved since your review passed (`eeec12d`, rebased on `a2c5485` which includes #1304). Since then the branch also got its first real-vehicle validation: fellek's live test on a Kia PV5 (EV, `_schedule_ev5_flat`) confirmed both the leave-unchanged scope-skip and the corrected `offpeakPowerFlag` semantics (details in the comments above; the enum correction is tracked in #1312, which this PR closes). Could you take another look?

Issue #1145, opened 2026-05-23
oneapp inspection - new auth and endpoints


Mirarkitty:
I played with Claude a bit to inspect the new app, the auth system, and the new endpoints. There's some interesting finds, but someone that knows better than me how to do it need to get the client id out of the system. 

"pet-care" was an interesting endpoint. Haven't seen that in the app yet. Ithink the analysis what we can get from kia-uvo is pessimistic, but there's some really nice new potential stuff like **battery conditioning**.

I don't know if this github repo is the right port of call for this, but it's where I would look for this, so...

# Kia OneApp EU 

Claude did a full static RE of the OneApp EU APK (jadx, no live traffic).
Auth flow, endpoint catalog, command surface are mapped. One hard blocker remains
(X-Stamp computed in native code).

| Host | Purpose | Stamp? |
|---|---|---|
| `idpconnect-eu.kia.com` | OAuth2/PKCE login — **shared with the captcha `peukiaidm-online-sales` flow** | No |
| `cci-api-eu.kia.com` | User profile, agreements, notifications | No |
| `egw-svchub-ccs-k-eu.eu-central.hmgmobility.com` | Vehicle metadata, MQTT host discovery, connection state | No |
| `oneapp.kia.com` (Ktor) | App initialization, charging-station search, loyalty | No |
| `gspa-ccs-eu.kia.com` | **All vehicle telemetry + all remote commands** | **Yes** |

Gating logic verified in `com.hmg.ccs.sdk.data.common.header.RequestHeader.a`:
Stamp is added iff `authority.contains("gspa")`. The Bearer token is unified
across all five hosts.

**blocked at:**

1. Production OneApp inner `client_id` (UUID) — xShield-encrypted in the APK
2. Outer login facade slug (likely `peukiaidm-<role>` — brute-discoverable against the IDP)
3. `libgspa-cipher.so` Stamp algorithm — needed only for the GSPA host
4. Production base URLs — xShield-encrypted, recoverable by several routes

# Feature Matrix

| Capability | kia_uvo | OneApp CCS2 | Notes |
|-----------|---------|-------------|-------|
| Vehicle status (battery, climate, doors) | Yes | Yes  | `stored-status` needs Stamp |
| Fresh GPS / location | Yes (workaround) | Yes  | `/location/vehicles/{carId}/stored-status` needs Stamp |
[...]

PR #1305, opened 2026-09-05
feat(eu): GSPA remote-control commands for Hyundai CCI (climate, windows, charge, valet) + live-probed envelope fixes

### LINKED ISSUE #1317: start_climate silently no-ops on Kia PV5 (EU, OneApp/CCI vehicle) — cloud acks success, car never starts climate
# Bug report draft — start_climate silently no-ops on Kia PV5 (EU, OneApp/CCI vehicle)

**Repo:** Hyundai-Kia-Connect / hyundai_kia_connect_api (used by HA custom integration `kia_uvo`)

## Summary

`start_climate()` on a **Kia PV5 Cargo (2026 PBV model, EU/Denmark)** never actually starts the vehicle's climate, despite the cloud reporting success. The vehicle speaks Kia's **new OneApp / CCI protocol** (`cci-api-eu.kia.com`, OneApp client `01b36c86`, package `com.kia.oneapp.eu`) — not the legacy SPA/ccs2 endpoint the current `start_climate` implementation targets.

Lock/unlock and stop_climate work reliably through the legacy path. start_climate "succeeds" (HTTP 200, resCode `0000`, retCode `S`) yet the car's climate ECU never receives the command — the official Kia OneApp DOES start climate on the same vehicle at the same time (verified multiple times, including at night with no sun load).

## Evidence

1. **First call of each coordinator cycle fails with DeviceIDError:**
   ```
   hyundai_kia_connect_api.exceptions.DeviceIDError: Invalid request body - Invalid deviceId.
     File ".../hyundai_kia_connect_api/ApiImplType1.py", line 1399, in start_climate
       _check_response_for_errors(response)
   ```
   The integration's `@_retry_on_device_id_error` decorator refreshes the device ID and retries; the retry returns `retCode S / resCode 0000` (accepted). **The cached deviceId going stale every cycle on this vehicle may itself be a symptom of the CCI vs SPA mismatch.**

2. **Success is fake:** resCode `0000` returned, climate never starts. Reproduced ~6+ times over two days (scheduled runs + manual), all vehicle states (parked, locked, charging), multiple payloads:
   - default null payload → no-op
   - all-fields payload (frseat/rlseat/rrseat/steering_wheel explicitly `0`) → no-op (rules out the "0 = unset" null hypothesis)
   - target temp 21°C and 27°C, duration 15–20 min → no-op
   - healthy latency (~70–78 s) → not a timeout issue

3. **CCI remote control is explicitly unimplemented in this library:** `KiaCciApiEU` raises `NotImplementedError` for prewakeup/remote actions ("awaits live verification"). The vehicle apparently routes ALL remote commands through CCI (OneApp), but the legacy SPA endpoint still acks the old-style climate call — a silent divergence between cloud and vehicle.

4. **Controlled comparison (sep 17 ~14:10):** Kia OneApp start_climate → cabin heats within minutes (confirmed, no sun). Integration start_climate seconds earlier → nothing. Same vehicle, same conditions.

## Vehicle

- Kia PV5 (PBV), model year 2026, EU market (Denmark), OneApp account (migrated from UVO)
- Works in official Kia OneApp: climate, lock/unlock, charge control
- Via library: lock/unlock ✅, stop_climate ✅, start_climate ❌ (silent no-op), forced refresh occasionally DeviceIDError

## Expected behavior

start_climate should either (a) reach the vehicle via the CCI protocol it actually speaks, or (b) fail loudly — not return resCode 0000 while doing nothing.

## Suggested direction

Implement CCI remote-control endpoints (`cci-api-eu.kia.com`) for OneApp-protocol vehicles, or detect CCI vehicles and surface an explicit "unsupported protocol" error instead of the silent fake success. Happy to provide sanitized debug logs of a no-op attempt.

blka:
## Summary

Adds the GSPA remote-control layer for the EU Hyundai CCI API, on top of the shared `GspaApiEU` base introduced in #1284:

- **Control dispatcher** on `HyundaiCciApiEU` (`_control_command`) routing CCS2-protocol vehicles to GSPA commands; pre-CCS2 vehicles raise `UnsupportedControlError` (Kia CCI door-split methods behind a `NotImplementedError` gate until Kia-side support is verified).
- **GSPA control commands** for Hyundai: climate/engine start/stop (extended `ClimateRequestOptions`: side mirror heating, temp unit, HVAC temp type, driver seat location), windows + per-seat curtains (`WindowRequestOptions`), charge settings (targets, off-peak windows), reservations, valet mode, pet care, horn-light.
- **PIN control-token flow** with a TTL cache (600 s — `expiresTime` is a TTL in seconds, not an epoch; verified live), Bearer-vs-CCSP header dispatch per endpoint (`GSPA_BEARER_ENDPOINTS`), CCI auth headers on PIN verification.
- **Action status polling**: control commands return `gspa:{SID}` action ids; `check_action_status` polls the update-status endpoint and maps `pollingState` WAIT/SUCCESS/FAILURE/TIMEOUT to `ORDER_STATUS`.
- **Typed error classification** from GSPA `metaInfo` envelopes and Spring error bodies: 401 → `AuthenticationError`, 400-004 → `DuplicateRequestError`, 404 → `UnsupportedControlError`, 5xx → `ServiceTemporaryUnavailable`.

Fixes #1317 — legacy SPA climate control acks at the cloud but silently no-ops on EU CCI/OneApp vehicles; the vehicle only honors GSPA control (live-verified on two PV5s).

## Live-probed fixes folded in

Verified against a real EU vehicle with an active remote-control entitlement:

- Success envelope is HTTP 202 with `{"data": {...}, "metaInfo": {"retCode": "S", "resCode": "202-000"}}` and the SID in `data` (`SID`/`svcSID` fallbacks) — the legacy `{rc, rs}` parsing missed this entirely on a real success.
- Some commands are accepted (202-000) with an **empty `data` object and no SID**; that path now returns the bare `gspa:` prefix instead of raising `InvalidAPIResponseError` for a command the server accepted.

## Tests

+1031 lines of tests: envelope/infra (`test_gspa_control_infra.py`), command builders (`test_gspa_control_commands.py`), Kia `NotImplementedError` gate. 721 tests pass on py3.14; ruff 0.16.4 clean (check + format).


hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -851,3 +864,637 @@ def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> No
                     self._update_vehicle_driving_history(vehicle, history)
             except Exception:
                 _LOGGER.debug(f"{DOMAIN} - Driving history fetch failed")
+
+    # ------------------------------------------------------------------
+    # Remote control (GSPA) — dispatcher
+    # ------------------------------------------------------------------
+
+    def _control_command(
```
cdnninja, line 872: Couldn't this be part of self._gspa_control_command?   It feels like alot of lines just to check support when it is universal.  I assume the same will apply on kia.

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -851,3 +864,637 @@ def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> No
...
+        if vehicle.ccu_ccs2_protocol_support:
+            return self._gspa_control_command(
+                token, vehicle, endpoint, body, path_prefix=path_prefix
+            )
+        raise UnsupportedControlError(
+            "Pre-CCS2 EU vehicles are not supported by the CCI region — "
+            "use region 1 (Europe) for remote control"
+        )
+
+    def check_action_status(
```
cdnninja, line 895: Couldn't this be in the parent class?  Nothing specific by the looks of it.

hyundai_kia_connect_api/HyundaiCciApiEU.py
```diff
@@ -851,3 +864,637 @@ def update_vehicle_with_cached_state(self, token: Token, vehicle: Vehicle) -> No
...
+        raise UnsupportedControlError(
+            f"Cannot poll action {action_id!r}: the CCI region only issues "
+            "'gspa:' action ids"
+        )
+
+    # ------------------------------------------------------------------
+    # Remote control (GSPA) — simple commands
+    # ------------------------------------------------------------------
+
+    def lock_action(
```
cdnninja, line 921: All these seem like they can be parent class methods.

cdnninja (commented):
Feels like we are setting up for a huge amount of duplicate code.  Couldn't almost all of this move to the europe parent file?  I honestly wonder if we even need model specific files.   I suspect we don't.

hyundai_kia_connect_api/HyundaiCciApiEU.py
blka: Done in 7e1ba70 — the CCS2 gate now lives at the top of `_gspa_control_command` and the wrappers call it directly. Your assumption holds for Kia too: it inherits the whole layer from GspaApiEU, gated by a `GSPA_REMOTE_CONTROL_VERIFIED` class flag (False for Kia → NotImplementedError before anything is sent) until it's verified live on a real Kia vehicle.

hyundai_kia_connect_api/HyundaiCciApiEU.py
blka: Moved to GspaApiEU in 7e1ba70 — nothing in it was brand-specific; the "gspa:" dispatch now sits next to `_gspa_check_action_status` on the parent.

hyundai_kia_connect_api/HyundaiCciApiEU.py
blka: Moved in 7e1ba70 — all of them (plus the schedule/reservation builders) live on GspaApiEU now. On model-specific files: the brand classes keep only their constants and read parsers (CCS2 vehicle-property parsing, driving info/history, breakdowns) — those are response-shape specific, and Kia's own parsers can't be written without a live response from a real Kia vehicle. If the Kia read shape turns out to match (same OneApp generation), KiaCciApiEU shrinks to constants and the file question becomes moot — that waits on a live Kia capture.

blka:
https://github.com/Hyundai-Kia-Connect/kia_uvo/discussions/1764#discussioncomment-18309201

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -52,13 +58,29 @@ class WindowRequestOptions:
     back_right: WINDOW_STATE = None
     front_left: WINDOW_STATE = None
     front_right: WINDOW_STATE = None
+    # Per-seat fields (GSPA window-curtain endpoint).
+    driver_seat_window: WINDOW_STATE | None = None
+    passenger_seat_window: WINDOW_STATE | None = None
+    rear_left_window: WINDOW_STATE | None = None
+    rear_right_window: WINDOW_STATE | None = None
```
cdnninja, line 66: How does this differ from back_right for window state?

hyundai_kia_connect_api/GspaApiEU.py
```diff
@@ -118,6 +181,10 @@ def __init__(self, region: int, brand: int, language: str) -> None:
         else:
             raise APIError(f"Unknown cipher brand: {self.CIPHER_BRAND}")
 
+        # PIN-derived control token cache (D5: per API instance, not Token).
+        self._control_token: str | None = None
```
cdnninja, line 292: What does this mean per API instance?    Can't we only have one to one mapping?

hyundai_kia_connect_api/GspaApiEU.py
```diff
@@ -780,6 +847,345 @@ def _validate_ccs_token(self, token: Token) -> None:
...
+                    f"GSPA auth/stamp: {spring_code} {spring_msg}"
+                )
+            if spring_code >= 500:
+                raise ServiceTemporaryUnavailable(
+                    f"GSPA transient: {spring_code} {spring_msg}"
+                )
+            raise APIError(f"GSPA error: rc={spring_code}, msg={spring_msg}")
+        res_code = meta.get("resCode") or data.get("rc")
+        msg = meta.get("message") or data.get("msg", "")
+        if res_code in ("400-004", "4004"):
```
cdnninja: Could we follow this approach?  This is alot of if statements and doesn't scale well. 

https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api/blob/7eaed03490807e00aa5307184f6f6689f3ae98fa/hyundai_kia_connect_api/KiaUvoApiCA.py#L188-L194

I believe we are trying to follow that approach in all areas if possible.

cdnninja (changes requested):
Initial thoughts.   I haven't made it through the whole thing.

hyundai_kia_connect_api/ApiImpl.py
blka: The existing `front_left`/`back_right` fields are position-based and map to the CCS2 window endpoint, where front-left is the same physical window in every market. The new fields map to the GSPA window-curtain endpoint, whose body keys are seat-based (`drvSeatWindow`, `psgSeatWindow`, `rlSeatWindow`, `rrSeatWindow`) — the driver's window is the left one on LHD but the right one on RHD, so merging the two models would silently drive the wrong window on RHD vehicles. The GSPA body also carries rear curtains and the driver-seat location (`drvSeatLoc`), which have no position-based equivalent.

`set_windows_state` accepts only scope commands (all / front); per-seat windows, curtains, and the seat location go through `set_window_curtain`.

hyundai_kia_connect_api/GspaApiEU.py
blka: Fixed in 545107c: the cache now records which Token it was derived from (`_control_token_source`, keyed by the CCI credential) and is only reused when the same source token is presented — a new Token (e.g. after a re-login) forces a fresh PIN verification instead of reusing the old control token. Covered by `test_control_token_cache_is_per_source_token`.

hyundai_kia_connect_api/GspaApiEU.py
blka: Converted in 545107c to a mapping dict in the KiaUvoApiCA `error_code_mapping` style (`GSPA_RES_CODE_MAPPING`: `400-004`/`4004` → DuplicateRequestError, `403`\* → AuthenticationError, `404`\* → UnsupportedControlError; exact code or 3-char prefix lookup).

Three branches can't be a dict and stay explicit:
- the `>= 500` range check (transient),
- the live-probed business state where the OTA check answers `404-007` with "No update info found by vin" — that must raise APIError ("nothing pending"), not UnsupportedControlError, so the message check takes precedence over the resCode mapping,
- the generic fallback carrying the raw server message.

hyundai_kia_connect_api/ApiImpl.py
cdnninja: Not sure I understand.  So Rear_right_window and back_right aren't both the back right windows?

hyundai_kia_connect_api/GspaApiEU.py
cdnninja: What I mean is shouldn't we be using it from the token?

blka: Done in 749756f — the control token now lives on the Token object (`control_token` / `control_token_expiry`), the same way ApiImplType1 caches the CCS2 control token for the Type1 regions. The instance-level cache and its per-source guard are gone: a fresh Token after a re-login trivially has no control token, so the one-to-one mapping is structural. Invalidation clears the Token fields (the 401-retry behavior is unchanged), and since the fields ride Token.to_dict/from_dict, a persisted Token keeps a still-valid control token across restarts.

hyundai_kia_connect_api/ApiImpl.py
blka: Yes — for the rear windows they are the same physical window: `rear_right_window` maps to the GSPA body key `rrSeatWindow`, `back_right` maps to the legacy CCS2 SPA v2 body key `backRight` (ApiImplType1.set_windows_state). The two field groups exist because the endpoints take different body schemas, and they only genuinely diverge on the front windows: CCS2 is position-based (`frontLeft`/`frontRight`), GSPA is seat-based (`drvSeatWindow`/`psgSeatWindow`) where the physical side depends on `drvSeatLoc` (LHD/RHD). The library can't derive that itself — `drvSeatLoc` is caller-supplied — so a position-based front model would still require it; the seat fields mirror the GSPA body one-to-one instead. I've made that explicit in the field docs (749756f).

blka:
New commit pushed (`ffcb396` → `a36402b`) while the review is pending:

The CCS2 property parser moved from `HyundaiCciApiEU` to the shared `GspaApiEU` base (pure relocation — no behavior change for Hyundai), and `KiaCciApiEU.update_vehicle_with_cached_state` now parses the GSPA stored-status response through it. Kia stays stored-status only: driving-info and driving-history fail server-side on Kia (confirmed live on an EV6, see Hyundai-Kia-Connect/kia_uvo#1764), so a cached-state update deliberately skips those two requests instead of throwing on every poll. Fixture + tests added; full suite 778 passed. Remote-control methods remain the gated stubs pending a live control capture.

cdnninja:
Being thousands of lines I will need to find time to sit down at a computer to review this more.  

In general it's good practice to do small PRs rather than huge ones.  HA for example wouldn't accept a PR of this size.

fellek:
## Test report: #1305 on a Kia PV5 (EU CCI)

Tested at `a36402b` on a Kia PV5 Passenger (2026, EV, CCS2 protocol 2, region EU CCI), 19 Sep 2026. Everything below ran against the PR code unchanged; the Kia control gate was lifted on the instance only, inside a test script.

### Summary

| Area | Result |
|---|---|
| Test suite (offline) | 778 passed, 15 snapshots passed |
| Login, vehicle list, `stored-status` with PR code | works, `S` / `200-000` — the `X-Request-Id` fix (`ffcb396`) is confirmed on Kia |
| Control transport (PIN token, envelope, SID) | works on Kia unchanged |
| Door lock via inherited `lock_action` | **works on Kia** (see below) |
| Kia cached-state parser (`a36402b`) | loses 11 values and sets 3 wrong on the PV5 (see below) |

### Door lock with PR code

One command per run, gate lifted on the instance only.

| Call | Request | Response | Vehicle |
|---|---|---|---|
| `lock_action(LOCK)` | `POST remote/vehicles/{carId}/door` `{"command": "close"}` | `202` `S` `202-000` | doors open → locked, hazard lights flash, new status 7 s after sending |
| `_gspa_control_command("door-lock", …)` | `{"command": "set"}` | `202` `S` `202-000` | hazard lights flash |
| `_gspa_control_command("door-lock", …)` | `{"command": "lock"}` (body from the Kia comment in `KiaCciApiEU.py`) | `400` `F` `400-002 Invalid request body` | – |
| `_gspa_control_command("door", …)` | `{"command": "CLOSE"}` (upper case, as listed in #1145) | `400` `F` `400-002 Invalid request body` | – |

Findings:

1. The inherited Hyundai-style `lock_action` (`door` + `"close"`) locks a Kia. Kia does not need the gated `lock_door` / `lock_door_safety` overrides or a separate endpoint for locking; lifting the gate for `lock_action` would be enough. Unlock (`"open"`) was not tested.
2. The Kia door comment in `KiaCciApiEU.py` names the wrong body. `door-lock` expects `{"command": "set"}`; the action is in the endpoint. `"lock"` is rejected.
3. Values are case-sensitive: `"close"` works, `"CLOSE"` is rejected.
4. SID shape differs by endpoint: `door` returns `data.SID` equal to `metaInfo.msgId`, `door-lock` returns a UUID. The PR handles both (`gspa:<SID>`).
5. Not tested with PR code: `check_action_status`. In earlier runs on this vehicle, `GET status/vehicles/{carId}/update-status` did not act as a result poll on Kia but triggered a new status upload itself (new SID per call, `400-004` in between). A fresh `stored-status.lastUpdateTime` was the reliable success signal.

### Cached-state parser

Same `stored-status` payload, parsed with the PR (`GspaApiEU._update_vehicle_properties_ccs2`, moved from `HyundaiCciApiEU`) and with the CCS2 parser of `ApiImplType1` (used as a mixin in my branch): 110 vs. 121 attributes set, 18 differences. The live payload of the PV5 has the same structure.

Values lost in the PR, because it reads paths the vehicle does not send:

| Attribute(s) | PR reads | PV5 sends |
|---|---|---|
| `ev_battery_water_temperature` | `Green.BatteryManagement.Temperature.Water` | `…Temperature.CoolingWaterInlet` |
| `ev_battery_pack_voltage` | `…BatteryPack.Voltage` | `…BatteryPackVoltage` |
| `ev_battery_chiller_rpm` | `…Chiller.RPM` | `…ChillerRPM` |
| `ev_off_peak_start_time`, `ev_off_peak_end_time`, `ev_off_peak_charge_only_enabled` | `Green.Reservation.OffPeakPower.*` | `Green.Reservation.OffPeakTime` (hour/min/mode) |
| `ev_schedule_charge_enabled` | `Green.Reservation.ChargeSchedule.Enable` | `OffPeakTime.Mode` |
| `ev_power_consumption_air_conditioning`, `_battery_cooling`, `_battery_heater` | `Green.EnergyConsumption.*.Value` | `Green.PowerConsumption.Moment.*` |
| `headlamp_status` | not read | `Body.Lights.Front.HeadLamp.SystemWarning` |

Wrong values in the PR:

| Attribute | PR | Correct | Cause |
|---|---|---|---|
| `ev_battery_precondition_enabled` | `True` | `False` | `bool(BatteryPreCondition)` tests the object `{Status, TemperatureLevel}`, which is always truthy; `Status` is `0`. The app maps `3`/`4` = on (kia_uvo #1823). |
| `ev_battery_winter_mode` | `True` | unset | set from `WinterModeOperation`; on EVs this is not a user toggle (kia_uvo #1823) |
| battery temperature unit | `"C"` | `"°C"` | literal instead of `TEMPERATURE_UNITS[0]` |

One value only the PR sets: `transmission_condition` from `Drivetrain.Transmission.ParkingPosition`.

### Suggestions

1. Let Kia (and probably Hyundai EU CCI, not verified here) use the `ApiImplType1` CCS2 parser instead of the moved copy, or fix the paths and the two mappings above.
2. Lift the Kia gate for `lock_action`; keep unlock gated until tested.
3. Fix the Kia door comment: `door-lock` takes `{"command": "set"}` (verified). In the Kia app 1.4.0, `door-unlock` uses a request class with the same single field and the  same default `"set"`; not tested live.

Raw results, scripts and dumps are available if useful.

edit: typo

blka:
Thanks for the thorough test run — this was exactly the evidence we needed. Your table is now in the PR (ac523e6):

1. **CCS2 parser paths fixed to the flat schema** (BatteryPackVoltage, ChillerRPM, CoolingWaterInlet, PowerConsumption.Moment.*); the nested paths no captured payload ever contained (BatteryPack.Voltage, Chiller.RPM, Temperature.Water, EnergyConsumption.*.Value) are gone. Also in: the OffPeakTime block (mode 0/2/3, no synthesised midnight window when absent), BatteryPreCondition.Status 3/4 = on, and headlamp from HeadLamp.SystemWarning.
2. **Kia door stubs** now document the per-action endpoints with `{"command": "set"}` bodies, case-sensitive.
3. **The gate is now per-endpoint**: `lock_action(LOCK)` goes through `door` + `close`, which you proved live; UNLOCK and the safety variants stay NotImplementedError until exercised on a real car. Your 1.4.0 observation lines up with the per-action `{"command": "set"}` protocol we ship.

Two asks when you have time:

1. **A raw stored-status JSON** from your PV5 (response body as-is, values anonymized) as a parser fixture — the flat paths above are built from your table plus a Sportage CCS2 fixture and a live EV6 capture, but a real PV5 payload would be the ground truth.
2. **The same capture script as the EV6 run**: https://gist.github.com/blka/11e3e50887169a3dcec0f3b820408878 — with `--control` it walks door lock/unlock, the safety variants, prewakeup, lamp, charge-target, climate and charge, and reads the status back after each step. A PV5 pass would cover the second half of the dispatch table; once those land, the remaining NotImplementedError gates can come off.

And a third, smaller one: in Hyundai-Kia-Connect/hyundai_kia_connect_api#1321, the fields without samples (V2L `VehicleToLoad.Mode`, `TurnSignal.LampState`, `Departure.Climate.{Activation,...}`, `SteeringWheel.Heat.RemoteControl.Step`) would map onto existing model fields — could you note the observed values while toggling each in the app? That would unblock a parser follow-up right after this merges.

Once the push lands, retesting the lock flow through your HA instance would be a great end-to-end check.

blka:
Script update — v8 is live (same link, https://gist.github.com/blka/11e3e50887169a3dcec0f3b820408878). Please use the current version if you grabbed the script earlier today:

- the per-action door endpoints now send `{"command": "set"}` — on Kia the action lives in the endpoint, and a body of `"lock"`/`"unlock"` (what the previous version sent) is rejected with 400-002;
- prewakeup sends `{"action": "prewakeup"}`;
- the confirmation step reads `stored-status.lastUpdateTime` instead of polling `update-status` — on Kia each `update-status` call triggers a fresh upload, so polling it works against the backend.

The rest of the `--control` flow (walk order, redaction, read-back after each step) is unchanged.

fellek:
Follow-up to my test report above: **unlock** is verified on the PV5 as well, still at `a36402b`, gate lifted on the instance only.

| Call | Request | Response | Vehicle |
|---|---|---|---|
| `_gspa_control_command("door-unlock", …)` | `{"command": "set"}` | `202` `S` `202-000` | unlocked |
| `lock_action(UNLOCK)` | `POST remote/vehicles/{carId}/door` `{"command": "open"}` | `202` `S` `202-000` | locked → unlocked, new status 4 s after sending |

So on Kia the inherited `lock_action` covers both directions; the gated `lock_door` / `unlock_door` overrides are not needed. The PV5 relocks by itself after 30 s if no door is opened. The `-safety` variants are still untested.

**Correction to finding 4 of the report:** the SID shape does not depend on the endpoint. When locking, `door-lock` returned a UUID and `door` the short `msgId` form; when unlocking it was the other way round. The PR handles both.

blka:
Unlock verified — the gate is lifted. As of 061e4ae, `lock_action` covers both directions on Kia through the shared `door` endpoint (`close`/`open`), and the gated `lock_door`/`unlock_door` stubs are gone entirely — per your point, Kia needs no separate endpoint for locking or unlocking. The `-safety` variants stay `NotImplementedError` until tested live.

The SID note is handled — the PR already accepts both the short `msgId` form and the UUID, so no change needed there. The ~30 s self-relock is now documented in the code comment.

Also: master moved today (#1318, #1303), so the branch is rebased onto it. If you retest the lock/unlock flow through your HA instance, the gate is now open in the PR code itself — no instance-side flag needed — which would be the end-to-end check.

blka:
Follow-up here so it doesn't get lost — posted a fuller note on the #1145 thread (comment 5744804615), the one open item relevant to this PR's dispatch table is the `charge-target` plugType enum. We send `0 = DC, 1 = AC` on the legacy SPA endpoint and read the same mapping from `targetSOClist`, but the tables in #1145 say `1 = AC, 2 = DC` — possibly a different enum on the GSPA path, unverified on Kia.

Your capture script already includes the charge-target step; if a `--control` pass happens on the PV5 anyway, the raw request body would settle the dispute before charge limits get enabled for Kia.

fellek:
Climate test on the Kia PV5 (for #1317), at `34ea5ca`, gate lifted on the instance only
(`GSPA_VERIFIED_ENDPOINTS = {"door", "temperature"}`). All calls through `start_climate` /
`stop_climate` of the PR unless noted. The climate code is unchanged in `061e4ae`, so the
findings below still apply after the rebase.

**1. `temperature` works on Kia** — start and stop, `202` `S` `202-000`, the car reacts within
5 s (about 21 s when it was asleep). Candidate for `GSPA_VERIFIED_ENDPOINTS`.

**2. Bug: `tempUnit` must be a string.**

| `ClimateRequestOptions` | sent | result |
|---|---|---|
| `temp_unit="C"`, `hvac_temp_type=1`, `driver_seat_location="L"`, `set_temp=22.0` | `"tempUnit": "C"` | `202`, set point 22 applied |
| same with `temp_unit=0` | `"tempUnit": 0` | `400-002 Invalid request body` |

`temp_unit` is annotated `int` (0 = Celsius) and passed through unchanged, and
`test_start_climate_full_options` asserts exactly the rejected form. The PR already sends
`"C"`/`"F"` for `pet-care` and the departure schedules; mapping 0 → `"C"`, 1 → `"F"` in
`start_climate` would fix it.

**3. Without `tempUnit`/`hvacTempType` the set point is ignored.** `set_temp=21.0,
duration=10` only → `{"command": "start", "hvacTemp": "21.0", "ignitionDuration": 10}`: the
climate runs, but the status shows the car's stored set point (20), not 21. With the full
form the value is applied (22, and 20 on a later run). kia_uvo does not pass `temp_unit` or
`hvac_temp_type`, so from HA the temperature setting would have no effect. Suggestion: when
`set_temp` is given, default `tempUnit` to `"C"` and `hvacTempType` to 1 (the PV5 reports
`Server.Option.HvacTempType = 1`).

**4. The set point of a remote start stays in the car** (next drive started at the last
remote value). Worth a note in the docs; fan speed / AUTO are not part of the request.

**5. Extra fields** (one per run, full form, 20 °C, 21.5–25.5 °C outside):

| field | result | status |
|---|---|---|
| `strgWhlHeating: 1` | `202` | `Cabin.SteeringWheel.Heat.State` and `Heat.RemoteControl.Step` → 1 |
| `windshieldFrontDefogState: true` | `202` | `Body.Windshield.Front.Defog.State` → 1, blower level 9 |
| `seatClimateInfo: {"drvSeatClimateState": 6}` | `202` | no seat state on the PV5 |
| `heating1: 2` | `202` | `Body.Windshield.Rear.Defog.State` stays 0 |
| `windshieldFrontHeatState: 1` (sent directly, **not in the PR**) | `202` | `Body.Windshield.Front.Heat.State` stays 0 |

`windshieldFrontHeatState` (and `airPurifierControl`) are fields of the app's request class
that `ClimateRequestOptions` does not cover yet. The two "stays 0" rows may just be the
warm weather; I will retest in the cold.

**6. Running state for the parser:** `Cabin.HVAC.Row1.Driver.Temperature.Value` (set point
instead of `"OFF"`) and `Blower.SpeedLevel > 0` are reliable.
`Green.PowerConsumption.Moment.ClimateAirConditioning` sometimes keeps its last value after
a stop (1260 W with `"OFF"` and blower 0), so it should not drive `air_control_is_on`.

**7. Small things:** `stop` is answered with `202` even when the climate already ended after
`ignitionDuration`; the PIN control token was valid for about 10–15 min; while the climate
runs the car pushes a fresh stored-status about every minute.

fellek:
> If you retest the lock/unlock flow through your HA instance,

Happy to do the end-to-end check on the PV5. Could you please answer the following Question?

How would you like me to get there? The library has had region 9 since #1284 (`369622d`,
4 Sep) and ships it in the current release (`REGION_EUROPE_CCI = "Europe (CCI/GSPA)"` in
`const.py`, v4.32.0). In `kia_uvo` 3.13.0, though, the `REGIONS` dict in its own `const.py`
ends at `8: Brazil`, and the config flow builds its picker from that dict — so region 9
(the one `VehicleManager` maps to `KiaCciApiEU`) cannot be selected.
Picking "Europe" there gives region 1 and `KiaUvoApiEU`, which is exactly the legacy path
that acks but never reaches the car on a PV5.

Is a region entry planned on the integration side, or would you rather I patch my instance
for the test — adding `9` to `REGIONS` and pointing the `manifest.json` requirement at this
branch?

If a local patch is the way, is there anything specific you want captured beyond
lock/unlock reaching the car: the debug log of one coordinator cycle, the entity states
after the command, or the behaviour of `check_action_status`?

blka:
Short answer: no, there is no kia_uvo PR for region 9 yet. PART B (#1892) only adds the SVM image entities — it does not touch REGIONS. So for an end-to-end lock/unlock test today, a local patch is indeed the way.

The library side is ready on this PR (head `57c891a`): lock/unlock (`door` + `{"command": "close"/"open"}`) and climate start/stop (`temperature`) both pass the verified-endpoint gate on Kia, with the tempUnit fix included — the API now sends `tempUnit: "C"/"F"` strings and defaults `hvacTempType` to 1 whenever a set point is given. So a patched kia_uvo instance pointing at this branch can exercise lock/unlock and HVAC directly.

For the kia_uvo side, the patch is small: add the region-9 entry to `REGIONS` in `const.py` and route Kia + CCI to `KiaCciApiEU` where the component picks the API class. I have not upstreamed that yet — it depends on the lib PR merging first (and kia_uvo pinning a version that contains it).

Things worth capturing from the patched instance, since they will tell us what Pakiet B needs:

1. **Force-refresh cycle** — `KiaCciApiEU` does not override `force_refresh_vehicle_state` yet, so a force-refresh cycle (restart, the force-refresh button, some coordinator paths) will log a `NotImplementedError`. That is a known gap (Pakiet B), not something you did wrong — but the log line + when exactly it fires is useful.
2. **`check_action_status` after a command** — intentionally gated on Kia: on this backend an update-status call triggers a *new* status upload rather than polling an action, so the library refuses for now. Capturing what the coordinator logs right after a lock/unlock would help design the eventual replacement.
3. **A debug log of one full coordinator cycle** (lock → wait → poll) so we can see the entity flip timing against the stored-status timestamps you already measured (7 s lock / 4 s unlock).

No rush — whenever you get to it.

blka:
Fair point on size — agreed that follow-up work stays in small PRs. This one is the minimum coherent unit for the GSPA control layer, though: the PIN control-token flow, the envelope handling and the command builders are one contract that only makes sense together, and the two live test runs (EV6, PV5) exercised exactly this head. Splitting the parser out now would leave it un-verifiable on its own.

What *is* planned as separate small PRs after this merges:

1. CCS2 parser follow-up for the no-sample fields (#1321);
2. kia_uvo region-9 entry + API-class routing (integration side, depends on this merge);
3. `force_refresh_vehicle_state` / a real `check_action_status` replacement for Kia (update-status triggers a new upload on Kia instead of polling — needs its own design).

hyundai_kia_connect_api/ApiImpl.py
cdnninja: I think the goal is to obfuscate and make this universal. I still don't see why we can't do that.

hyundai_kia_connect_api/ApiImpl.py
```diff
@@ -56,13 +63,35 @@ class WindowRequestOptions:
...
+    # endpoint-specific, and only the front seats genuinely differ: the GSPA
+    # driver/passenger split depends on drvSeatLoc (LHD/RHD), which the
+    # library cannot derive itself.
+    driver_seat_window: WINDOW_STATE | None = None
+    passenger_seat_window: WINDOW_STATE | None = None
+    rear_left_window: WINDOW_STATE | None = None
+    rear_right_window: WINDOW_STATE | None = None
+    rear_left_curtain: WINDOW_STATE | None = None
+    rear_right_curtain: WINDOW_STATE | None = None
+    driver_seat_location: str | None = None  # "L" or "R" (GSPA drvSeatLoc)
```
cdnninja: Should be an ENUM like the others.

blka:
Done in f323ea8 — the seat-based fields are gone. `WindowRequestOptions` now takes only the position-based vocabulary (`front_left`/`front_right`/`back_left`/`back_right`, physical windows) plus the two curtains. The GSPA seat-based body keys are derived inside the API layer: `drvSeatLoc` comes from `_get_drv_seat_loc(vehicle)` (moved from `ApiImplType1` to `ApiImpl` so both layers share it) with `driver_seat_location` as an explicit override, and on RHD `front_left` maps to `psgSeatWindow` while `front_right` maps to `drvSeatWindow`. So a caller says "close the front-left window" in one vocabulary regardless of which endpoint the region uses.

blka:
Done in f323ea8 — `driver_seat_location` is now a `SEAT_LOCATION` str-enum (`LEFT`/`RIGHT`, wire values `"L"`/`"R"`) on both `WindowRequestOptions` and `ClimateRequestOptions` (the climate options carried the same raw-string field). The API layer now derives the value from the vehicle (`_get_drv_seat_loc`, odometer-unit based, region overrides still work) and the option is the override.

PR #1306, opened 2026-09-05
fix(EU): accept flat resMsg location shape in _get_location

### LINKED ISSUE #1301: _get_location() silently drops valid location data when resMsg has no gpsDetail wrapper (EU, non-CCS2)
At first I had noticed that the HomeAssistent integration based on this library - "sometimes" - only delivers "location=null".
I then started to debug the behavior using the library itself as a standalone "get_location" script.

With help from claude.ai I found the following and claude formulated this report:

### Library version

```
hyundai_kia_connect_api==4.27.3
```

`pip show` output:

```
Name: hyundai_kia_connect_api
Version: 4.27.3
Home-page: https://github.com/Hyundai-Kia-Connect/hyundai_kia_connect_api
```

### Environment

- Region: Europe (`REGION_EUROPE`)
- Brand: Kia
- Vehicle: Kia Sorento (2021), ICE
- `protocolType: 0`, `ccuCCS2ProtocolSupport: 0` (older, non-CCS2 telematics protocol)
- Login flow observed: `idpconnect-eu.kia.com` OAuth2 → `cci-api-eu.kia.com` token exchange (`serviceType=CCS`) → `prd.eu-ccapi.kia.com:8080` vehicle API

### Summary

`KiaUvoApiEU._get_location()` assumes the `/location` (and, by the same helper, `/location/park`) endpoint always wraps its payload in a `resMsg.gpsDetail` object:

```python
def _get_location(self, token: Token, vehicle: Vehicle) -> dict:
    url = self.SPA_API_URL + "vehicles/" + vehicle.id + "/location"
    ...
    response = self.session.get(url, headers=...).json()
    gps_detail = response["resMsg"].get("gpsDetail")
    if gps_detail is None:
        _LOGGER.warning(
            f"{DOMAIN} - gpsDetail not found in location response, "
            "vehicle may be offline or returning partial status"
        )
    return gps_detail
```

For my account/vehicle, the live `/location` response does **not** use a `gpsDetail` wrapper — the fields (`coord`, `head`, `speed`, `accuracy`, `time`) are top-level inside `resMsg` directly. As a result `gpsDetail` is always `None` even though the HTTP call succeeded and returned fully valid, fresh GPS data (including non-zero `head`/`speed` fields). The misleading "vehicle may be offline" warning fires despite the vehicle being online and the response containing good data, and — more importantly — **the valid location update is silently discarded**, since `force_refresh_vehicle_state()` does:

```python
state = self._get_forced_vehicle_state(token, vehicle)
state["vehicleLocation"] = self._get_location(token, vehicle)   # -> None
```

`_update_vehicle_properties()` only writes `vehicle.location` when `get_child_value(state, "vehicleLocation.coord.lat")` is truthy, so with `state["vehicleLocation"] = None` the branch is skipped entirely. The net effect: **a `force_refresh_vehicle_state()` call that successfully talked to the car does not update `vehicle.location` / `vehicle.location_latitude` / `vehicle.location_longitude` / `vehicle.location_last_updated_at` at all**, with no exception raised — only a debug-level warning that undersells the severity.

### Call made

```python
vm.force_refresh_vehicle_state(vehicle_id)
```

Internally triggers (non-CCS2 branch):

```python
state = self._get_forced_vehicle_state(token, vehicle)
state["vehicleLocation"] = self._get_location(token, vehicle)
```

### Relevant log lines

Raw, successful HTTP response from `/location` (redacted coordinates/ids):

```
10:58:45 [DEBUG] hyundai_kia_connect_api.KiaUvoApiEU: hyundai_kia_connect_api - _get_location response: {'retCode': 'S', 'resCode': '0000', 'resMsg': {'coord': {'lat': 50.0607xx, 'lon': 8.3661xx, 'alt': 0, 'type': 0}, 'head': 0, 'speed': {'value': 0, 'unit': 0}, 'accuracy': {'hdop': 0, 'pdop': 0}, 'time': '20260903105844'}, 'msgId': '<redacted>'}
```

Immediately followed by the misleading warning, even though the response above is a well-formed, complete payload with fresh `time`:

```
10:58:45 [WARNING] hyundai_kia_connect_api.KiaUvoApiEU: hyundai_kia_connect_api - gpsDetail not found in location response, vehicle may be offline or returning partial status
```

Resulting `vehicle.data["vehicleLocation"]` after the force-refresh call (dumped via `dataclasses.fields` + all properties) — note the discarded location:

```json
"vehicleLocation": null,
```

And `vehicle.location_last_updated_at` unchanged from the *previous, cached* read (10:40:06) despite the force-refresh call completing successfully at 10:58:42–10:58:45 (16.8s round-trip):

```json
"_location_last_set_time": "2026-09-03T10:40:06+02:00",
"location_last_updated_at": "2026-09-03T10:40:06+02:00",
```

For comparison, the *cached* read a few steps earlier in the same run used `get_cached_vehicle_status` (`/status/latest`), whose payload embeds the location under `resMsg.vehicleStatusInfo.vehicleLocation` — no `gpsDetail` wrapper there either, and that one parses fine because `_update_vehicle_properties` reads `vehicleLocation.coord.lat` directly, not through `_get_location()`:

```
10:58:28 [DEBUG] hyundai_kia_connect_api.KiaUvoApiEU: hyundai_kia_connect_api - get_cached_vehicle_status response: {'retCode': 'S', 'resCode': '0000', 'resMsg': {'vehicleStatusInfo': {'vehicleLocation': {'coord': {'lat': 50.0607xx, 'lon': 8.3661xx, 'alt': 0, 'type': 0}, 'head': 127, 'speed': {'value': 0, 'unit': 0}, 'accuracy': {'hdop': 0, 'pdop': 1}, 'time': '20260903104006'}, ...
```

So the same account/vehicle exposes `head`/`speed` consistently at the top level of the location object across *both* endpoints (`/status/latest` and `/location`) — never nested under `gpsDetail`. This suggests the `gpsDetail` wrapper assumption in `_get_location()` does not hold for at least some EU/non-CCS2 accounts (possibly a legacy vs. new EU platform migration difference, or brand/region-specific response shape).

### Expected behavior

`_get_location()` should tolerate both response shapes:
- `resMsg.gpsDetail.{coord,head,speed,...}` (wrapped), and
- `resMsg.{coord,head,speed,...}` (flat, as observed here),

e.g. by falling back to `resMsg` itself when `gpsDetail` is absent but `resMsg.coord` is present, before deciding the response is actually empty/offline. At minimum, the "vehicle may be offline" warning should not fire when the response demonstrably contains a valid, fresh `coord`/`time` payload — that log message actively misleads troubleshooting.

### Actual behavior

- `_get_location()` returns `None` even though the HTTP call succeeded and the JSON contained complete, fresh location data (including `head` and `speed`).
- `force_refresh_vehicle_state()` (non-CCS2 branch) then sets `state["vehicleLocation"] = None`, so `_update_vehicle_properties()` skips updating `vehicle.location*` entirely — no exception, no indication to the caller that the location update was dropped.
- A misleading "vehicle may be offline or returning partial status" warning is logged despite the vehicle being online and the response being complete.

### Suggested fix

```python
def _get_location(self, token: Token, vehicle: Vehicle) -> dict:
    url = self.SPA_API_URL + "vehicles/" + vehicle.id + "/location"
    try:
        response = self.session.get(
            url,
            headers=self._get_authenticated_headers(
                token, vehicle.ccu_ccs2_protocol_support
            ),
        ).json()
        _LOGGER.debug(f"{DOMAIN} - _get_location response: {response}")
        _check_response_for_errors(response)
        res_msg = response.get("resMsg", {}) or {}
        gps_detail = res_msg.get("gpsDetail")
        if gps_detail is None and res_msg.get("coord"):
            # Some accounts/regions return the location fields flat under
            # resMsg instead of nested under gpsDetail.
            gps_detail = res_msg
        if gps_detail is None:
            _LOGGER.warning(
                f"{DOMAIN} - gpsDetail not found in location response, "
                "vehicle may be offline or returning partial status"
            )
        return gps_detail
    except Exception as e:
        _LOGGER.exception(f"{DOMAIN} - _get_location failed: {e}")
        return None
```

I hope this all makes sense and gives some hints to improve or fix this issue.

blka:
## Problem

Some EU (non-CCS2) accounts return `GET /location` with the location fields directly under `resMsg`, without the `gpsDetail` wrapper:

```
'resMsg': {'coord': {'lat': …, 'lon': …, 'alt': 0, 'type': 0},
           'head': 0, 'speed': {'value': 0, 'unit': 0},
           'accuracy': {'hdop': 0, 'pdop': 0},
           'time': '20260903105844'}
```

`_get_location()` reads only `resMsg.gpsDetail`, so for these accounts:

- the HTTP call succeeds and the response contains valid, fresh location data, but
- `_get_location()` returns `None` and logs `gpsDetail not found … vehicle may be offline or returning partial status` — misleading, since the vehicle is online and the payload is complete,
- `_update_vehicle_properties()` then skips the `vehicle.location*` fields entirely: a successful `force_refresh_vehicle_state()` does not update the location at all (reported in #1301 with the raw response).

## Fix

Fall back to `resMsg` itself when `gpsDetail` is absent but it carries a `coord`. This mirrors the existing parsing of `GET /location/park` (`_set_cached_location_park`), which already reads the flat `resMsg` shape directly — the same API family, same shape.

The offline/partial warning now only fires when the response truly contains no location data, and its wording no longer references `gpsDetail` (which is not what is missing in the flat case).

Downstream (`_set_location_from_gps_detail`, UTC handling from #1292) is unchanged: the flat `resMsg` carries exactly the keys it reads (`coord`, `time`).

## Tests

New `tests/test_eu_location_flat_resmsg.py` (built from the raw response in #1301):

- flat `resMsg` → location applied, time converted as UTC like the wrapped shape
- wrapped `gpsDetail` (both present) → unchanged behavior, wrapper takes precedence
- neither → no fallback, previously known position kept, warning logged

Closes #1301

### DIFF
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -1291,10 +1291,15 @@ class KiaUvoApiEU(ApiImplType1):
             ).json()
             _LOGGER.debug(f"{DOMAIN} - _get_location response: {response}")
             _check_response_for_errors(response)
-            gps_detail = response["resMsg"].get("gpsDetail")
+            res_msg = response["resMsg"]
+            gps_detail = res_msg.get("gpsDetail")
+            if gps_detail is None and res_msg.get("coord"):
+                # Some accounts return the location fields flat under resMsg
+                # (GET /location/park already uses the same flat shape).
+                gps_detail = res_msg
             if gps_detail is None:
                 _LOGGER.warning(
-                    f"{DOMAIN} - gpsDetail not found in location response, "
+                    f"{DOMAIN} - no location data in location response, "
                     "vehicle may be offline or returning partial status"
                 )
             return gps_detail
--- a/hyundai_kia_connect_api/KiaUvoApiEU.py
+++ b/hyundai_kia_connect_api/KiaUvoApiEU.py
@@ -1292,16 +1292,9 @@ class KiaUvoApiEU(ApiImplType1):
             _LOGGER.debug(f"{DOMAIN} - _get_location response: {response}")
             _check_response_for_errors(response)
             res_msg = response["resMsg"]
-            gps_detail = res_msg.get("gpsDetail")
-            if gps_detail is None and res_msg.get("coord"):
-                # Some accounts return the location fields flat under resMsg
-                # (GET /location/park already uses the same flat shape).
-                gps_detail = res_msg
-            if gps_detail is None:
-                _LOGGER.warning(
-                    f"{DOMAIN} - no location data in location response, "
-                    "vehicle may be offline or returning partial status"
-                )
+            # Some accounts return the location fields flat under resMsg
+            # (GET /location/park already parses the same flat shape, #1301).
+            gps_detail = res_msg.get("gpsDetail") or res_msg
             return gps_detail
         except Exception as e:
             _LOGGER.exception(f"{DOMAIN} - _get_location failed: {e}")  # noqa: TRY401
@@ -1320,6 +1313,10 @@ class KiaUvoApiEU(ApiImplType1):
         offset early. Convert here, as the CCS2 'Date' field does (#1147).
         """
         if not gps_detail or not get_child_value(gps_detail, "coord.lat"):
+            _LOGGER.warning(
+                f"{DOMAIN} - no location data in location response, "
+                "vehicle may be offline or returning partial status"
+            )
             return
         # A missing timestamp must stay None, which HA shows as "unknown".
         # See kia_uvo #1771.
--- a/tests/test_eu_location_flat_resmsg.py
+++ b/tests/test_eu_location_flat_resmsg.py
@@ -1,0 +1,123 @@
+"""EU forced refresh: GET /location may return the location fields flat under
+resMsg, without the gpsDetail wrapper (#1301).
+
+Some EU non-CCS2 accounts (Kia Sorento 2021 reported in #1301) return
+resMsg.{coord, head, speed, accuracy, time} directly, the same flat shape GET
+/location/park already parses. The old code read only resMsg.gpsDetail and
+silently dropped the position.
+"""
+
+import datetime as dt
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from hyundai_kia_connect_api.KiaUvoApiEU import KiaUvoApiEU
+from hyundai_kia_connect_api.Vehicle import Vehicle
+
+# Raw response shape from #1301: the fields sit directly in resMsg.
+FLAT_LOCATION = {
+    "retCode": "S",
+    "resCode": "0000",
+    "resMsg": {
+        "coord": {"lat": 52.52, "lon": 13.405, "alt": 0, "type": 0},
+        "head": 127,
+        "speed": {"value": 0, "unit": 0},
+        "accuracy": {"hdop": 0, "pdop": 0},
+        "time": "20260903105844",
+    },
+}
+WRAPPED_LOCATION = {
+    "retCode": "S",
+    "resCode": "0000",
+    "resMsg": {
+        "gpsDetail": {
+            "coord": {"lat": 48.85, "lon": 2.35},
+            "time": "20260903110000",
+        }
+    },
+}
+FORCED_STATUS = {"retCode": "S", "resCode": "0000", "resMsg": {"doorLock": True}}
+
+
+@pytest.fixture
+def eu_api() -> KiaUvoApiEU:
+    api = KiaUvoApiEU.__new__(KiaUvoApiEU)
+    api.SPA_API_URL = "https://test.invalid/api/v1/spa/"
+    api.data_timezone = KiaUvoApiEU.data_timezone
+    api.temperature_range = KiaUvoApiEU.temperature_range
+    api.session = MagicMock()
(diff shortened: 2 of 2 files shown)


hyundai_kia_connect_api/KiaUvoApiEU.py
```diff
@@ -1291,10 +1291,15 @@ def _get_location(self, token: Token, vehicle: Vehicle) -> dict:
             ).json()
             _LOGGER.debug(f"{DOMAIN} - _get_location response: {response}")
             _check_response_for_errors(response)
-            gps_detail = response["resMsg"].get("gpsDetail")
+            res_msg = response["resMsg"]
+            gps_detail = res_msg.get("gpsDetail")
```
cdnninja: gps_detail = res_msg.get("gpsDetail") or res_msg.

Then pull the below if statement and remove the warning.  Warning can live here re-using an existing if statement. 

https://github.com/blka/hyundai_kia_connect_api/blob/d129f14dcdb04a8e21744692bcc136d4ce121904/hyundai_kia_connect_api/KiaUvoApiEU.py#L1323

blka: Done in 48db2f7 — the fallback is now a single expression and the offline warning lives in the existing coord guard of `_set_location_from_gps_detail`. The guard fires only from the forced-refresh path (its only caller), so the warning keeps its meaning. Tests updated and green (642 passed).

PR #1307, opened 2026-09-05
fix: isolate vehicle-level API errors in aggregate update loops

### LINKED ISSUE #1280: One unreachable vehicle (resCode 5031) takes down all others — no per-vehicle error isolation in update_all_vehicles_with_cached_state
## Summary

`VehicleManager.update_all_vehicles_with_cached_state` iterates over all vehicles with no per-vehicle error handling. The first vehicle that raises aborts the entire loop, and the exception propagates up to Home Assistant as `ConfigEntryNotReady`.

The practical consequence: **one vehicle the backend no longer serves makes every other vehicle in the account permanently unavailable**, even when their data was fetched successfully moments earlier in the same cycle.

```python
def update_all_vehicles_with_cached_state(self) -> None:
    for vehicle_id in self.vehicles:
        self.update_vehicle_with_cached_state(vehicle_id)   # first raise ends everything
```

## Concrete case (Hyundai EU, 2 vehicles on one account)

- Vehicle A (2026, `ccuCCS2ProtocolSupport: 1`) → `retCode: 'S'`, full CCS2 payload, valid data
- Vehicle B (2023, `protocolType: 0`) → `retCode: 'F', resCode: '5031'`, "Unavailable remote control - Service Temporary Unavailable"

Vehicle B was sold and is no longer owned, but is still listed on the Bluelink account and cannot currently be removed via the app. It is unlikely to ever return a successful response again.

Debug logging over several setup cycles showed a clean 1:1 alternation — 6 successful `get_cached_vehicle_status` responses and 6 × `5031`, i.e. exactly one success and one failure per cycle. Vehicle A is fetched successfully every single time, and its data is then discarded because vehicle B raises immediately afterwards.

Log line (redacted):

```
DEBUG [hyundai_kia_connect_api.KiaUvoApiEU] get_cached_vehicle_status response:
  {'retCode': 'F', 'resCode': '5031', 'resMsg': 'Unavailable remote control - Service Temporary Unavailable', ...}
```

Resulting Home Assistant state: `Config Not Ready: Unavailable remote control - Service Temporary Unavailable`, retried indefinitely, **zero** entities for either vehicle.

Versions: `kia_uvo` v3.10.1, `hyundai_kia_connect_api` 4.26.5 (both current; login/WAF handling from 4.26.3 works correctly here — the vehicle list is fetched without any problem).

## Why this is distinct from #1774 (kia_uvo)

In [kia_uvo#1774](https://github.com/Hyundai-Kia-Connect/kia_uvo/issues/1774), `5031` is explained as a transient server-side outage where retrying is the correct behaviour. That reasoning is sound for a single-vehicle account, and this report does not dispute it.

The separate issue here is the **blast radius**: when `5031` is permanent for one vehicle, retrying never converges, and healthy vehicles are taken down with it. That is a robustness property of the loop, independent of which vehicle fails or why.

## Suggested fix

Isolate failures per vehicle, and only raise when *no* vehicle could be updated — so a genuine full outage is still reported rather than silently swallowed:

```python
def update_all_vehicles_with_cached_state(self) -> None:
    last_error = None
    updated = 0
    for vehicle_id in list(self.vehicles):
        try:
            self.update_vehicle_with_cached_state(vehicle_id)
            updated += 1
        except Exception as err:
            last_error = err
            _LOGGER.warning("Skipping vehicle, cached update failed: %s", err)
    if updated == 0 and last_error is not None:
        raise last_error
```

`update_vehicle_with_cached_state` already honours `if vehicle.enabled:`, so an alternative (or complement) would be for the integration to expose `disable_vehicle()` — currently it exists in the library (`VehicleManager`, line ~350) but is never called from `kia_uvo`, leaving users with no supported way to skip a dead vehicle.

The same pattern applies to `force_refresh_all_vehicles_states` and `check_and_force_update_vehicles`.

I am running the per-vehicle isolation above as a local patch in `coordinator.py` and it resolves the situation — the working vehicle reports normally while the dead one is logged and skipped. Happy to open a PR against the library if the approach looks right to you.

blka:
## Problem

A single unreachable vehicle (resCode `5031`) aborts the whole update. The three aggregate loops in `VehicleManager` have no per-vehicle error handling, so one failing vehicle raises out of the loop and the caller never reaches the remaining vehicles. In Home Assistant, where the account runs on a single coordinator (`DataUpdateCoordinator`), one bad VIN means one `UpdateFailed` for the whole account: **every** entity of **every** vehicle goes unavailable, including the healthy ones (#1280). kia_uvo uses all three loops (`check_and_force_update_vehicles`, `update_all_vehicles_with_cached_state`, `force_refresh_all_vehicles_states`), so the failure path is shared.

## Fix

The three aggregate loops now catch vehicle-level API errors per vehicle, log a warning naming the vehicle, and continue with the next one:

```python
for vehicle_id in self.vehicles:
    try:
        self.update_vehicle_with_cached_state(vehicle_id)
    except RateLimitingError:
        raise
    except APIError as e:
        _LOGGER.warning(f"{DOMAIN} - Failed to update vehicle {vehicle_id}: {e}")
```

### What propagates, and why

- **`AuthenticationError`** propagates: an expired token fails identically for every vehicle; swallowing it would mask the re-auth path behind "success".
- **`RateLimitingError`** propagates (`except RateLimitingError: raise`): rate limiting is account-level — isolating it would let the caller keep polling at the same pace and make the problem worse.
- **Other exceptions** (network, non-library) propagate: they are account-level too.
- **`APIError` and its vehicle-level subclasses** (`InvalidAPIResponseError`, `DeviceIDError`, `DuplicateRequestError`, `ServiceTemporaryUnavailable`, `NoDataFound`, …) are isolated — resCode `5031` is this case.

### Explicit single-vehicle calls are unchanged

`update_vehicle_with_cached_state(vehicle_id)`, `check_and_force_update_vehicle(...)`, `force_refresh_vehicle_state(vehicle_id)` still raise — a direct call on one vehicle is an explicit request and the caller must see the failure.

### Known tradeoff

The failing vehicle keeps its last known data (stale, with a warning in the log) instead of its entities going unavailable. That is the pre-existing behavior for that vehicle; the change is that the healthy vehicles keep updating instead of being dragged down with it.

## Tests

New `tests/test_vehicle_manager_update_isolation.py` (7 tests):

- each of the three loops isolates a failing vehicle and still updates the second
- `AuthenticationError` and `RateLimitingError` propagate out of the loop
- a direct single-vehicle call still raises
- the warning names the vehicle and the error

Closes #1280

### DIFF
--- a/hyundai_kia_connect_api/VehicleManager.py
+++ b/hyundai_kia_connect_api/VehicleManager.py
@@ -36,7 +36,7 @@ from .const import (
     VALET_MODE_ACTION,
     VEHICLE_LOCK_ACTION,
 )
-from .exceptions import APIError, AuthenticationOTPRequired
+from .exceptions import APIError, AuthenticationOTPRequired, RateLimitingError
 from .HyundaiBlueLinkApiBR import HyundaiBlueLinkApiBR
 from .HyundaiBlueLinkApiUSA import HyundaiBlueLinkApiUSA
 from .HyundaiCciApiEU import HyundaiCciApiEU
@@ -140,7 +140,14 @@ class VehicleManager:
 
     def update_all_vehicles_with_cached_state(self) -> None:
         for vehicle_id in self.vehicles:
-            self.update_vehicle_with_cached_state(vehicle_id)
+            try:
+                self.update_vehicle_with_cached_state(vehicle_id)
+            except RateLimitingError:
+                raise
+            except APIError as e:
+                _LOGGER.warning(
+                    f"{DOMAIN} - Failed to update vehicle {vehicle_id}: {e}"
+                )
 
     def update_vehicle_with_cached_state(self, vehicle_id: str) -> None:
         vehicle = self.get_vehicle(vehicle_id)
@@ -160,7 +167,14 @@ class VehicleManager:
 
     def check_and_force_update_vehicles(self, force_refresh_interval: int) -> None:
         for vehicle_id in self.vehicles:
-            self.check_and_force_update_vehicle(force_refresh_interval, vehicle_id)
+            try:
+                self.check_and_force_update_vehicle(force_refresh_interval, vehicle_id)
+            except RateLimitingError:
+                raise
+            except APIError as e:
+                _LOGGER.warning(
+                    f"{DOMAIN} - Failed to update vehicle {vehicle_id}: {e}"
+                )
 
     def check_and_force_update_vehicle(
         self, force_refresh_interval: int, vehicle_id: str
@@ -184,7 +198,14 @@ class VehicleManager:
 
     def force_refresh_all_vehicles_states(self) -> None:
         for vehicle_id in self.vehicles:
-            self.force_refresh_vehicle_state(vehicle_id)
+            try:
+                self.force_refresh_vehicle_state(vehicle_id)
+            except RateLimitingError:
+                raise
+            except APIError as e:
+                _LOGGER.warning(
+                    f"{DOMAIN} - Failed to update vehicle {vehicle_id}: {e}"
+                )
 
     def force_refresh_vehicle_state(self, vehicle_id: str) -> None:
         vehicle = self.get_vehicle(vehicle_id)
--- a/tests/test_vehicle_manager_update_isolation.py
+++ b/tests/test_vehicle_manager_update_isolation.py
@@ -1,0 +1,103 @@
+"""Per-vehicle error isolation in the aggregate update loops (#1280).
+
+One unreachable vehicle (resCode 5031) used to abort the whole update, taking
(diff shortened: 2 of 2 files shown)


cdnninja:
This masks all our hard work on proper exceptions and also breaks some of HAs built in retry logic.    I think we should solve this HA side.

blka:
Agreed — closing this. On reflection, isolating vehicle-level errors at the library level hides the typed exceptions from the caller and short-circuits the retry semantics the Home Assistant coordinator is built around (`UpdateFailed` → `unavailable` → recovery on the next poll). Your direction from the issue is the right layer: vehicle-specific polling in the integration, honoring Home Assistant's own `disabled_by` device state, rather than any exception handling here.

PR #1309, opened 2026-09-09
feat(KR): add MyHyundai Korea support

tastelessjolt:
## Summary

Add experimental Hyundai Korea support for the unified MyHyundai app as region `10`.

This includes:

- Pleos browser OAuth login using the production Korean OneApp client and offline scope
- renewable CCI token serialization that omits the password, PIN, and short-lived control token
- Korean vehicle discovery and correct ICE/HEV/PHEV/EV classification
- cached and forced CCS2 vehicle status
- native capability-aware Korea window-status interpretation
- door lock and PIN-authorized unlock
- GEN2 ICE and EV remote climate request models
- remote-start eligibility and capability discovery
- per-seat heating and ventilation settings
- heated steering-wheel settings
- a browser-login setup script and usage documentation

## Live validation

Tested against one Korean Hyundai ICE vehicle using the current MyHyundai app protocol:

- Pleos browser authorization and token exchange
- CCI token rotation and session restoration
- vehicle discovery and cached status
- forced refresh
- door lock, unlock, and relock
- GEN2 engine/climate start and stop
- driver and passenger low heat, including observed status
- driver low ventilation, including observed status
- heated steering wheel on, including observed status
- final stop with engine, seats, and steering-wheel heat observed off

The live vehicle reported passenger-seat ventilation and rear-seat climate as unsupported, so those commands were not sent. EV controls use the same current app request model but were not live-tested because the available vehicle is not an EV.

Window-status parsing was corrected after the API initially reported all four physically closed windows as open. The implementation now follows the same `windowSafetyOption` / `windowSafetyOption2` and `OpenLevel` rules as the native app. Korea does not advertise window operation support because this PR only implements window status.

## Authentication notes

Pleos rejects scripted password login for this account type, so the initial login is interactive. The OAuth callback page redirects desktop browsers immediately. The included setup script explains how to capture the registered `oneapp.hyundai.com/redirect` request using the browser network log with Preserve log enabled.

Saved sessions omit the password, PIN, and short-lived vehicle-control token. Refresh tokens may rotate, so callers must save the current token after a successful refresh. Generic `Token.save()` / `Token.load()` filesystem helpers have been split into draft PR #1326; this PR keeps file handling inside the Korea setup script.

## Shared EU CCI compatibility changes

Korea builds on the existing GSPA/CCI implementation, so this PR also makes the following changes visible to existing EU CCI users:

- token validation now treats network failures, HTTP 429, and HTTP 5xx as inconclusive instead of forcing a new login; HTTP 401/403 still rejects the token
- token refresh maps HTTP 401/403 to `AuthenticationError` and transient network, HTTP 429, and HTTP 5xx failures to `ServiceTemporaryUnavailable`
- refresh authorization headers and `idToken` are configurable for Korea; the EU defaults remain enabled
- RSA password encryption selects RSA-OAEP-SHA256 when advertised by the identity-provider certificate and otherwise retains PKCS#1 v1.5
- vehicle metadata accepts boolean-like strings and integers and compares engine/fuel codes case-insensitively
- unknown temperature, distance, and seat-status enum values are preserved as unavailable instead of raising `KeyError`

## Tests

- local `pytest -q`: 775 passed on Python 3.13
- local Ruff check and format: passed
- repository-configured pre-commit suite: passed; the existing mypy file pattern does not cover package files
- GitHub Test Builds and CodeQL: awaiting maintainer workflow approval

## Known limitations

On the live-tested Korean vehicle/account, MyHyundai's `infolist_v2.do` capability endpoint currently returns HTTP 500. This endpoint supplies optional metadata used for window-status interpretation, remote-start eligibility, seat heating/ventilation, and steering-wheel heating. The implementation treats non-authentication failures from capability discovery as non-fatal: cached vehicle status and independent controls remain available, affected capability fields stay unknown, and capability-dependent commands fail safely instead of sending an unvalidated request. The four window-status sensors therefore remain unavailable until capability discovery succeeds. The endpoint-specific cause is still under investigation.

Korea force refresh is synchronous. It waits about 25 seconds before the first cached-status readback and can perform two additional 10-second waits, so the caller may remain occupied for roughly 45 seconds plus network time.


tastelessjolt:
Hello @cdnninja, I would like to request review for this PR. I'm sorry for the ping as I don't have permission to request review.

blka (commented):
Thanks for the huge amount of work on this — the protocol-level test coverage, the atomic `Token.save()`, and the honest documentation of the `infolist_v2.do` limitation are all great. A few things worth addressing before merge:

## 1. Undeclared `Vehicle` attributes (should block merge)

`HyundaiConnectApiKR` sets ~10 attributes on `Vehicle` that are not declared on the dataclass — `Vehicle.py` is not part of this PR:

- `window_safety_option`, `window_safety_option2`
- `remote_control_generation`, `supports_remote_start`, `remote_control_waiting_time`
- `hvac_temperature_type`
- `steering_wheel_heater_option`, `steering_wheel_heating_option`, `supports_steering_wheel_heater`
- `window_status_capabilities_loaded`

Python allows the assignments at runtime (no `__slots__`), so the unit tests pass, but:

- `tests/vehicle_snapshot_serializer.py` serializes `dataclasses.fields()` only — these attributes are invisible in snapshot tests and in the diagnostics dump the HA integration produces, so field-mapping bugs on KR vehicles can't be diagnosed from user dumps.
- It breaks the repo convention that all vehicle state lives as declared `Vehicle` fields.
- mypy flags all of these (`attr-defined`) when pointed at the file — they only go unnoticed because the pre-commit mypy hook's `files` pattern (`^(/.+)?[^/]+\.py$`) matches root-level `.py` files only, so mypy effectively never checks the package (pre-existing config issue, separate problem).

Suggest adding the fields to the `Vehicle` dataclass in this PR. Note that adding fields requires regenerating the vehicle snapshots in the same change.

## 2. CI on this PR never actually ran

"Test Builds" (pytest) and CodeQL are stuck in `action_required` (first-time-contributor workflow approval), so the "774 passed" in the description has not been confirmed by CI — a maintainer needs to approve the workflow runs on the Actions tab.

## 3. EU-side behavior changes bundled in, and absent from, the description

The PR also changes live behavior of `GspaApiEU`/`HyundaiCciApiEU` for existing EU users, but the description presents this as purely a KR feature:

- `test_token`: transport failures and 429/5xx now return `True` ("inconclusive") instead of `False`. Individually I think this is right (a transient failure shouldn't force a re-login), but it changes EU semantics and deserves its own paragraph in the description.
- `_refresh_cci_token` error taxonomy reworked (401/403 → `AuthenticationError`, 429/5xx → `ServiceTemporaryUnavailable`), and the idToken header is now conditional.
- `HyundaiCciApiEU` parsers switched from `TEMPERATURE_UNITS[...]`/`DISTANCE_UNITS[...]`/`SEAT_STATUS[...]` to `.get(...)` — graceful handling of unknown enum values.

None of these look wrong to me, but they're user-visible EU changes riding in a KR PR; please call them out explicitly in the description (or split into a separate PR).

## 4. Blocking `time.sleep` in force refresh

`force_refresh_vehicle_state` sleeps up to 45 s (25+10+10) inside the coordinator's update thread. The library is synchronous so this is consistent with the rest of the codebase, but it's by far the longest force-refresh path in the library — worth a mention in the README/integration docs so users understand force refresh on KR vehicles blocks for ~45 s.

## 5. Minor

`scripts/myhyundai_kr_login.sh` carries generator boilerplate ("Generated by the /wizard skill", "Everything above the STAGES marker is the wizard library: do not hand-edit it") that reads oddly in a published repo — consider trimming it to a plain step-by-step script header.

## Optional: fully non-interactive login

Since the initial login is the only interactive step: it should be scriptable with a small local headless-browser script (e.g. Playwright) that drives the Pleos OAuth consent and captures the `oneapp.hyundai.com/redirect` request automatically, instead of asking users to fish the request URL out of browser devtools. Happy to help sketch that if useful.

tastelessjolt:
Thanks for the detailed review. I addressed the merge-blocking and documentation points in `876ae79` after rebasing the branch onto current `master` (`4.31.0`):

- Declared all 14 Korea capability attributes on `Vehicle` (the 10 listed plus the four seat-climate capability fields), regenerated every vehicle snapshot, and added a regression assertion that the populated values survive `dataclasses.asdict()`.
- Replaced the corresponding defensive dynamic reads in `HyundaiConnectApiKR` with declared-field access.
- Documented the shared EU CCI behavior changes in the PR description, including transient token validation, refresh error taxonomy/default-preserving header flags, enum tolerance, RSA-OAEP selection, and metadata normalization.
- Documented the synchronous 25–45 second Korea force-refresh timing in both the README and PR description.
- Removed the generated-wizard wording and unused generic helper code from the Pleos login script.
- Corrected the test section: 776 tests pass locally on Python 3.12, 3.13, and 3.14, and the repository-configured pre-commit suite passes. The description now explicitly notes the existing mypy file-pattern gap and that GitHub Test Builds/CodeQL still need maintainer workflow approval.

I have left fully automated headless Pleos login out of this PR. It would introduce a substantial browser dependency and credential-automation surface for a one-time login, so I think it is better considered separately after the core region support is reviewed.

blka:
Verified `876ae792` locally against the full suite, plus the EU shared-path impact specifically (Hyundai and Kia EU both inherit login/refresh/test_token from `GspaApiEU`):

- `pytest`: 776 passed; `ruff check` and `ruff format --check` clean.
- Round-1 blocker resolved: all 14 capability fields declared on `Vehicle`, snapshots regenerated (new keys are `None`/`False` for existing regions), the dynamic `getattr` reads replaced, and the `asdict()` regression assertion present.
- EU neutrality checked: the `LOGIN_*` class-attr defaults reproduce the previous hardcoded values, the `CCI_REFRESH_*` flags default to the previous request shape, and I probed both EU cert endpoints (`idpconnect-eu.hyundai.com`, `idpconnect-eu.kia.com`) — neither advertises `RSA-OAEP-256`, so the new cipher selection falls back to PKCS#1 v1.5, unchanged for EU login. The only residual observation is that the authorize URL is now percent-encoded where it was raw; that endpoint sits behind the WAF from #1273, so worth watching on the first EU login report.
- The PR body now covers the shared EU CCI changes, the README covers the 45-second force-refresh, and the login script is de-wizarded. Agreed on deferring headless Pleos login.

Two non-blocking mypy-strict nits in `HyundaiConnectApiKR.py`, both invisible to CI because of the pre-existing mypy file-pattern gap noted in round 1:

1. `SUPPORTED_LANGUAGES = ("ko", "en")` overrides the base class's `list[str]` — align the types.
2. The `SEAT_CLIMATE_FIELDS` dispatch (`getattr(vehicle, capability_field, None)` → `SEAT_CLIMATE_STATES.get(capability)` → `not in allowed`) surfaces as `Any | None` under strict mypy; the runtime guards are correct, but keying off the declared fields would type-check cleanly.

Merge is still gated on CI rather than review: Test Builds and CodeQL are awaiting maintainer workflow approval. One rebase-order note: #1305 touches the same `GspaApiEU` class-attribute block, so whichever merges first will require the other to rebase.

Ready to approve once CI runs.

blka (changes requested):

tastelessjolt:
Addressed both type-check nits in `e6eb7ac`, after rebasing onto current `master` (`a2c5485`):

- `SUPPORTED_LANGUAGES` is now consistently declared as `ClassVar[list[str]]` on the shared base and Korea override.
- Removed the string-based `getattr`/`setattr` seat dispatch. Korea capability parsing now assigns the four declared `Vehicle` fields directly, and a typed helper supplies the corresponding capability and request values for payload construction and validation. Malformed non-integer capability values remain unavailable instead of entering the vehicle model.

The reported mypy errors at the language override and seat validation are gone. Verification is clean:

- 776 tests passed on Python 3.12, 3.13, and 3.14
- repository-configured pre-commit suite passed
- focused Korea and shared EU tests: 116 passed
- branch is zero commits behind `master`

Test Builds and CodeQL will still need maintainer approval for the new head revision.

hyundai_kia_connect_api/Token.py
```diff
@@ -45,6 +51,44 @@ def to_dict(self) -> dict:
...
+    def to_persistent_dict(self) -> dict:
+        """Return refreshable session data without account or control secrets."""
+        data = self.to_dict()
+        data["password"] = None
+        data["pin"] = None
+        data["control_token"] = None
+        data["control_token_expiry"] = 0
+        return data
+
+    def save(self, path: str | os.PathLike) -> None:
```
cdnninja: This looks like a feature that isn't related to korea.  Please add it in another PR.

hyundai_kia_connect_api/Token.py
```diff
@@ -45,6 +51,44 @@ def to_dict(self) -> dict:
...
+            os.replace(temporary_name, destination)
+            os.chmod(destination, 0o600)
+        finally:
+            try:
+                os.unlink(temporary_name)
+            except FileNotFoundError:
+                pass
+
+    @classmethod
+    def load(cls, path: str | os.PathLike) -> "Token":
```
cdnninja: Same goes for this.

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -79,6 +79,22 @@ class Vehicle:
     # Not part of the API, enabled in our library for scanning.
     enabled: bool = True
 
+    # Connected-car capability metadata
+    remote_control_generation: str | None = None
```
cdnninja, line 83: This is "generation" above.

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -79,6 +79,22 @@ class Vehicle:
...
+    supports_remote_start: bool | None = None
+    remote_control_waiting_time: int | None = None
+    hvac_temperature_type: int | None = None
```
cdnninja, line 86: What does this mean?  Is this unit?

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -79,6 +79,22 @@ class Vehicle:
...
+    front_left_seat_climate_capability: int | None = None
```
cdnninja, line 87: What does the data in these mean?

hyundai_kia_connect_api/Vehicle.py
```diff
@@ -79,6 +79,22 @@ class Vehicle:
...
+    front_right_seat_climate_capability: int | None = None
+    rear_left_seat_climate_capability: int | None = None
+    rear_right_seat_climate_capability: int | None = None
+    steering_wheel_heater_option: int | None = None
+    steering_wheel_heating_option: int | None = None
+    supports_steering_wheel_heater: bool | None = None
```
cdnninja, line 93: If starting to do more "supports" aspect I think we should have a different PR and refactor all regions.

hyundai_kia_connect_api/Vehicle.py
tastelessjolt: No, this is not the Celsius/Fahrenheit unit.

I checked the signed MyHyundai 1.6.0 APK. The app reads `hvacTempType` from `infolist_v2.do`, normalizes `1` to `1` and anything else to `0`, stores it with the vehicle capability data, and passes it unchanged beside the requested temperature in both engine and climate requests. Temperature unit is modeled separately.

The best-supported interpretation is a selector for the vehicle HVAC temperature encoding or lookup profile: `1` is the alternate/newer profile and `0` the default/legacy profile. Hyundai does not expose names for these profiles, so I do not want to label them more specifically. The tested Korean Avante reports `1`.

I will document the field as: "Vehicle-advertised selector for the remote HVAC temperature encoding profile; separate from unit and echoed unchanged in climate-control requests."

hyundai_kia_connect_api/Token.py
tastelessjolt: Removed from #1309 in e124863. The generic file-persistence methods and their tests are now isolated in draft PR #1326.

hyundai_kia_connect_api/Token.py
tastelessjolt: Removed from #1309 in e124863. `Token.load()` and its tests are now isolated with `Token.save()` in draft PR #1326.

tastelessjolt:
@cdnninja I traced these fields through BetterBlueKit and the signed MyHyundai Korea 1.6.0 APK. BetterBlueKit has no Korea capability model, so the APK is the useful source here.

The raw values mean:

- `appMode` is not the same concept as the existing Hyundai USA integer `Vehicle.generation`. The APK parses it as an `AppMode` enum with values including `GEN1`, `GEN2`, `MOZEN`, `FE`, several EV/PHEV modes, and commercial-vehicle modes. It uses that enum for window-status support, remote engine control, engine remaining time, diagnostics, and other behavior. `remote_control_generation` is therefore a poor name. `app_mode` matches the source and meaning better.
- `startYn` is a backend eligibility flag for remote engine/climate control. `supports_remote_start` states more than that one flag proves because the app also considers `AppMode` and other options. `remote_start_eligible` would be safer if we keep a normalized boolean.
- `remoteControlTime` becomes `remoteControlWaitingTime` in the app model and defaults to `96` when missing. I have not found evidence for its unit, so I would keep the current waiting-time name without documenting it as seconds.
- `hvacTempType` is not Celsius/Fahrenheit. It selects the vehicle's temperature encoding table and is sent beside the temperature in engine/climate requests. The app normalizes it to `0` or `1`. BetterBlueKit independently uses the same flag to choose between two lookup tables. `hvac_temperature_profile` describes this better than `hvac_temperature_type`.
- The four seat values are capability profiles, not current seat states. Codes `1`/`2` advertise two/three heat levels, `3`/`4` two/three ventilation levels, `5`/`6` combined heat and ventilation, and `8`/`9` reduced one-level variants. Code `0` means no seat climate control. The decompiled mapper gives code `7` no active state, but its business meaning is not named. Korea command states are `2` for off, `3`-`5` for ventilation, and `6`-`8` for heat. The app ignores the seat capability object when `airControlType == 0`.
- `strgWhlHeatOption` and `strgWhlHeatingOption` are separate nullable values. The first value being `2` is specifically used for stepped/two-level steering-wheel heating. I could not establish the full value table for the second. The app derives ordinary and stepped steering support from several inputs, so our `supports_steering_wheel_heater = either_option > 0` is too broad.
- `windowSafetyOption == 1` selects the binary `Open` field. `windowSafetyOption2` values `3` and `4` select `OpenLevel`; value `4` also makes the rear-window states invalid. The APK additionally checks `AppMode.isSupportStatusWindow()`.
- `window_status_capabilities_loaded` is only our cache sentinel. It is not vehicle metadata.

I agree with the concern about adding more general `supports_*` fields to `Vehicle`. Instead of exposing every Korea-only raw option and derived boolean at the shared top level, I suggest grouping the request/status inputs:

```python
@dataclass
class KoreaVehicleCapabilities:
    app_mode: str | None = None
    remote_start_eligible: bool | None = None
    remote_control_waiting_time: int | None = None
    hvac_temperature_profile: int | None = None
    seat_climate_profiles: dict[str, int] = field(default_factory=dict)
    steering_wheel_heater_option: int | None = None
    steering_wheel_heating_option: int | None = None
    window_safety_option: int | None = None
    window_safety_option2: int | None = None

# One declared field on Vehicle, so diagnostics still serialize it.
korea_capabilities: KoreaVehicleCapabilities | None = None
```

The Korea client can derive command eligibility from this object when needed. `None` already means capability discovery has not completed, so the separate loaded sentinel can go away. This also avoids publishing uncertain support booleans before we understand the app's complete predicates.

I have not changed the implementation for this yet. I will wait for your view on this model boundary and naming before restructuring the PR.

PR #1314, opened 2026-09-16
feat(svm): SVM image rendering in the library (crop, dewarp, encode)

blka:
Moves the SVM presentation-layer image logic out of kia_uvo (requested in
Hyundai-Kia-Connect/kia_uvo#1745 by fuatakgun) into the library:

- new `svm_image` module: `render_views(details, dewarp=False)` returns all
  5 views (front/rear/left/right/top) as JPEG bytes in one call; public
  low-level `crop_view` / `dewarp_fisheye` / `camera_fov_deg` /
  `to_jpeg_bytes` / `VIEWS` for power consumers (panoramic viewers)
- dewarp math moved 1:1 from the integration; FOV auto-derived from
  `valid_angle_of_view`, output FOV = clamp(fov*0.78, 50, 85); TOP is never
  dewarped
- optional `image` extra (`Pillow>=10.0.0`, `numpy>=1.26.0`); imports stay
  lazy — the module imports without the extra, calling raises ImportError
  with an install hint
- views whose crop fails are omitted from the returned dict
- test workflow installs `-e .[image]`

kia_uvo will consume `render_views()` in a follow-up PR against
`hyundai_kia_connect_api[image]==4.32.0`.


hyundai_kia_connect_api/svm_image.py
```diff
@@ -0,0 +1,263 @@
...
+import math
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+from .svm import SVMDetails
+
+if TYPE_CHECKING:
+    from PIL import Image
+
+IMAGE_EXTRA_HINT = "pip install hyundai_kia_connect_api[image]"
```
cdnninja, line 34: I don't believe all systems use pip.    I need to read more if this approach is best practice or not.

blka: Fair point — the message now names the extra instead of the tool, with pip only as an example: `install the hyundai_kia_connect_api[image] extra (e.g. pip install 'hyundai_kia_connect_api[image]')`. Home Assistant consumers get the extra installed from the integration's manifest anyway, so the hint is only for direct library consumers. Pushed as 99f8d0b.

PR #1318, opened 2026-09-18
fix(EU): treat a battery state of health of 0 as unknown

smitec11:
## What this fixes

My Kia Niro EV (Europe) reports `vehicleStatus.evStatus.batterySoh: 0` in every update. The field is present, so the value is passed straight through and `ev_battery_soh_percentage` becomes 0. In Home Assistant that shows up as a battery state of health of 0%, which then also lands in long-term statistics and would trip any automation watching battery health.

A vehicle that drives does not have a 0% state of health, so this looks like "not reported" rather than a real measurement. This keeps the value unset when the vehicle sends 0, so it stays unknown instead of wrong.

```python
if battery_soh := get_child_value(state, "vehicleStatus.evStatus.batterySoh"):
    vehicle.ev_battery_soh_percentage = battery_soh
```

## Tests

Two tests in `tests/test_eu_battery_soh.py`, both using the existing EU fixture: a reported value of 88 ends up on the vehicle, and a reported 0 leaves it as `None`. The second one fails without this change.

`tests/test_eu_vehicle_properties.py` still passes (17 tests in total), and ruff is clean on both files.

## Scope

Only the legacy EU path (`KiaUvoApiEU`), which is the one I can demonstrate with my own vehicle. The CCS2 paths in `ApiImplType1` and `HyundaiCciApiEU` read `Green.BatteryManagement.SoH.Ratio` and have the same shape, so they could need the same guard, but I have no vehicle reporting that to confirm it. Happy to extend this PR to those two if you want it.

*(Disclosure: I used Claude Code to look into this and to write the change and its tests, on my own vehicle data.)*


fellek:
A data point from the CCS2 / EU CCI side that supports treating 0 as unknown:

- The GSPA `stored-status` of a Kia PV5 (2026, EV, EU CCI) contains no
  `Green.BatteryManagement.SoH` block at all, maybe there is need for change to publish a value.
- The Kia Driving Insight endpoint `GET /ev-battery/status` returns `data: null` for the same
  car.
- The Kia Driving Insight web app handles this value the same way as this PR: it throws
  `"soh is 0."` and shows "no data" instead of a percentage when `soh === 0`.

So for the CCS2 paths (`ApiImplType1`, `HyundaiCciApiEU`) the same guard would match what the
official app does.

smitec11:
Thanks @fellek, that settles the part I could not demonstrate myself. I extended the PR to `ApiImplType1` and `HyundaiCciApiEU` with the same guard.

While doing that, one existing fixture in this repository turned out to report it as well: `eu_kia_sportage_phev_2027_ccs2.json` has `SoH.Ratio: 0`, so its snapshot now expects `ev_battery_soh_percentage: None` instead of 0. That snapshot change is the visible proof that the guard does what it should.

Added CCS2 tests alongside the existing EU ones: a reported 95 ends up on the vehicle, a reported 0 leaves it unset. Full suite passes (729 tests) and ruff is clean.

PR #1322, opened 2026-09-20
fix(eu): recover from a wedged legacy refresh-token login

blka:
## Summary

Accounts configured before headless password login existed can hold a **48-char refresh token as their stored password** (the pre-password model, where the password field was the refresh token). When such an entry's token set expires, `refresh_access_token()` falls back to a full login with `Token.password` — and the login dispatch (`login()` in `KiaUvoApiEU`) re-interprets that 48-char value as a refresh-token grant. The fallback can never succeed, and the account wedges permanently with `AuthenticationError: Received unexpected statusCode` even though its real credentials are valid.

Observed live in https://github.com/Hyundai-Kia-Connect/kia_uvo#1888: a reporter's integration recovered only after manually restoring the real password and clearing the persisted token, while the same credentials logged in fine headlessly (our probe script: T1 passes end-to-end). Several other #1888 reporters show the same signature ("app works, integration won't re-auth").

## Fix

`VehicleManager.check_and_refresh_token()` catches the `AuthenticationError` from `refresh_access_token()` and retries once with the credentials the manager was constructed with — but **only** when all of the following hold:

1. `token.password` matches the legacy 48-char refresh-token shape (the wedge signature),
2. the manager's configured password differs from it,
3. the configured password itself is not a 48-char refresh token (retry would dead-end identically).

All other authentication failures propagate unchanged, so no additional credential requests are made on any other path.

Closes https://github.com/Hyundai-Kia-Connect/kia_uvo#1888 — for the accounts affected by this wedge (the bucket-B reporters, where both the legacy and the SPA sign-in reject valid credentials, remain a separate Hyundai-side matter tracked in the issue).

## Tests

`tests/vehicle_manager_stale_token_test.py` (4 cases, written test-first):

- wedged token + real configured password → retry happens with the configured password, refresh succeeds
- failing password is the real password (not the wedge shape) → **no** retry, error propagates
- both passwords are 48-char refresh tokens → **no** retry (retry would dead-end identically)
- plain-password auth failure → no retry, error propagates

Full suite: 767 passed (3 pre-existing local-only failures in untracked `tests/integration/`, not part of CI).


hyundai_kia_connect_api/VehicleManager.py
```diff
@@ -240,9 +241,29 @@ def check_and_refresh_token(self) -> bool:
...
+                # a fallback login with it re-enters the refresh-token grant
+                # and wedges the account (kia_uvo #1888). Retry once with the
+                # credentials the manager was constructed with.
+                if (
+                    re.fullmatch(r"[A-Z0-9]{48}", self.token.password or "")
+                    and self.password
+                    and self.password != self.token.password
+                    and not re.fullmatch(r"[A-Z0-9]{48}", self.password)
+                ):
+                    _LOGGER.warning(
```
cdnninja, line 259: Why would we clutter logs with this if it may recover no issue?

cdnninja:
Actually thinking on this more, why doesn't our token match setup? Is this a Kia uvo bug?

blka:
Both, in a sense — it's an interaction across the two layers:

**Why the token doesn't match the setup:** `Token.password` is a snapshot taken when the token object was created, and the refresh fallback (from #1277) re-uses it as the login credentials. For accounts configured before headless password login existed, that stored "password" *is* a 48-char refresh token — the pre-password model. When the token set expires, the fallback re-enters `login()`, which re-interprets the 48-char value as a refresh-token grant and dead-ends forever, even though the manager's configured password is valid.

**The kia_uvo side:** there is also a genuine kia_uvo bug — reconfigure merges the config instead of replacing it, so it persists the stale `CONF_TOKEN` while discarding the fresh token the full login just produced (fix in Hyundai-Kia-Connect/kia_uvo#1899; the OTP path already persisted correctly). That's how some modern accounts ended up with stale tokens in storage too.

**Why a lib fix anyway:** the fallback lives in the lib and trusts `Token.password` as current credentials — an assumption that only holds for tokens freshly created by this manager. Even with the kia_uvo persistence fixed, already-wedged accounts (48-char stored password) stay wedged, so this PR is the recovery path for them. The retry gate keeps it narrow: only when the stored password matches the legacy 48-char shape, the configured password differs, and the configured password isn't itself a 48-char token. Merge order: this PR first, then the kia_uvo persistence fix.

blka:
@cdnninja reply posted to your question above (5750257173) — short version: the token mismatch is the pre-password model's stored password meeting the #1277 fallback, and the kia_uvo reconfigure-merge bug (Hyundai-Kia-Connect/kia_uvo#1899) explains how modern accounts got stale tokens too. When you get a chance, a review here would unblock the recovery path for the wedged accounts.

